PR #2, opened 2018-02-14
Add Travis CI builds for Linux and macOS.

brcolow:
This PR adds support for building OpenJFX using [Travis CI](https://travis-ci.org/).

In order to make this work, the project will have to be (freely) registered with Travis. I personally login with my Github credentials to link a project when I use Travis.

Here is an example Travis build with the same configuration (just on my temporary openjfx repo):

https://travis-ci.org/brcolow/openjfx/builds/341230348

Note that the builds that are listed as "untagged-xxxxx" would not appear on this repo, that was just for testing purposes to automatically upload failed test screenshots.

### DIFF
--- a/.ci/script.sh
+++ b/.ci/script.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+set -uo pipefail
+
+BOOT_JDK10_VER="42"
+BOOT_JDK10="jdk-10-ea+42"
+
+echo "which java: $(which java)"
+ulimit -c unlimited -S
+
+gradle build -x :web:test --no-daemon --stacktrace --info
+
+# Print core dumps when JVM crashes.
+RESULT=$?
+
+if [[ ${RESULT} -ne 0 ]]; then
+  if [ ! -z ${PRINT_CRASH_LOGS+x} ]; then
+    if [[ "${TRAVIS_OS_NAME}" == osx ]]; then FIND="gfind"; else FIND="find"; fi
+    ${FIND} . -name "hs_err_pid*.log" -type f -printf '\n====== JVM CRASH LOG ======\n%p\n' -exec cat {} \;
+
+    CORES=''
+    if [[ "${TRAVIS_OS_NAME}" == osx ]]; then
+      CORES="$(find /cores -type f -regex '*core.[0-9]{4}' -print)"
+    else
+      CORES="$(find . -type f -regex '*core.[0-9]{6}' -print)"
+    fi
+
+    if [ -n "${CORES}" ]; then
+      for core in ${CORES}; do
+        printf '\n\n======= Core file %s =======\n' "$core"
+        if [[ "${TRAVIS_OS_NAME}" == osx ]]; then
+          lldb -Q -o "bt all" -f "$(which java)" -c "${core}"
+        else
+          gdb -n -batch -ex "thread apply all bt" -ex "set pagination 0" "$(which java)" -c "${core}"
+        fi
+      done
+    fi
+  fi
+  exit ${RESULT}
+fi
+
+# Download JDK 10 to use for boot JDK for building JDK 11.
+cd "$TRAVIS_BUILD_DIR"
+if [[ "${TRAVIS_OS_NAME}" == osx ]]; then
+  BOOT_JDK_MAC="${BOOT_JDK10}_osx-x64_bin.dmg"
+  wget https://download.java.net/java/jdk10/archive/"$BOOT_JDK10_VER"/BCL/"$BOOT_JDK_MAC"
+  hdiutil attach "$BOOT_JDK_MAC" -mountpoint ~/_mount
+  # The following dance is done to avoid calling "installer" which requires root.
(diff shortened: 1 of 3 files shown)


.ci/script.sh
```diff
@@ -0,0 +1,79 @@
...
+  ORACLE_JDK10='jdk-10-ea+42_linux-x64_bin.tar.gz'
+  wget https://download.java.net/java/jdk10/archive/42/BCL/"$ORACLE_JDK10"
+  # wget https://download.java.net/java/jdk10/archive/42/GPL/"$OPEN_JDK10"
+  tar xvf "$ORACLE_JDK10"
+fi
+
+
+# Now that OpenJFX has been built, we will build OpenJDK and configure
+# it to use our newly built OpenJFX>
+hg clone http://hg.openjdk.java.net/jdk/jdk
```
johanvos: A shallow clone of the AdoptOpenJDK might be more appropriate here? (git clone --depth=1 -b dev https://github.com/adoptopenjdk/openjdk-jdk10.git) It will be faster as it doesn't download the history

bourgesl:
I have travis CI credentials so we can use mine if you agree; or use the credentials for the javafxports account ?

How long is the build ? yours take 1h15 mins so it is quite long ... shorter build time would be greater or only enable the CI on a specific branch ?

brcolow:
CI should be enabled on master and for pull-requests, IMO.

tiainen:
I think it's the most convenient to link Travis CI with the javafxports account itself. So that what I've done.

bourgesl:
I think travis CI (OSS) has limits in terms of build duration and log size...
I could try your travis CI PR on my jfx fork to see it in action...

brcolow:
You can see it in action here: https://travis-ci.org/brcolow/openjfx/builds/341230348. Building and running the test suite is completed in the time alloted by Travis. Building OpenJDK with OpenJFX overlay modules is not possible in the time alloted by Travis.

.ci/script.sh
```diff
@@ -0,0 +1,79 @@
...
+  fi
+  exit ${RESULT}
+  # exit 0 # Force success so images are uploaded.
+fi
+
+# Download JDK 10 to use for boot JDK for JDK 11.
+cd "$TRAVIS_BUILD_DIR"
+if [[ "${TRAVIS_OS_NAME}" == osx ]]; then
+  ORACLE_JDK10='jdk-10-ea+42_osx-x64_bin.dmg'
+  wget https://download.java.net/java/jdk10/archive/42/BCL/"$ORACLE_JDK10"
```
bourgesl: Could you define a BOOT_JDK variable to define which jdk file (prefix) to download ?
Why do you need ORACLE jdk instead of OpenJDK EA builds ? (need jfx modules ?)

.ci/script.sh
```diff
@@ -0,0 +1,79 @@
...
+  hdiutil attach "$ORACLE_JDK10" -mountpoint ~/_mount
+  # The following dance is done to avoid calling "installer" which requires root.
+  pkgutil --expand ~/_mount/JDK\ 10.pkg ~/jdk-10-compressed
+  mkdir -p ~/jdk-10
+  tar xvf ~/jdk-10-compressed/jdk10.pkg/Payload -C ~/jdk-10 --strip-components 3 # Strip Contents/Home/
+else
+  OPEN_JDK10='openjdk-10-ea+42_linux-x64_bin.tar.gz'
+  ORACLE_JDK10='jdk-10-ea+42_linux-x64_bin.tar.gz'
+  wget https://download.java.net/java/jdk10/archive/42/BCL/"$ORACLE_JDK10"
```
bourgesl: Idem for linux

johanvos:
While it would be nice if the inclusion in OpenJDK could be part of the Travis build, I'm not worried that it won't happen at the moment. That integration seems to belong in the AdoptOpenJDK build process. In our repository here, we need to make sure the builds succeed and tests run, and it is a huge step if we can use Travis CI for this.

.ci/script.sh
brcolow: Using OpenJDK as the boot JDK I got errors when running the OpenJDK configure script that said the boot JDK was the wrong version. The errors did not happen with Oracle JDK.

bourgesl: I mean use a more generic variable name and possibly move BOOT_JDK_MAC and BOOT_JDK_LINUX at the top of the script.
Ideally these could share the 'jdk-10-ea+42' prefix.
That's good enough and can be made later

brcolow: Okay, I did something very close to that, where it is easy to change the JDK10 version at the top of the script.

brcolow:
As a Travis CI security measure the build won't run using the `.travis.yml` in this PR until after it's merged. I believe that only happens when `master` does not yet contain ` .travis.yml`. Just FYI.

bourgesl:
Looks good for me,
Waiting for another opinion, johan ?

bourgesl:
Thank you so much, it is tricky CI scripts

bourgesl:
Will do as soon as I have some time to switch on my laptop... today

bourgesl:
Fixed in master: it works perfectly !

PR #9, opened 2018-03-07
Set the locale of the javadoc

ebourg:
Setting the locale to a fixed value when generating the javadoc ensures the output of the build doesn't depend on developer specific settings. This improves the [reproducibility](https://reproducible-builds.org) of the build.


kevinrushforth:
This looks fine to me.

kevinrushforth:
I see from the history that I (inadvertently) closed this PR. I'm not sure how that happened; I was just trying to add a comment.

johanvos:
Each system has its own strange behaviour :)
I think there is a button "Close and comment" that might be mistaken for "close this comment"

PR #10, opened 2018-03-07
Increase the timeout for processing the ANTLR grammar

ebourg:
Increasing the ANTLR timeout fixes build failures on slow machines.


johanvos:
I think this should not be in buildSrc/build.gradle but rather in build.gradle, line 1762

johanvos:
@ebourg do you agree this should be done in build.gradle instead of buildSrc/build.gradle?

ebourg:
Yes that makes sense.

johanvos:
These changes are part of PR #68 (they had to go in build.gradle instead of a platform-specific one). Added the original author (@ebourg) to the Contributed-by entry.

ebourg:
Thanks!

Issue #8, opened 2018-03-07
GitHub Bot


brcolow:
As I have mentioned on the openjfx-dev mailing list, I believe running a simple bot will automate the process of getting contributions in the state where they are more likely to be accepted by upstream OpenJFX. My idea is:

Configure this GitHub repository to have a webhook for pull requests that points to a URL of the bot.

Use the GitHub status API to add a status check which is managed by the bot. This is similar to how we currently have status checks for Travis CI and Appveyor.

The bot does a couple things:

1.) Checks to see if the user who opened the PR has signed the OCA. We can keep a simple list of associations between github usernames and the name/email they signed the OCA under. The bot can post a comment on the PR asking the user if they have signed the OCA and if so under what name/email. If they reply they haven't signed it yet, the bot can instruct them how to do so.
2.) Take the raw patch of the PR (the result of `git format-patch`) and convert it to a mercurial commit which can then be `hg import`ed into a local instance of the OpenJFX upstream mercurial repository.
3.) Generate a webrev using the mercurial commit against OpenJFX upstream master.
4.) Runs [jcheck](http://openjdk.java.net/projects/code-tools/jcheck/) against the mercurial commit/changeset.
5.) Checks to see if this PR is associated with a JBS bug report and if so, links to it on the status page of that PR.

The status check will provide a link to a simple page that shows the OCA status, the mercurial patch, the webrev, jcheck results, and any linked JBS bug reports  (this is similar to how clicking on the Travis CI status check brings one to the Travis CI build for that PR).

The status of the status check will be set depending on:

* Pending when we first receive the pull request event from the webhook.
* If the user has not yet signed the OCA, the status is set to failed.
* If the mercurial commit does not apply cleanly to OpenJFX upstream master, the status is set to failed.
* If for some reason we can't generate the webrev, the status is set to failed.
* If jcheck fails, the status is set to failed.
* Otherwise, the status is set to successful.

GitHub status API:

https://developer.github.com/v3/repos/statuses/

Github Pull request webhook event:

https://developer.github.com/v3/activity/events/types/#pullrequestevent

At first I thought we could easily use some type of "serverless" compute like Amazon Lambda, but having to interact with a mercurial repository of OpenJFX complicates things and I think things will be much easier having an application running on a server. We can do something very simple in Java, say, using Jersey. It would expose one REST endpoint (that github will send the pull request events to via the webhook).

[...]

PR #11, opened 2018-03-07
Fix PosixPlatform.cpp compilation with GCC 6

ebourg:
PosixPlatform.cpp fails to compile with GCC 6:

    /<<PKGBUILDDIR>>/modules/fxpackager/src/main/native/library/common/PosixPlatform.cpp:235:10: error: 'wait' was not declared in this scope
         wait();
              ^

This issue has been fixed in Debian.

### DIFF
--- a/modules/jdk.packager/src/main/native/library/common/PosixPlatform.cpp
+++ b/modules/jdk.packager/src/main/native/library/common/PosixPlatform.cpp
@@ -53,6 +53,7 @@
 #include <algorithm>
 #include <dlfcn.h>
 #include <signal.h>
+#include <sys/wait.h>
 
 
 PosixPlatform::PosixPlatform(void) {
@@ -380,7 +381,7 @@ bool PosixProcess::Wait() {
     pid_t wpid = 0;
 
 #ifdef LINUX
-    wait();
+    wait(&status);
 #endif
 #ifdef MAC
     wpid = wait(&status);


bourgesl:
FYI I proposed already such fix on JBS... I can review & sponsor this PR soon

bourgesl:
I looks good. 
Do you know if wait(status) is available in gcc <= 6 ? I suppose yes, for long time.

 I wonder what is the current gcc version on openjfx builds & travisCI (I use gcc 6.3 on my machine)

ebourg:
I'm not an expert but it seems this function has been in glibc since at least 1995:
https://github.com/lattera/glibc/blob/28f540f45b/posix/sys/wait.h#L102

bourgesl:
I checked and travis CI uses the Ubuntu Trusty distribution (click on 'system info' in the build log):
Description:	Ubuntu 14.04.5 LTS
gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4

So it uses gcc 4.8 !
@kevinrushforth what is the target gcc version for OpenJFX 11 ?

Workflow:
1. I will test locally on my machine (gcc 6.3) & make a clean build
2. Approve the PR on the developp branch (github)
3. Make the webrev & upload it on http://cr.openjdk.java.net/
4. Send a request for review (RFR) on the openjfx-dev mailing-list

@johanvos is it the right procedure ?

ebourg:
That's a bit heavy for a one line change. I hope the workflow can be simplified in the future. Thank you for your help Laurent.

ebourg:
My email is ebourg@apache.org, and I already signed the OCA if that helps.

bourgesl:
I confirm this patch fixes the build on gcc 6 (lots of warning remain).

bourgesl:
I propose to accept the PR on the develop branch.
Any objection ?

kevinrushforth:
We use gcc 4.9.1 for producing builds, so if it works with gcc 4.8 it should be good to go. This change looks good to me, but is it enough to allow the build to pass on gcc 6?

I note that this bug is tracked by [JDK-8189689](https://bugs.openjdk.java.net/browse/JDK-8189689), and Laurent indicated that he also had to disable warnings as errors. I haven't tried it with gcc 6 myself yet.

kevinrushforth:
This PR is a good test case of the process as well, and thanks @bourgesl for sponsoring this. I'll assign [JDK-8189689](https://bugs.openjdk.java.net/browse/JDK-8189689) to you.

We can discuss ways to make it a bit less heavy of a process in the future, such as eliminating redundant steps. For example, even though webrev is typically what we use, I think we might consider a pointer to the changeset on GitHub (after it has been merged into the develop branch) as an alternative to webrev for the review request to openjfx-dev, as long as the contributor is an OCA signatory (which is a requirement for accepting the PR in the first place). There are probably other optimizations we can make as we get more experience with the flow.

modules/jdk.packager/src/main/native/library/common/PosixPlatform.cpp
```diff
@@ -53,6 +53,7 @@
 #include <algorithm>
 #include <dlfcn.h>
 #include <signal.h>
+#include <sys/wait.h>
```
kevinrushforth: Maybe it is better to put this within "#ifdef LINUX" ?

kevinrushforth:
I did add one comment inline (not sure if this is the usual way of commenting on code under review). As long as it builds on all three platforms, I don't really care all that much whether the include is within "#ifdef LINUX" or not. I just wanted to ask the question.

ebourg:
> This change looks good to me, but is it enough to allow the build to pass on gcc 6?

Good question, I don't remember well but I suspect this change in `buildSrc/linux.gradle` was also necessary:

    -LINUX.launcherlibrary.ccFlags = ["-DJAVAARCH=\"$OS_ARCH\"", "-I$JDK_HOME/include", "-I$JDK_HOME/include/linux", "-c", "-fPIC"]
    +LINUX.launcherlibrary.ccFlags = ["-DJAVAARCH=\"$OS_ARCH\"", "-I$JDK_HOME/include", "-I$JDK_HOME/include/linux", "-c", "-fPIC", "-std=gnu++98"]

And I also have some changes ready to support GCC 7:

https://anonscm.debian.org/cgit/pkg-java/openjfx.git/tree/debian/patches/21-gcc7-compatibility.patch

modules/jdk.packager/src/main/native/library/common/PosixPlatform.cpp
ebourg: I don't know, but this is a standard POSIX header, so I guess any os/arch building PosixPlatform.cpp is likely to support it.

ebourg:
I confirm `-std=gnu++98` was required to build with GCC 6:
https://anonscm.debian.org/cgit/pkg-java/openjfx.git/commit/debian/patches?id=7402e3f

kevinrushforth:
It seems you will need to update this patch, then, to allow it to actually build on gcc 6. I don't have a system handy to test this, but I presume that @bourgesl will do a build with gcc 6 before accepting the PR.

Separately, I can file a new bug to fix any gcc 7 errors. Depending on the time frame we want to do this, we might want to take the webkit patch you pointed out (we will be updating WebKit anyway in about 3 months).

ebourg:
I opened another pull request for `buildSrc/linux.gradle`: https://github.com/javafxports/openjdk-jfx/pull/16

modules/jdk.packager/src/main/native/library/common/PosixPlatform.cpp
bourgesl: No need to add ifdef, it is good

johanvos:
To clarify the procedure we agreed upon (and described above by Laurent): after discussion (referenced to at JBS) the github contributor (in this case @ebourg) creates a PR for the develop branch. 
The github reviewer (in this case @bourgesl ) merges the PR in develop, and creates and uploads a webrev, and sends the request to the openjfx-dev mailing list, where an OpenJFX reviewer (e.g. @kevinrushforth ) who followed the discussion can review it (this should hopefully be just a formal step, as discussion should already have happened at this point). The github reviewer (@bourgesl ) commits the patch in OpenJFX.

This means some administration overhead for the "github reviewer/OpenJFX committer" that we can hopefully reduce in the future. The GitHub bot proposed in https://github.com/javafxports/openjdk-jfx/issues/8 might help in creating the webrev based on the PR. 

One question for @kevinrushforth : what is the correct way to acknowledge contributors (in this case @ebourg ) when committing in OpenJFX?

kevinrushforth:
The above sounds good to me. As for acknowledging a contribution:

if (contributor has an openjdk ID)  // i.e., is a Project Author
    commit the patch with 'hg commit -u CONTRIBUTORS_OPENJDK_ID' (or with 'hg qfinish -u' if using MQ)
else
   add the following line in the hg commit message after the 'Reviewed-by' line:
    "Contributed-by: CONTRIBUTORS-EMAIL-ADDRESS@SOMEWHERE.COM"
endif

bourgesl:
Approved.
Fixes #11

bourgesl:
Thank you, emmanuel.
you can follow the jfx-dev integration in #11

PR #14, opened 2018-03-08
First small text

hendrikebbers:
better than a todo ;)


johanvos:
There is a README already -- crossed this PR

Issue #6, opened 2018-03-05
update README + CONTRIBUTING files


bourgesl:
In master, those files are almost empty although it is THE first place newcomers will see.

johanvos:
Yes, I have been thinking about that as well.
We have to be careful though as we're making changes to the mirror already (travis and appveyor) that are not going back upstream. 
But a README (that also includes CI status) makes perfect sense.
@kevinrushforth do you have an opinion on this?

kevinrushforth:
I think as long as changes that are specific to the mirror (i.e., not going back upstream) are kept localized, ideally in separate files, that seems fine.

johanvos:
Unless someone disagrees, I will move the CONTRIBUTING.md from https://github.com/johanvos/openjfxorg/blob/master/CONTRIBUTING.md (which I mentioned on the mailing list http://mail.openjdk.java.net/pipermail/openjfx-dev/2018-February/021509.html) to the master, without first pushing it to develop and upstream to OpenJFX.

bourgesl:
Only on github, I suppose ?

johanvos:
yes

eugener:
One small improvement we can do here is move all those mirror specific files into `.github` folder to keep things clean. This is one of the standard practices on GitHub. This folder can also contain issue templates, pull request templates and more.

kevinrushforth:
I like the idea of a separate folder for mirror-specific files. In fact if you do this, it would make it easier / more palatable to merge these upstream in the near future, if desired.

bourgesl:
Good idea.
Will github show the .github/readme on the project page ? that would be perfect

eugener:
It does appear to be the case according to [GitHub Docs ](https://help.github.com/articles/about-readmes/)

johanvos:
sounds good to me.
That means the README (we still have to write this) and the CONTRIBUTING.md only, or also other files?

eugener:
New PR #18 covers the discussion around `.gihub` folder and related files. I also took a liberty to create an initial version of readme file, which includes build status badges from both TravisCI and AppVeyor. Check it out at my fork at https://github.com/eugener/openjdk-jfx

johanvos:
PR #18 is merged

PR #15, opened 2018-03-09
Move github related files under .github folder

eugener:
This PR is related to discussion on issue #6


bourgesl:
Which README will be shown ?

bourgesl:
I mean the hg repo had a readme...

eugener:
This PR moves it to `.github` folder. So if it is not in the root - it will be shown from there.
Check it out in my fork at https://github.com/eugener/openjdk-jfx

bourgesl:
There is a conflict as the openjfx hg forrest has a README file at root.
Having another one in .github/ is OK, but the first one will be displayed.

eugener:
@kevinrushforth how important is to have the `README` file at the root in hg? Can we move it to `.github` folder here?

kevinrushforth:
You mean the existing one (which is almost content-free)? That would seem odd in a .github directory, and since it doesn't really add any value, I think we can just remove it altogether.

kevinrushforth:
The LICENSE, on the other hand, should remain at the top level. Will that cause problems?

bourgesl:
Sorry to insist, but it is important for me to put some content (CI badges, links, description) in the README as described in #6 that is displayed on the homepage of the project, for example:
https://github.com/bourgesl/marlin-renderer
Newcomers need to understand what is this project

johanvos:
I agree with @kevinrushforth and unless I'm missing something that should also fix the concerns from @bourgesl  : remove the README from the OpenJFX repository (there is no real content in there) and create a new README in .github that will be shown at the github homepage.
There might be some issues in syncing as the OpenJFX README has a history so it is somehow there, but that is something we can fix in the sync script.

kevinrushforth:
If you remove the README file from top of the repo and then create a brand new README.md file in .github dir, then I would think there should be no problems w.r.t. history. Or is the concern that the patch already moved the existing README from top level to .github? In that case, yeah, you will need to fix it up somehow.

And just to check, there is no need to move the LICENSE file, right? We could (if needed) duplicate it in .github but it does need to be at the top of the repo.

johanvos:
I think the LICENSE can indeed just stay where it is.

eugener:
I will update the PR based on comments here

eugener:
To keep the commit history clean I'm closing this pull request. The new one will be referenced from the  related issue #6

PR #16, opened 2018-03-09
Fix launcher library compilation with GCC 6

ebourg:
Another trivial change required to build with GCC 6.


kevinrushforth:
Two comments:

1. Do you also need to remove "-Werror" from the LINUX.glass.glass.ccFlags or was that not needed?

2. It seems better to fold this into #11 so that we get a single changeset to fix the bug (rather than two changesets), since that will make it easier to review and merge upstream.

ebourg:
> 1. Do you also need to remove "-Werror" from the LINUX.glass.glass.ccFlags or was that not needed?

This wasn't necessary.

> 2. It seems better to fold this into #11 so that we get a single changeset to fix the bug

I don't think I can merge the two PRs because I used the GitHub editor to submit them. But maybe @bourgesl can merge or squash them into a single feature branch.

kevinrushforth:
OK, in that case I'll leave it up to @bourgesl to figure out how best to deal with this (I'm new to github, so still learning what is easy / possible).

In general, we need the ability to iterate on a bug fix review by having the contributor push additional changesets to the branch for that bug and folding them into a single changeset. I'll do some experimenting at some point to get a better feel for this.

bourgesl:
Which gradle task is failing on gcc 6 ?

bourgesl:
Please use the develop branch for PR. 
Could you merge this change into your patch-3 branch ? so the PR #11 will be updated:
https://github.com/ebourg/openjdk-jfx/tree/patch-3
That would be simpler.

bourgesl:
Kevin, as I will make a webrev anyway, I will merge #11 and #16 into a single hg changeset.
I could try merging both on github, but I am not a git expert and it seems too complicated for such small changes.

kevinrushforth:
OK, thanks.

ebourg:
> Could you merge this change into your patch-3 branch ?

Ok, I didn't realize GitHub forked the repository under my namespace when I edited the file. I'll do that.

bourgesl:
@ebourg Don't waste your time, I will approve both PR on javafxports:develop branch and then prepare the patch on the openjfx hg repo now

bourgesl:
Approved. I will merge both changes within the JBS webrev for  #11

ebourg:
Thanks!

PR #17, opened 2018-03-09
FX print tests fail with NPE in some environments

bourgesl:
Here is a simple patch to solve the windows CI (NPE happening).

First attempt ...


bourgesl:
CI failed but seems not related to my code changes.
Unstable build or env ?

johanvos:
Seems the JVM crashed with the error log at C:\projects\openjdk-jfx\tests\system\hs_err_pid2864.log
Is there a way we can see that log?

bourgesl:
I will enable appveyor on my fork and debug the CI ...
This PR was just a trial but CI still fails.

johanvos:
Actually, the build passes now on AppVeyor (see line 23436 of the build log).
However, after building, there is another pass of the test script, and that cause the jvm error.

On travis, we simply do this:
gradle build -x :web:test --no-daemon --stacktrace --info

On AppVeyor, we do this:
gradle build -PCOMPILE_WEBKIT=false --stacktrace -x :web:test --info --no-daemon
gradle test -x :web:test -PFULL_TEST=true --stacktrace --info --no-daemon

johanvos:
running on Linux with -PFULL_TEST=true also results in jvm crash.
I could see the log for this, and it is due to webkit calling a static void method:

V  [libjvm.so+0x8aa239]  jni_CallStaticObjectMethodV+0x59
C  [libjfxwebkit.so+0x11f8896]  JNIEnv_::CallStaticObjectMethod(_jclass*, _jmethodID*, ...)+0x86

I propose we create a separate issue for this, and modify Appveyor so that it doesn't do the FULL_TEST.

bourgesl:
Excellent, you are so good at diagnosing core dump. 
You proposal seems reasonable.
I will have some time tomorrow night to test & look at appveyor build changes.

Could you file a JBS issue about Full test crash ?

johanvos:
We now see a "Build successful" in the logs, but AppVeyor still says it failed?

bourgesl:
Yes it is test uploads that make the build fail...
Trial & error... I never played with win powershell before

johanvos:
There are only 4 directories with testresults/test output it seems. I removed the others in
https://github.com/johanvos/openjdk-jfx/blob/master/appveyor.yml

However, still fails for me, this time in building:
Cannot run program "C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.12.25827/bin/HostX64/x64/cl.exe" (in directory 
"C:\projects\openjdk-jfx-vwnmn\modules\javafx.graphics"): CreateProcess error=2, The system cannot find the file specified

bourgesl:
Seems appveyor updated vs2017 today:
https://www.appveyor.com/updates/

appveyor.yml
```diff
@@ -7,24 +7,37 @@ environment:
...
+  - ARCHITECTURE: x64
+    MSVC_VER: 15
+    APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+
+matrix:
+  fast_finish: true
 
 shallow_clone: true
 
+clone_depth: 5                      # clone entire repository history if not defined
```
brcolow: This is not necessary because shallow_clone sets clone depth to 1.

appveyor.yml
```diff
@@ -7,24 +7,37 @@ environment:
...
-  MSVC_VER: 14.12.25827
+  # MSVC_VER: 14.12.25827
   APPVEYOR_CACHE_ENTRY_ZIP_ARGS: "-t7z -m0=lzma -mx=9"
+  matrix:
+  - ARCHITECTURE: x64
+    MSVC_VER: 15
+    APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+
+matrix:
+  fast_finish: true
```
brcolow: We only have one build...so adding this currently does nothing.

appveyor.yml
```diff
@@ -7,24 +7,37 @@ environment:
...
       choco install gradle --version 4.3.0
       refreshenv
-  - cmd: |
-      gradle build -PCOMPILE_WEBKIT=false --stacktrace -x :web:test --info --no-daemon
+      # detailed gradle build: add --info
+      gradle build -x :web:test -PCOMPILE_WEBKIT=false --no-daemon --stacktrace
+
+# Full test crash
+#test_script:
+#  - gradle test -x :web:test -PCOMPILE_WEBKIT=false --no-daemon --stacktrace
```
brcolow: Removing this line makes the test directories non-existent, so the test results uploading will fail, because no tests are being run.

appveyor.yml
```diff
@@ -7,24 +7,37 @@ environment:
   VCINSTALLDIR: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build
   VS150COMNTOOLS: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build
   VSVARS32FILE: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars32.bat
-  MSVC_VER: 14.12.25827
+  # MSVC_VER: 14.12.25827
   APPVEYOR_CACHE_ENTRY_ZIP_ARGS: "-t7z -m0=lzma -mx=9"
+  matrix:
+  - ARCHITECTURE: x64
+    MSVC_VER: 15
```
brcolow: I believe we discovered on the mailing list that OpenJFX gradle build files do not expect the actual MSVC version number, but some weird pseudo-version which happened to be `14.12.25827` at the time of my making the appveyor PR. I did not find a way to fetch this version from the Appveyor environment, so it would need to be changed if Appveyor updates their environment. I believe having MSVC_VER set to 15 will never work with building OpenJFX on Windows.

bourgesl:
Sorry for the noise. 
I tried many things:
- 2nd test pass crashes vm
- upload failed due to missing test folders
- vs2017 changed and no more works

I propose to withdraw this PR and fix appveyor CI on my fork (or yours) for now

johanvos:
I couldn't get it to green either. @brcolow since you seem to be knowledgeable about building on Windows, can you try to modify the appveyor.yml file so that the tests pass?

brcolow:
I will try and devote some time to that in the next couple days. Sorry I can't do it before then.

brcolow:
Just had time to look at this very briefly. If you go back to the `fix NPE in getMediaName()` commit, remove `test_script`, and change `build_script` to `.\gradlew test [...]`, that may work. We may have just hit the build time limit, and nothing was actually broken about the appveyor script. Like I said, I'll look into this in more detail as soon as I can.

brcolow:
@johanvos Investigating in #24

bourgesl:
Thanks michael for fixing appveyor build in #24, my FXPrinter patch seems working.

I will rebase my fork on your latest changes and make tests soon

PR #24, opened 2018-03-16
Try and fix Appveyor.

brcolow:
@johanvos @bourgesl

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,32 +5,31 @@ image: Visual Studio 2017
 environment:
   JAVA_HOME: C:\\jdk9
   VCINSTALLDIR: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build
-  VS150COMNTOOLS: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build
-  VSVARS32FILE: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars32.bat
-  MSVC_VER: 14.12.25827
   APPVEYOR_CACHE_ENTRY_ZIP_ARGS: "-t7z -m0=lzma -mx=9"
 
 shallow_clone: true
-
-build:
-  verbosity: detailed
 
 build_script:
   - ps: |
       choco install jdk9 --version 9.0.4.11 -params 'installdir=C:\\jdk9'
       choco install gradle --version 4.3.0
+      $msvcToolsVer = Get-Content "$env:VCINSTALLDIR\Microsoft.VCToolsVersion.default.txt"
+      if ([string]::IsNullOrWhitespace($msvcToolsVer)) {
+        # The MSVC tools version file can have txt *or* props extension.
+        $msvcToolsVer = Get-Content "$env:VCINSTALLDIR\Microsoft.VCToolsVersion.default.props"
+      }
+      $env:MSVC_VER = $msvcToolsVer
+      $env:VS150COMNTOOLS = $env:VCINSTALLDIR
+      $env:VSVARS32FILE = "$env:VCINSTALLDIR\vcvars32.bat"
       refreshenv
-  - cmd: |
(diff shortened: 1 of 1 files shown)


brcolow:
Some info about MSVC versions: https://blogs.msdn.microsoft.com/vcblog/2017/03/06/finding-the-visual-c-compiler-tools-in-visual-studio-2017/ - we need what is referred to as the "MSVC Toolset Version".

brcolow:
I found a way to programmatically fetch the MSVC toolset version, and the new version after the latest Appveyor MSVC update is `14.13.26128`. I believe this is the only reason Appveyor was failing, but we will see. This will prevent that from happening after future Appveyor updates.

brcolow:
Okay so here's where we stand:

The reason the Appveyor build is now failing, even though all tests seemingly pass and the following message is printed:

```
BUILD SUCCESSFUL in 7m 56s
190 actionable tasks: 190 executed
Stopped 1 worker daemon(s).
```

the `gradle test` invocation in `build_script` is returning a non-zero exit code. So in this case it is not Appveyor's fault. This is confirmed by adding: `echo "Gradle exit code: $?" directly after the call to `gradle`. This prints: `Gradle exit code: False` which indicates the exit code in non-zero.

So the question is, why is gradle returning a non-zero exit code? Is something going wrong somewhere in the build? Would it be advisable to instead of rely on the exit code from gradle, somehow check manually if any tests failed and if so, only then return a non-zero exit code so that the build fails? Bypassing the exit code from gradle seems like a bad idea to me, and so this boils down to finding out the answer to the first question.

johanvos:
I prefer to fix this issue first before merging anything else. We want to see green badges with every single commit.
So back to this issue, the remaining problem is
1. gradle, although stating the build was successful, returns with a non-0 value
2. appveyor will consider this a failed build.

correct?

bourgesl:
I am testing now your latest changes and will try to diagnose:
see https://github.com/bourgesl/openjdk-jfx/tree/printer

bourgesl:
I found a potential issue:
- appveyor powershell interprets std.err outputs as failures !
See https://help.appveyor.com/discussions/problems/10014-false-build-fail-status

As gradle has compilation warnings (gradle --debug):
 [ERROR] [system.err] Note: Some input files use or override a deprecated API.
07:11:51.221
 [ERROR] [system.err] Note: Recompile with -Xlint:deprecation for details.
07:11:51.221
 [ERROR] [system.err] Note: Some input files use unchecked or unsafe operations.
07:11:51.221
 [ERROR] [system.err] Note: Recompile with -Xlint:unchecked for details.

Maybe gradle status code is related to compiler warnings ?

bourgesl:
I got a FIRST successfull build:
https://ci.appveyor.com/project/bourgesl/openjdk-jfx/build/printer%2026

Will propose another PR after ensuring it fails if either code build or test is broken

brcolow:
@bourgesl Trying to use your suggestion and call gradle using `cmd.exe`. I originally had it that way IIRC, but forgot why!

bourgesl:
I explicitly broke my build (printer_fail branch) and appveyor detected the build failure (test failed):
https://ci.appveyor.com/project/bourgesl/openjdk-jfx/build/printer_fail%2028

@brcolow Could you merge my appveyor.xml changes with yours to let you upgrade this PR ?
I prefer avoiding creating another PR for the same issue.

See: https://github.com/bourgesl/openjdk-jfx/blob/3c1f5c6cf06fc161031e7a15a26cee745d122471/appveyor.yml

I think CI is now working well and I should propose an openjfx patch for the J2DPrinter as well, isn't it @johanvos  ?

brcolow:
@bourgesl I made the necessary changes. It is now passing :). Thanks for catching the stderr issue!

bourgesl:
Excellent work, then.
Approved ?

johanvos:
Excellent!
For clarity, I think we need to create 2 PR's: one that fixes the Appveyor script, and another one that fixes the failing tests? The one with the failing tests should also go back to openjfx.

brcolow:
@johanvos I will drop the cherry-picked commit so that this PR only addresses Appveyor.

johanvos:
I tested this locally, with the NPE's fixed, and the test passed.

Issue #22, opened 2018-03-15
IDE specific files found in index


aalmiray:
The following files/dirs are found in the index

 * .idea
 * .classpath
 * .project

This means anyone working on either IntelliJ and/or Eclipse have a high chance of overwriting these settings and send a PR with invalid information. These files should be added to `.gitignore` (see #21) and removed from the index.

aalmiray:
There is ore than one location where `.classpath` and `.project` files are found. Also, a `netbeans` directory is found at the root.

eugener:
Duplicate of #4

PR #26, opened 2018-03-18
Add Gradle wrapper files. Relates to #23

aalmiray:
Signed OCA sent on March 15th, confirmation pending.

PR #28, opened 2018-03-19
Remove IDE files from index

aalmiray:
Fix for #22 

Netbeans dir is still found in the index as I suspect I may be needed. Need help from a  Netbeans savvy person to figure out if this is true or not.


hendrikebbers:
same as https://github.com/javafxports/openjdk-jfx/pull/28

hendrikebbers:
About netbeans:
I think the folders should be removed anyway. If they contain files that are needed for the general (gradle) build the files should be moved to general build folders

nlisker:
If you just remove the Eclipse files then someone who clones the repo wouldn't be able to work on it. The .project files specify the project folders and the .classpath files are necessary for compilation.

aalmiray:
@nlisker: IDE files are generated by the IDE and usually contain user specific settings. Opening up the project in Eclipse should be enough to re-generate the IDE files.

johanvos:
We need reports from people using NetBeans, IntelliJ, Eclipse to confirm that removing the files doesn't block them.
I'm using vi for my OpenJFX work, so I can't test.

aalmiray:
@johanvos Indeed. I can confirm that simply opening up the project in IntelliJ will create new files under `.idea` and override existing ones.

nlisker:
@aalmiray 

The .classpath contains module CL commands like --add-reads which are necessary to compile the projects on Eclipse specifically. How will Eclipse regenerate these files? Can you show the .classpath files for the base and graphics modules that Eclipse created from scratch?

The .project files contain info like project name and "natures". You can't open up a project without this file because there is no project, you would have to import it instead. Additionally, all projects use Gradle commands, so the Gradle project nature will have to be specified. Again, I don't understand how Eclipse knows this by itself. 

I might be missing something big here, but it seems to be the same discussion I had initially at https://bugs.openjdk.java.net/browse/JDK-8198795.

hendrikebbers:
+1 for removing the files. But we need to check that everything is working fine and add a short part of to the readme.md.
@aalmiray I tried to get it running in IntelliJ and had some problems:
- I had no permission to execute the `gradlew` file. Can you do a `chmod` and recommit pls.
- Once it is imported as gradle project in intelliJ the sources are not found. Even when I refresh the project in the gradle view nothing happens.
fx
- I can not start the build from shell. Using Java 10:

> > Task :controls:processResources 
> Error occurred during initialization of boot layer
> java.lang.module.FindException: Error reading module: /Users/hendrikebbers/git/openjdk-jfx/modules/javafx.base/build/classes/java/main/javafx.base
> Caused by: java.lang.module.InvalidModuleDescriptorException: Package javafx.beans.property not found in module
> 
Any idea?

So based on this I'm not able to build the current version or use it in the IDE (IntelliJ) :(

hendrikebbers:
mmh, looks like the gradle build error only happens in the IntelliJ terminal... Will check this later

hendrikebbers:
ok, `build` is not the right gradle target. That was my problem by building it from terminal. So terminal / shell is running but still no modules recognized in IntelliJ

hendrikebbers:
About netbeans folder: As far as I can say this folders include the ANT build files that are still needed. Wanted to do a rerun of the build and by calling the `clean` target the build crashes on my machine since I have not installed ANT 🙈
@aalmiray can you convert this to Gradle?? As much as I hate gradle it is much better than ANT 😎

aalmiray:
`gradlew` cannot be made executable by default (at the moment) due to some upstream restrictions (see https://github.com/javafxports/openjdk-jfx/pull/26#discussion_r175915770)

Would love to convert all Ant based builds to gradle but that'll take a long time.

kevinrushforth:
A couple of comments on this:

1) I added a basic .gitignore file in PR #60 -- since that file is not in the upstream repo, any additions to that file should be done as a separate PR.

2) There is a bug in JBS tracking this, [https://bugs.openjdk.java.net/browse/JDK-8198795](https://bugs.openjdk.java.net/browse/JDK-8198795). There are two related GitHub issues: #22 and #28. If we end up pursuing this fix, we will need a single issue in GitHub that addresses JDK-8198795 (including answering all all of the concerns).

3) Please change this PR to be based against the 'develop' branch rather than 'master'.

aalmiray:
I'm closing this PR as it provides mixed fixes (`.gitignore` file and IDE files). Also, the current problem are the `.idea` files in the index; Eclipse files and netbeans can stay.

PR #32, opened 2018-03-21
fix NPE in getMediaName()

bourgesl:
This patch was tested already in PR#24 to fix NPE in printer tests


bourgesl:
Thanks,  CI is now working !

PR #40, opened 2018-03-25
Gradle build failure should show informative message

abhinayagarwal:
Gradle build fails when one or more packages are not present on linux environment. Instead of throwing some useful information, the following message is shown:

```
FAILURE: Build failed with an exception.

* Where:
Script '~/Documents/openjdk-jfx/buildSrc/linux.gradle' line: 88

* What went wrong:
A problem occurred evaluating script.
> Cannot invoke method split() on null object
```

With this fix, when the build fails because of missing packages, a more informative message is shown.

```
FAILURE: Build failed with an exception.

* Where:
Script '~/Documents/openjdk-jfx/buildSrc/linux.gradle' line: 94

* What went wrong:
A problem occurred evaluating script.
> GTK2 development packages not found. If GTK2 packages are installed, please remove the build directory and try again.
```

### DIFF
--- a/buildSrc/linux.gradle
+++ b/buildSrc/linux.gradle
@@ -85,8 +85,14 @@ setupTools("linux_gtk2",
         propFile << "libsGTK2=" << results3.toString().trim()  << "\n";
     },
     { properties ->
-        gtk2CCFlags.addAll(properties.getProperty("cflagsGTK2").split(" "))
-        gtk2LinkFlags.addAll(properties.getProperty("libsGTK2").split(" "))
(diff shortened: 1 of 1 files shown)


johanvos:
+1 from me.
I had this error before as well, and a cryptic message about a NULL object doesn't inform the developer of the cause.

abhinayagarwal:
Hi Kevin,

I had signed and mailed the OCA on March 27th.

kevinrushforth:
OK, thanks.

PR #44, opened 2018-03-26
WIP: JDK-8186429 FXMLLoader setStaticLoad should be public API

abhinayagarwal:
This is useful for SceneBuilder to open FXML files with event handler attributes.  Currently, SceneBuilder for JDK 9+ uses reflection as a workaround.


abhinayagarwal:
There is already a [webrev](https://bugs.openjdk.java.net/browse/JDK-8186429?focusedCommentId=14112059) in the issue with potential fix. Does it make this PR redundant?

modules/javafx.fxml/src/test/java/test/javafx/fxml/FXMLLoader_ScriptTest.java
```diff
@@ -39,7 +38,7 @@
     @SuppressWarnings("deprecation")
     public void testStaticScriptLoad() throws IOException {
         FXMLLoader fxmlLoader = new FXMLLoader(getClass().getResource("static_script_load.fxml"));
-        FXMLLoaderHelper.setStaticLoad(fxmlLoader, true);
+        fxmlLoader.setStaticLoad(true);
```
hendrikebbers: Since this is only called 2 times directly after the constructor it would make sense to add this directly as a param to the constructor. By doing so the field could be defined as final

modules/javafx.fxml/src/test/java/test/javafx/fxml/FXMLLoader_ScriptTest.java
```diff
@@ -39,7 +38,7 @@
     @SuppressWarnings("deprecation")
     public void testStaticScriptLoad() throws IOException {
         FXMLLoader fxmlLoader = new FXMLLoader(getClass().getResource("static_script_load.fxml"));
```
hendrikebbers: `final FXMLLoader fxmlLoader`

modules/javafx.fxml/src/test/java/test/javafx/fxml/FXMLLoader_ScriptTest.java
abhinayagarwal: Sorry for the long gap :)

Are you trying to suggest to we can overload the constructor? Something similar to:

```
public FXMLLoader(boolean  staticLoad) {
    this.staticLoad = staticLoad;
}
```
Since FXMLoader already has a long list of constructors, do you think this would be a good idea?

kevinrushforth: I do not recommend adding any more constructor parameters.

If you want to take this new feature forward, it  will take a fair bit of work to specify what the effect of staticLoad should be on loading fxml files. It seems like a somewhat of a SceneBuilder-specific mode, but if it has more general use then it could be something to consider. Also, there will need to be several new tests to validate the behavior of various loading scenarios when using static load, including checking what the effects are of switching between modes for successive loads. It might seem simple to "just make this method public", but there is more to it than that.

I would also want @johanvos to weigh in, since Gluon now maintains SceneBuilder.

abhinayagarwal: Thank you for the pointers. 

I will try to find if this feature has use-cases outside of Scene Builder.

tobiasdiez:
The static load mode is also helpful if you want to use the FXMLLoader more as a parser. For example, we use it to verify that each fxml file in our project can be parsed successfully (as a unit test), and to extract all localized strings. In these cases, the static load mode is needed.

kevinrushforth:

NOTE: since this is a feature / enhancement request it needs to be discussed on the [openjfx-dev](mailto:openjfx-dev@openjdk.java.net) mailing list if you want it to be considered.

Issue #45, opened 2018-03-27
build failure on travis, qt can't be downloaded


johanvos:
Travis reports a build failure.

Looking in the logs shows this:
==> Downloading https://homebrew.bintray.com/bottles/qt-5.10.1.sierra.bottle.tar.gz
curl: (22) The requested URL returned error: 502 Bad Gateway
Error: Failed to download resource "qt"
Download failed: https://homebrew.bintray.com/bottles/qt-5.10.1.sierra.bottle.tar.gz

I manually tried to download that file (qt) and it succeeded. Hence, I assume this was a temporal network problem and the build should continue to green with the next build. 
I keep this issue open until the build is green.

ararunprasad:
Qt is no longer needed to build OpenJFX, Once it was required to build JFX along with WebKit.(i.e. -PCOMPILE_WEBKIT=true), but now OpenJFX WebKit uses cmake as a meta build system.

brcolow:
@ararunprasad It would be nice if you could go through the list of deps declared in the `.travis.yml`: https://github.com/javafxports/openjdk-jfx/blob/develop/.travis.yml#L6 and let us know if any are now unnecessary. I added them following the "Building OpenJFX" wiki.

ararunprasad:
@brcolow , Sure. I will raise a PR.

johanvos:
fixed by #48

PR #48, opened 2018-03-28
Qt-dev tools are no longer needed by OpenJFX

ararunprasad:
This changeset is related to #45. Qt is no longer needed to build OpenJFX, Once it was required to build JFX along with WebKit.(i.e. -PCOMPILE_WEBKIT=true), but now OpenJFX WebKit uses cmake as a meta build system.

### DIFF
--- a/.ci/before_install.sh
+++ b/.ci/before_install.sh
@@ -4,8 +4,6 @@ set -euo pipefail
 if [[ "${TRAVIS_OS_NAME}" == osx ]]; then
   brew update
   brew install findutils
-  brew install qt5
-  brew link qt5 --force
   brew cask reinstall java
   brew outdated gradle || brew upgrade gradle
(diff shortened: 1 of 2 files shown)


kevinrushforth:
This looks good to me. Do we still have any outdated Wiki pages or README files that refer to Qt? If so, we should fix them.

ararunprasad:
@kevinrushforth , I can't find any wiki/README page which mentions Qt dependency.

kevinrushforth:
@ararunprasad In that case this probably came from the [Building OpenJFX](https://wiki.openjdk.java.net/display/OpenJFX/Building+OpenJFX) Wiki page, which referenced Qt up until I fixed it a few months ago.

ararunprasad:
@kevinrushforth Most probably :)

PR #49, opened 2018-03-28
Additional updates needed for JDK-8199841 to add gradle wrapper files to build

kevinrushforth:
This PR updates the existing gradle wrapper files in the develop branch to match what I plan to send out for a review to integrate [JDK-8199841](https://bugs.openjdk.java.net/browse/JDK-8199841) into the openjfx/jfx-dev master, so that there will be no merge conflicts.

The changes are as follows:

1. Added the gradle.md license file
2. Replaced gradle-wrapper.jar with one generated from gradle 4.3 to match the version in the license file (we shouldn't need to update this again after this)
With the above changes, there should be no merge conflicts when JDK-8199841 is integrated to the upstream openjfx/jfx-dev repo.


kevinrushforth:
@aalmiray @johanvos  please take a look

johanvos:
+1 from me.
I would assume the gradle-wrapper.jar that is currently in the gradle directory is the exact same as the new one, as the existing one should also have been generated with 4.3 but I'm not sure about that.
Either way, CI still works which means the jars are correct.

kevinrushforth:
Thanks for looking at it. The existing gradle-wrapper.jar was generated with gradle 4.6 which is why I regenerated it with 4.3.

aalmiray:
LGTM :smile:

ararunprasad:
lgtm 👍

PR #57, opened 2018-04-04
Add script used by jenkins job to sync with openjfx mercurial repository

tiainen:
Add the script that is used by the jenkins job to synchronise the OpenJFX mercurial repository with the fork on github. This allows others to make changes to the synchronisation process.

### DIFF
--- a/.github/sync_upstream.sh
+++ b/.github/sync_upstream.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+echo "Enter hg"
+cd hg
+
+pushd openjfx/jfx-dev/rt/root
+echo "Update openjfx/jfx-dev/rt -> (root)"
+git hg fetch "http://hg.openjdk.java.net/openjfx/jfx-dev/rt"
+git hg pull "http://hg.openjdk.java.net/openjfx/jfx-dev/rt"
+popd
+
+echo "Exit hg"
+echo "Enter combined"
+
+cd ../combined
+
+echo "Check out master"
+git checkout master
+
+echo "Fetch (root)"
+git fetch "imports/openjfx/jfx-dev/rt/root"
+
+echo "Merge (root)"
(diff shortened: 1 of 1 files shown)


johanvos:
Unless there are objections, I will merge this in master. It doesn't have anything to do with the upstream code, so the .git directory is perfect for this.

kevinrushforth:
+1

eugener:
LGTM

.github/sync_upstream.sh
```diff
@@ -0,0 +1,39 @@
...
+cd ../combined
+
+echo "Check out master"
+git checkout master
+
+echo "Fetch (root)"
+git fetch "imports/openjfx/jfx-dev/rt/root"
+
+echo "Merge (root)"
+git merge "imports/openjfx/jfx-dev/rt/root/master" -m "Merge from (root)" --no-ff
```
brcolow: Removing `-no-ff` is a step in the right direction towards getting rid of the merge commits as mentioned by @kevinrushforth in #8 - the second step IIUC would be to use `git rebase` instead of `git merge`.

ararunprasad: I'm bit skeptical about using `git rebase`, I agree removing `-no-ff` would help to avoid the extra merge commit, but at the same time the merge candidate should be fast forwardable.

ararunprasad: Probably adding the following should remove the extra commit,
    git fetch github master
    git merge github/master

    echo "Merge (root)"
    git merge "imports/openjfx/jfx-dev/rt/root/master"

eugener: Don't you, guys, think that it would better to replace `-no-ff` with `--ff-only`? This way we will bail out if it can't fast-forward.

brcolow: I am not 100% certain that a rebase will never be required. If you guys are comfortable making that assumption, `--ff-only` would be fine.

PR #58, opened 2018-04-04
JDK-8195669 Cross Compile Arm Swing no longer fails

dellgreen:


dellgreen:
OCA has now been signed and excepted for Tyba Design Limited

johanvos:
Looks ok to me.
Maybe the preceding comment in build.gradle can now be removed as well:
```
    // Now that JMX is gone, we can likely uncomment the following as
    // a fix for JDK-8195669
```
Or do we want to keep that there for historical reasons? @kevinrushforth

kevinrushforth:
I would expect that comment should be removed, since its only purpose was to hint at the change that is now being made.

build.gradle
```diff
@@ -2158,13 +2158,11 @@ project(":controls") {
 }
 
 project(":swing") {
-    // Now that JMX is gone, we can likely uncomment the following as
```
johanvos: can you remove this comment?

johanvos:
also, can you rebase the PR to the current develop branch?

dellgreen:
thanks for doing this, I was flat out and didn't get a chance to follow up on your requests.

PR #61, opened 2018-04-05
JDK-8200446: Update minimum boot JDK to 10

kevinrushforth:
Fixes [JDK-8200446](https://bugs.openjdk.java.net/browse/JDK-8200446)

I'm doing it here first to make sure that the CI system for this sandbox is already using JDK 10 and thus won't be affected.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1684,7 +1684,7 @@ allprojects {
     // All of our projects are java projects
 
     apply plugin: "java"
-    sourceCompatibility = 1.9
+    sourceCompatibility = 10
 
     // By default all of our projects require junit for testing so we can just
     // setup this dependency here.
--- a/build.properties
+++ b/build.properties
@@ -72,5 +72,5 @@ javadoc.header=JavaFX&nbsp;11
 #
 ##############################################################################
 
-jfx.build.jdk.version.min=9
-jfx.build.jdk.buildnum.min=181
+jfx.build.jdk.version.min=10
+jfx.build.jdk.buildnum.min=46


kevinrushforth:
I see the following from the travis-ci build (and similar from appveyor) build:

>$ java -Xmx32m -version
>Picked up _JAVA_OPTIONS: -Xmx2048m -Xms512m
>java version "9.0.4"
>Java(TM) SE Runtime Environment (build 9.0.4+11)
>Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)
>...
>:verifyJava FAILED
>:verifyJava (Thread[main,5,main]) completed. Took 0.019 secs.
>FAILURE: Build failed with an exception.
>* Where:
>Build file '/home/travis/build/javafxports/openjdk-jfx/build.gradle' line: 250
>* What went wrong:
>Execution failed for task ':verifyJava'.
> FAIL: java version mismatch: JDK version (9.0.4) < minimum version (10)

And that's why I did the PR here first!

kevinrushforth:
I filed issue #62 to track updating the boot JDK used by the CI builds. That is a prerequisite for merging this PR.

kevinrushforth:
Once issue #62 is resolved and we get a successful CI build, I'll squash the commits into one.

kevinrushforth:
As mentioned in [this message](http://mail.openjdk.java.net/pipermail/openjfx-dev/2018-April/021715.html) on the openjfx-dev list, I would like to hold off on this request for a couple days, say, until Monday to see if there are other comments.

kevinrushforth:
I mis-diagnosed the problem with the CI builds. They are already doing a clean build, so that was never the issue.

The root cause of the problem is that this proposed fix bumps the generated class file format to that of JDK 10, and gradle 4.3 does not support the JDK 10 classfile format. This causes 'gradle test' to generate more than 7000 warnings of the form:

```
Malformed class file [XXX.class] found on compile classpath, which means that this class will cause a compile error if referenced in a source file. Gradle 5.0 will no longer allow malformed classes on compile classpath.
```

Running a test build with '--info --stacktrace' generates an 85 Mbyte log file. This causes the Travis CI build to fail (once it exceeds 4 Mbytes it will kill the build). The Appveyor build passes, but truncates the logfile, so that only about the last 3Mbytes are stored.

The short answer is that we need to update the version of gradle used by the CI builds. Since we will likely need to update to 4.8 once it is release (so we can build with JDK 11), it might be best to hold off on this PR until then. Or we could do it in two steps -- update to gradle 4.6 now and update to 4.8 when it released (after ensuring it will work with JDK 11) -- but it would be a bit more bookkeeping.

I will file a JBS issue for this, but will need to put this on hold for a couple of weeks due to other priorities.

kevinrushforth:
I filed [JDK-8201553](https://bugs.openjdk.java.net/browse/JDK-8201553) to track updating gradle.

kevinrushforth:
Note that this this will be unblocked after #92 is merged, which fixes [JDK-8201553](https://bugs.openjdk.java.net/browse/JDK-8201553). Once it is, I'll rebase this on top of the develop branch and then we can resume reviewing this.

johanvos:
Applying these changes after I applied the PR from #92 make the test work (locally on linux), so once #92 is merged, I'm +1 for merging this too. Travis/Appveyor are important checks for making sure it works on all platforms.

nlisker:
#92 has been merged a week ago. Is there any reason for not merging this PR yet?

kevinrushforth:
It should be ready to go in soon. I'll rebase the patch again and send a note to openjfx-dev today as a heads up that this is back on review.

PR #63, opened 2018-04-05
Use JDK 10 on Travis CI and Appveyor.

brcolow:


brcolow:
Well this is strange. I have the `https://github.com/javafxports/openjdk-jfx` remote added to my local repository as `upstream`. Doing `git fetch upstream master`, `git rebase upstream/master`, shows no conflicts when this branch in checked out.

kevinrushforth:
I looked at the log file, and the tests are failing because there is an extra '\' in the path. Due to a bug in gradle, the value of JAVA_HOME must exactly match the actual file system path (so no sym-links on Linux or Mac).

kevinrushforth:
The Travis CI build is failing to compile and run the tests because the JDK 10 it is using is an OpenJDK build without JavaFX modules (which is not yet supported for building and running apps and tests).

brcolow:
Whoops, I used the wrong argument for `install-jdk.sh`. Should be `-L BCL` not `-L GPL`. In terms of the PATH...I am not sure where that is being set incorrectly? Is it on [this line](https://github.com/sormuras/bach/blob/master/install-jdk.sh#L143) ?

kevinrushforth:
The JDK_HOME path problem is for the Windows Appveyor build.

The setting appears to be:

JAVA_HOME=C://jdk10

it should be:

JAVA_HOME=C:/jdk10

brcolow:
We will be able to use the usual `jdk: oraclejdk10` syntax when https://github.com/travis-ci/travis-ci/issues/9368 lands in non-staging - but this will work even after that, it's just a bit noisy.

PR #65, opened 2018-04-08
IntelliJ IDEA should use GIT instead of Mercurial

abhinayagarwal:
Until .idea directory is added to the ignore list, this helps IntellIJ IDEA to recognise the project as GIT based.

**NOTE:** This should not be pushed upstream to hg.openjdk.java.net/openjfx


kevinrushforth:
We currently have only a small number of files that exist in the GitHub branch that don't exist on the master OpenJDK hg repo. We have no files that have different content between the two, and I don't want to down that path, since it is a recipe for confusion and merge conflicts.

PR #67, opened 2018-04-15
Generate PDB (debugging files) and upload them via Appveyor, use them to get native backtrace.

brcolow:
The intent of this pull-request is to generate and then upload the debugging symbols (`*.pdb` and `*.map` files) generated during the Windows Appveyor build of OpenJFX. The files will be uploaded using [Appveyor's artifacts feature](https://www.appveyor.com/docs/packaging-artifacts/). Such debugging files are generated by switching the gradle `CONF` property from the default `Release` to `DebugNative`. These [`*.pdb` files](https://docs.microsoft.com/en-us/visualstudio/debugger/specify-symbol-dot-pdb-and-source-files-in-the-visual-studio-debugger) can be used to get symbol information when debugging a JVM crash on Windows. Going forward this could be useful whenever we touch native Windows code which will definitely need to be done in order to keep JavaFX modern.

This PR is a follow-up to the comment I put in `.appveyor.yml` about making the `cdb.exe` back-traces *more* useful. In fact, they were not useful at *all* because what was there didn't work. Now that I am working on https://github.com/javafxports/openjdk-jfx/issues/66 this gave me the motivation to try and actually implement this functionality.

You can see the uploaded artifact by clicking on the [Appveyor "Artifacts" tab](https://ci.appveyor.com/project/javafxports-github-bot/openjdk-jfx/build/develop%20223/artifacts) on the Appveyor build page of this PR. I also took this opportunity to upload the JavaFX modules which will be especially useful when it will be possible to "overlay" those modules on top of the JDK.

At least one thing that is still WIP is figuring out how we can get PDB files for the JDK itself for whatever Java version that Appveyor build is using. I temporarily and for testing purposes only used the PDB files published by https://github.com/ojdkbuild/ojdkbuild.

Update: It appears that the builds offered by [AdoptOpenJDK](https://adoptopenjdk.net/) come with PDB files. I will most likely use their API to fetch the Java symbols.

Update 2: I think it makes sense to do this is a future PR as this PR addresses the most common use-case where the crash occurs in the JavaFX native windows code.

### DIFF
--- a/.ci/script.sh
+++ b/.ci/script.sh
@@ -16,7 +16,6 @@ if [[ ${RESULT} -ne 0 ]]; then
   if [ ! -z ${PRINT_CRASH_LOGS+x} ]; then
     if [[ "${TRAVIS_OS_NAME}" == osx ]]; then FIND="gfind"; else FIND="find"; fi
     ${FIND} . -name "hs_err_pid*.log" -type f -printf '\n====== JVM CRASH LOG ======\n%p\n' -exec cat {} \;
-
     CORES=''
     if [[ "${TRAVIS_OS_NAME}" == osx ]]; then
       CORES="$(find /cores -type f -regex '*core.[0-9]{4}' -print)"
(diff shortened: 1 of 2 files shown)


brcolow:
Working on not using choco and instead installing the JDK manually. This is back to WIP until I get that right.

brcolow:
Actually, that is going to cause more problems than it's worth down the line, I went back to Choco. Our use-case is simpler than TestFX which has to install multiple JDKs.

brcolow:
I think this is rather low-risk to merge and improves upon what we have now.

kevinrushforth:
I don't have any way to test this, but it looks OK to me.

Since this patch is limited to modifying files that exist only on github, it might be better to merge it directly to the master branch. @eugener or @johanvos can comment on that.

eugener:
This looks ok to me, even though it is pretty difficult to test. 
I think it is safe enough to merge, as long as @johanvos is ok with it too.

johanvos:
The symbolification can clearly help, so +1 from me.
I agree with @kevinrushforth that we should do it in the master branch. Develop branch will pick it up (at least I think so, @tiainen ?) and since it is not intended to go to the OpenJFX repository, this is the way to do it.

tiainen:
@johanvos yes, develop branch is kept in sync with master (https://github.com/javafxports/openjdk-jfx/blob/a5e91cedf31e62215a040bee1183554997eaae8f/.github/sync_upstream.sh#L29)

PR #74, opened 2018-04-30
(travis) Set CONF gradle property to 'DebugNative'.

brcolow:


kevinrushforth:
This is (at least somewhat) related to #67 but for the Travis CI builds. Since this only affects the CI builds used for testing, this seems fine to me. As with the fix for #67 it can go straight into master since it is limited to modifying files that exist only on github,

johanvos:
Related to #67 which has just been merged into master. I'll merge this one too into master.

Issue #77, opened 2018-05-09
JDK-8203345: Memory leak in VirtualFlow when screen reader is enabled


belu:
In class VirtualFlow there exists a private method `getPrivateCell(int)` which at the end could add a new cell to an array called `privateCells`. Elements from this array are never removed.
An element could be e. g. a TableRow (via TableView) which could use hundreds of KB of memory. Therefore if lots of such elements are added this could lead to an OutOfMemory error.

The call hierarchy of this method shows that the method `getPrivateCell(int)` is only called via the method `queryAccessibleAttribute(AccessibleAttribute, Object...)`. This could imply that this issue is only relevant when Accessability is turned on.

According to the source code this could affect the controls TableView, TreeTableView and ListView.

belu:
As a workaround one can set the system property `glass.accessible.force` to `false`.
This turns off Accessability and therefore does not cause the memory leak. Note that Accessability is turned on by default only on Mac 10.9 and Windows 7 or greater according to notes in `com.sun.glass.ui.View`.

belu:
Sample application to reproduce the issue can be found [here](https://gist.github.com/belu/aacfe33e619b1727b937eabba2abe61f).
Tested with OpenJDK 10.0.1 and OpenJFX 11 EA.

When the application is running, start a screen reader (e. g. narrator under Windows 10). When Accessibility is turned on this is displayed in the console. Use a tool like [visualvm](https://visualvm.github.io/) to monitor the number of TableRow's.

When you click "Add" once you will see around 7-8 rows being allocated (about the number of visible rows). When you continue to click on "Add" and the first row leaves the visible part then new TableRow's will be created. However these TableRow's are not released as long as the table is referenced - even when you clear all items in the table.

If you start the application with the workaround turned on (via a boolean flag at the top of the source file) and keep on adding rows, then just the number of visible rows will be allocated.

PR #78, opened 2018-05-17
JDK-8203345: Memory leak in VirtualFlow when screen reader is enabled

belu:
This PR fixes issue #77

### DIFF
--- a/tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
+++ b/tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
@@ -1,3 +1,28 @@
+/*
+ * Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
 package test.javafx.accessibility.virtualflow;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
--- a/tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
+++ b/tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
@@ -197,4 +197,3 @@ public class VirtualFlowMemoryLeakTest {
         Util.runAndWait(runnable);
     }
 }
-
--- a/tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
+++ b/tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
@@ -62,15 +62,6 @@ public class VirtualFlowMemoryLeakTest {
     private static CountDownLatch startupLatch;
     private static CountDownLatch screenReaderLatch = new CountDownLatch(1);
 
-    public static void main(final String[] args) throws Exception {
-        initFX();
-        try {
-            new VirtualFlowMemoryLeakTest().test_JDK8203345();
-        } finally {
-            teardown();
-        }
-    }
-
     public static class TestApp extends Application {
         @Override
         public void start(final Stage stage) throws Exception {
--- a/tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
+++ b/tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
@@ -167,6 +167,8 @@ public class VirtualFlowMemoryLeakTest {
 
     @Test
     public void test_JDK8203345() throws Exception {
+        assumeTrue(PlatformUtil.isMac() || PlatformUtil.isWindows());
+
         final boolean screenReaderPresent = screenReaderLatch.await(5_000, MILLISECONDS);
         assumeTrue(screenReaderPresent);
 
--- a/tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
+++ b/tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
@@ -1,0 +1,173 @@
+package test.javafx.accessibility.virtualflow;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+import static org.junit.Assume.assumeTrue;
+
+import java.lang.ref.WeakReference;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import javafx.application.Application;
+import javafx.application.Platform;
+import javafx.beans.property.ReadOnlyObjectProperty;
+import javafx.beans.property.SimpleObjectProperty;
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
+import javafx.scene.Parent;
+import javafx.scene.Scene;
+import javafx.scene.control.TableColumn;
+import javafx.scene.control.TableRow;
+import javafx.scene.control.TableView;
+import javafx.scene.control.TableView.TableViewSelectionModel;
+import javafx.stage.Stage;
+import test.util.Util;
+
+public class VirtualFlowMemoryLeakTest {
+
+    private static CountDownLatch startupLatch;
+    private static CountDownLatch screenReaderLatch = new CountDownLatch(1);
+
+    public static void main(final String[] args) throws Exception {
+        initFX();
+        try {
+            new VirtualFlowMemoryLeakTest().test_JDK8203345();
+        } finally {
+            teardown();
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


kevinrushforth:
At first glance, the fix to clear the privateCells list seems fine. Testing will be the challenge. I think there are two possible approaches:

1. Fully automated test -- this has the challenge that a screen reader, such as VoiceOver on Mac or Narrator on Windows, needs to be enabled, which may not be possible (or desirable, even if it is possible) from the test program or test harness. My guess is that this is somewhere between infeasible and not worth the effort.

2. Semi-automated test -- an automated test that only does useful testing if a screen reader is turned on. It can run as part of our regular test suite, which might be OK as long as we occasionally run the test suite with the screen reader turned on (which might be interesting in its own right to see whether there are any regressions in existing tests).

3. Manual test -- we generally do this only if there is no good way to test it in an automated manner, so I recommend at least spending a little bit of time writing an automated test (to run in a semi-automated mode).

I presume the cells being held by the privateCells list are ones that the application creates? If so, then testing it should be easy. The test program can do whatever is needed to cause the leak (presuming that a screen reader is on) -- that is, to cause the cell in question to show up in the privateCells list; store a weak reference to that cell, and then make sure that any hard refs in the test program are cleared or go out of scope. Then it is a matter of forcing gc and checking to see if the reference is null. Do that in a loop with a sleep a few time, and if the ref never goes to null you have a leak.

See [TooltipFXTest.java](https://github.com/javafxports/openjdk-jfx/blob/c28bf083925ed79f7a88c531b28e191e5fd828a5/tests/system/src/test/java/test/robot/javafx/web/TooltipFXTest.java), which was just committed earlier today, for an example of this.

kevinrushforth:
@premyi can you take a look at this and help test / evaluate the fix?

kevinrushforth:
Note that we also need to wait for your OCA to be processed before acting on this PR.

belu:
@kevinrushforth @premyi 
Thanks for the example test case. That helps a lot in setting up a semi-automatic test for this memory leak.

What I will do in the test is to use a CountDownLatch that waits a few seconds for a screen reader. Only if a screen reader is found the test will be performed.

premyi:
proposed fix looks good to me.

Observations: (with latest code, on Windows 10, screen reader: Narrator)
List privateCells size does not get incremented every time when a new entry is added to the table.
It get incremented once initially and when the table grows more than the window size it gets incremented with every new entry added to the table.

I am also getting null pointer exception intermittently when the table grows more than the window size,
did you get similar exception???
Exception in thread "JavaFX Application Thread" java.lang.NullPointerException
	at javafx.graphics/com.sun.glass.ui.win.WinAccessible.GetPatternProvider(WinAccessible.java:570)
	at javafx.graphics/com.sun.glass.ui.win.WinApplication._runLoop(Native Method)
	at javafx.graphics/com.sun.glass.ui.win.WinApplication.lambda$runLoop$3(WinApplication.java:175)
	at java.base/java.lang.Thread.run(Thread.java:831)

belu:
@premyi 
Yes, these Exception occured also for me. It does not seem to be related to the give fix as this also happens without the fix (e. g. with Java 8). That would be a separate issue to look into.

belu:
@kevinrushforth @premyi 
What would be the right location for a memory leak test for VirtualFlow?

premyi:
FYI: i have created a bug to handle NullPointerException JDK-8203463.

It would be good if we can keep all accessibility related tests separately and i would suggest to create a new directory "Accessibility" in the following path 
"rt/tests/system/src/test/java/test/javafx".(with sub folders).

i.e., "rt/tests/system/src/test/java/test/javafx/Accessibility/VirtualFlow"

belu:
@premyi 
Preferably with lower-case folder names as the others?

premyi:
Yes

belu:
@premyi 
Fyi, have commited a test case. Please review and verify.
Unfortunately I am not able to build and run the tests via gradle under Windows (only Linux). However under Windows I ran the test against the unpatched and patched javax.controls module to very failure/success of the test (and also a test without a screen reader to verify that the test is marked as skipped).

Explanation of what the test case does:
The test waits for up to 5 seconds for a screen reader. If none is present the test will be marked as skipped.

The test then initially adds one item and clears the list. This should trigger the creation of all required TableRows (depending on how much rows are visible).
The next row that is added now is kept in a WeakReference. That is the one we want to verify that is is GC'ed. We then add a few rows so that the weakly referenced row disappears from the visible part of the TableView (new rows are added at the top).
Then a few GCs are triggered to verify that the TableRow from the WeakReference disappears.

kevinrushforth:
I note that the [OCA](http://www.oracle.com/technetwork/community/oca-486395.html#l) for @belu Bernhard Lutzmann has been recorded.

Once the review is finished (mainly the review of the test at this point), this PR can be merged.

premyi:
On Windows 10, test works as expected.
On macOS (Version 10.13.1) it crashes on application shutdown, this looks like known issue JDK-8113976. 
@kevinrushforth can you take a look?

Java has been detached already, but someone is still trying to use it at -[GlassAccessible accessibilityAttributeValue:]:/Users/AGHAISAS/prem/jfx-dev/rt/modules/javafx.graphics/src/main/native-glass/mac/GlassAccessible.m:74
0   libglass.dylib                      0x000000012d64f4cb -[GlassAccessible accessibilityAttributeValue:] + 123
1   AppKit                              0x00007fff2a9ee523 NSAccessibilityGetObjectForAttributeUsingLegacyAPI + 371
2   AppKit                              0x00007fff2af9fe37 NSAccessibilityPerformEntryPointObject.llvm.734619A0 + 16
3   AppKit                              0x00007fff2afa0003 _NSAccessibilityEntryPointValueForAttribute.llvm.734619A0 + 182
4   AppKit                              0x00007fff2ad47811 -[NSObject(NSAccessibilityInternal) _accessibilityValueForAttribute:clientError:] + 351
5   AppKit                              0x00007fff2ad4e474 CopyAppKitUIElementAttributeValueNoCatch + 98
6   AppKit                              0x00007fff2ad4c0ae CopyAttributeValue + 172
7   HIServices                          0x00007fff2bc4c231 _AXXMIGCopyAttributeValue + 282
8   HIServices                          0x00007fff2bc561ac _XCopyAttributeValue + 455
9   HIServices                          0x00007fff2bc2e939 mshMIGPerform + 212
10  CoreFoundation                      0x00007fff2d356219 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE1_PERFORM_FUNCTION__ + 41
11  CoreFoundation                      0x00007fff2d356165 __CFRunLoopDoSource1 + 533
12  CoreFoundation                      0x00007fff2d34dd50 __CFRunLoopRun + 2848
13  CoreFoundation                      0x00007fff2d34cfa3 CFRunLoopRunSpecific + 483
14  java                                0x000000010459fbc3 CreateExecutionEnvironment + 401
15  java                                0x000000010459bebc JLI_Launch + 1568
16  java                                0x000000010459ad4c main + 384
17  java                                0x000000010459abc4 start + 52

belu:
@kevinrushforth Did you had a chance to take a look at the issue with macOS?

kevinrushforth:
Not yet, sorry. I'll look at it this week.

kevinrushforth:
I can reproduce the crash on exit on the Mac, although if I run the unit test it only crashes for me without the fix. When I run it with the fix it doesn't crash (although I might be just getting lucky).

Let's not hold up getting this fix in to resolve this. If the test case is crashing with the fix, then we should probably mark the test as unstable and file a new bug for the failure (the one Prem pointed to above is resolved / fixed a long time ago, so while the symptoms might be the same, we should treat this as a new bug)

@premyi please file a new bug for this failure. You can link it to the one you mentioned above. Also, can you comment as to whether you see the crash on exit with the patch for this bug applied?

tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
```diff
@@ -0,0 +1,173 @@
+package test.javafx.accessibility.virtualflow;
```
kevinrushforth: You need to add a copyright header (you can copy it from one of the other test program or from VirtualFlow.java, with just `2018,` as the single year instead of a pair of years.

tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
```diff
@@ -0,0 +1,173 @@
...
+            MILLISECONDS.sleep(100);
+        }
+        assertEquals(null, firstRowRef.get());
+    }
+
+    private void runAndWait(final Runnable runnable) {
+        Util.runAndWait(runnable);
+    }
+}
+
```
kevinrushforth: Since you need to change this file anyway (to add the copyright), can you also delete the extra blank line at the end of the file?

kevinrushforth:
@premyi - thanks for filing the new bug.

tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    @AfterClass
+    public static void teardown() {
+        Platform.exit();
+    }
+
+    @Test
+    public void test_JDK8203345() throws Exception {
```
kevinrushforth: Please also add the following, since this test is only applicable on Windows and Mac (no need to waste 5 seconds on a test that will not ever run on Linux).

```
    assumeTrue(PlatformUtil.isMac() || PlatformUtil.isWindows());
```

tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
belu: Done.

tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
belu: Added copyright header and platform check. Furthermore removed the main method as requested by @premyi

tests/system/src/test/java/test/javafx/accessibility/virtualflow/VirtualFlowMemoryLeakTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+        tableView.getColumns().add(idColumn);
+        return tableView;
+    }
+
+    @BeforeClass
+    public static void initFX() {
+        startupLatch = new CountDownLatch(1);
+        new Thread(() -> Application.launch(TestApp.class, (String[]) null)).start();
+        try {
```
ararunprasad: It is better to replace the bunch of if .. try .. catch with assertions.

```
public static void initFX() throws InterruptedException {
    ..
    ..
    assertTrue("Timeout waiting for FX runtime to start", startupLatch.await(10, SECONDS));
}
```

kevinrushforth: Yes, that does seem cleaner, but I note that we have several other tests that do it this way (and I'm sure this logic was copied from one of them), so I would recommend to address this as part of a larger cleanup to unify the startup of system tests that need to launch an FX Application (e.g., some of newer tests wait for a Stage shown event).

kevinrushforth: I note that as part of the review for [JDK-8199614](https://bugs.openjdk.java.net/browse/JDK-8199614) I suggested to file a similar follow-up issue to unify the startup for tests, but it looks like the bug didn't get filed. I just filed [JDK-8206430](https://bugs.openjdk.java.net/browse/JDK-8206430) to track this.

kevinrushforth:
While getting ready to merge this, I noticed that the PR is against the wrong branch on javafxports. Bug fixed need to be merged into the `develop` branch of javafxports, not the master branch.

I will change it now, and wait for a test build (there should not be any conflicts).

PR #82, opened 2018-05-23
Use correct status badge for appveyor

tiainen:
This PR fixes the status badge for appveyor to point to the correct URL for loading the svg image.


johanvos:
+1. This is a fix in the badge only, and it doesn't go upstream.

eugener:
Looks good to me too.

PR #83, opened 2018-05-23
Build standalone modules (jars) to be uploaded to remote repositories

johanvos:
In this PR, we deal with the work to be done in order to upload artifacts to maven central and similar repositories. 
While jmods already contain the native libraries, they are not in the runtime picture. The current plan is to create jars with native code, and upload those. 
The jars-with-native-code should not overwrite the jars used in the standalone SDK (without native libs)


saudet:
BTW, I've already given this much thought as part of the JavaCPP Presets. It probably doesn't fit JavaFX like a glove, since JavaFX uses Gradle instead of Maven for one thing, but I'm sure a lot of ideas could be useful: https://github.com/bytedeco/javacpp-presets

I would also be very interested in a `gradle-javacpp` plugin, and if we can consolidate all the work necessary for JavaFX inside such a plug-in, it would automatically become available for Deeplearning4j and the projects that are part of the presets (FFmpeg, OpenCV, TensorFlow, etc), including for platforms like iOS!

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -29,14 +29,21 @@
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.HashSet;
+import java.io.*;
+import java.nio.*;
+import java.nio.file.*;
```
kevinrushforth: Can you expand these (we generally avoid wildcard imports)

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -29,14 +29,21 @@
...
 public class NativeLibLoader {
 
     private static final HashSet<String> loaded = new HashSet<String>();
+    private static final String tmpdir = System.getProperty("java.io.tmpdir");
 
     public static synchronized void loadLibrary(String libname) {
         if (!loaded.contains(libname)) {
-            loadLibraryInternal(libname);
+            StackWalker walker = AccessController.doPrivileged((PrivilegedAction<StackWalker>) () ->
+            StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE));
```
kevinrushforth: For this method youdon't need `AccessController.doPrivileged` since the caller needs to assert privileged before calling. It isn't harmful, though.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -150,6 +161,29 @@ private static void loadLibraryInternal(String libraryName) {
...
     }
 
+   /**
+    * If there is a library with the platform-correct name at the
+    * root of the resources in this jar, use that.
+    */
+    private static boolean loadLibraryFromResource(String libraryName, Class caller) {
+        try {
+            String reallib = "/"+libPrefix+libraryName+libSuffix;
+            InputStream is = caller.getResourceAsStream(reallib);
```
kevinrushforth: This looks like a good approach. It might require a few more internal qualified exports from calling modules to javafx.graphics, but that won't be a problem.

saudet:
> The easiest approach might be to have one platform-specific .jar file per module per platform, and a platform-independent .pom file per module that depends on the platform-specific .jar file for that platform.

@kevinrushforth Yes, that's exactly how JavaCPP does it. Could you let me know about the reasons why using existing tools such as JavaCPP, among others, isn't being considered?

kevinrushforth:
One reason is that we have policies on using third-party code. We do that when there is no viable alternative, but otherwise try to avoid additional third-party dependencies.

On the technical front, the bulk of the changes will be in the custom build.gradle logic anyway, which is specific to FX. Finally, we already have a NativeLibLoader method that is used in several places in three different modules. Modifying that one utility method is straight-forward.

Johan can chime in if he has additional insights.

saudet:
Well, I get that it's not a lot of code to maintain, but it is quite tricky and you will encounter a lot of issues...

BTW, what about including this in the JDK? Why hasn't this been done before?

johanvos:
I really like JavaCPP, but I see this as a slightlyt different usecase. I use JavaCPP to access existing native libraries via a X-platform Java interface.
In this case, we talk about native libraries we created ourselves, that only contain a very limited number of API's whose signature might even be platform-dependent. The amount of jni code is very limited here, and that's not a problem. Hence, generating the Java code is not needed here.

The concept of the Loader is very relevant here, so we could use it or similar. 

Bringing this in the OpenJDK project sounds very good to me, and there are procedures for doing so (e.g. creating a JEP).

saudet:
Yeah, I know about the JEP, but until I understand why this isn't something that people want, I don't feel inclined to spend my time writing that kind of thing.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
johanvos: done

johanvos:
The latest changes will build a jar containing classes + libs for each module, for each compileTarget. 
The name of the jarfile contains the compileTarget as classifier (e.g. javafx.graphics-linux.jar). The pom files contain classifiers in their dependencies as well (e.g. javafx.controls-linux.jar depend on javafx.graphics-linux.jar). This allows maven to resolve transitive dependencies.

Each jar has a module-info where the name of the module does not contain a compileTarget classifier. 

I tested this in 2 ways:
1) with a pom.xml, where I have a single dependency on openjfx:javafx.controls-linux:11.00-SNAPSHOT
2) with command-line invocation, with --module-path pointing to the location of the jar files and --add-modules javafx.controls 

Both approaches work.

kevinrushforth:
I confirm that the general approach works. A couple general comments:

1. Media does not work. The libgstreamer-lite.so, libfxplugin.so and the libavplugin-*.so files are not loaded by Java code, but by native code that expects to find them in the same directory as libjfxmedia.so was loaded from. If I manually unzip the files into /tmp then it works. We might need a new method to NativeLibLoader (e.g., "cacheLibaray" or similar) that the Media module can call on its list of libraries that are not loaded via Java.

2. In order for more than one app to run concurrently, we will either need some form of caching with locking mechanism so they can share the downloaded .jar files or else we will need a unique directory for each invocation. If we use a caching approach, then the directory will need to be versioned (or you will need checksums or similar). Also we will likely need a per-user directory rather than /tmp/ or we could run into filesystem permission problems. Given the complexity, it might be better to go with a unique directory name for each invocation. You will need to think about cleaning it up.

kevinrushforth:
Also for debugging, it would be handy if a "loaded from ...." message were printed in the maven case when the "javafx.verbose" property is set.

saudet:
I understand that you can't use JavaCPP directly, but you could take a look
at what it's doing and replicate that to save you a lot of trouble...

johanvos:
Samuel has a good point there, the JavaCPP Loader already deals with a list of preloaded libraries that need to be loaded. It would be nice to have something generic in Java, for now I'll add it to the NativeLibLoader.

johanvos:
Libraries are now stored in a cache in the user directory and are version dependent. TODO: don't replace if checksum is the same.
I added a loadLibary method that takes a List<String> containing the required dependencies for a particular native library. This is platform-dependent.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -150,6 +173,60 @@ private static void loadLibraryInternal(String libraryName) {
...
+   /**
+    * If there is a library with the platform-correct name at the
+    * root of the resources in this jar, use that.
+    */
+    private static boolean loadLibraryFromResource(String libraryName, List<String> dependencies, Class caller) {
+        try {
+            // first preload dependencies
+            if (dependencies != null) {
+                for (String dep: dependencies) {
+                    loadLibraryFromResource(dep, null, caller);
```
kevinrushforth: The mechanism generally looks like it works. However, I think should just unpack the dependencies and not load them. The media code loads only one of the N libavplugin-xxxx.so libraries, and does so from native code.

build.gradle
```diff
@@ -296,6 +296,9 @@ ext.IS_MAC = OS_NAME.contains("mac") || OS_NAME.contains("darwin")
 ext.IS_WINDOWS = OS_NAME.contains("windows")
 ext.IS_LINUX = OS_NAME.contains("linux")
 
+ext.MAVEN_GROUP_ID = "openjfx"
```
kevinrushforth: Should now be org.openjfx I think.

modules/javafx.media/src/main/java/com/sun/media/jfxmediaimpl/NativeMediaManager.java
```diff
@@ -104,15 +105,18 @@ protected NativeMediaManager() {
...
             AccessController.doPrivileged((PrivilegedExceptionAction) () -> {
+                ArrayList<String> dependencies = new ArrayList<>();
                 if (HostUtils.isWindows() || HostUtils.isMacOSX()) {
                     NativeLibLoader.loadLibrary("glib-lite");
                 }
 
                 if (!HostUtils.isLinux() && !HostUtils.isIOS()) {
                     NativeLibLoader.loadLibrary("gstreamer-lite");
+                } else {
+                    dependencies.add("gstreamer-lite");
```
kevinrushforth: For Linux, the following are also needed as dependencies:

```
                    if (HostUtils.isLinux()) {
                        dependencies.add("fxplugins");
                        dependencies.add("avplugin-54");
                        dependencies.add("avplugin-56");
                        dependencies.add("avplugin-57");
                        dependencies.add("avplugin-ffmpeg-56");
                        dependencies.add("avplugin-ffmpeg-57");
                    }
```

With the addition of the above, I can run media (although I get warnings...see my comment in NativeLibLoader).

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -150,6 +173,60 @@ private static void loadLibraryInternal(String libraryName) {
...
+                throw new IOException ("Cache exists but is not a directory: "+cacheDir);
+            }
+        } else {
+            if (!cacheDir.mkdirs()) {
+                throw new IOException ("Can not create cache at "+cacheDir);
+            }
+        }
+        // we have a cache directory. Add the file here
+        File f = new File(cacheDir, name);
+        // if it exists, remove. Todo: check checksum and keep if same.
```
kevinrushforth: Yes, this will be needed in order to run multiple applications at the same time.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
saudet: If media code relies on the search path, not preloading them might not work on Windows because it doesn't support rpath.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
saudet: If the filenames of different versions are the same, they need to be extracted in different directories though.

modules/javafx.media/src/main/java/com/sun/media/jfxmediaimpl/NativeMediaManager.java
saudet: I don't see them in NativeLibLoader.java. What warnings would that be?

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
johanvos: Directories are already different for different versions: String userCache = System.getProperty("user.home") + "/.openjfx/cache/" + jfxVersion;

However, when 2 applications using the same version are executed together, they will share the same directory. Libraries should never differ between versions, so I start to wonder if a checksum is needed at all. If the checksum of the "new" library is different from the installed one, there is an issue. This should not happen, and it indicates that either the old install is tampered with, or the new resource is corrupt. Both causes are evil, and I don't know what the best generic solution would be?

saudet: Yes, if you enforce it with version numbers, that should work fine too. That's what SWT does.

saudet: But that doesn't work when running custom builds during development...

johanvos: in that case, it makes sense to use the one from the resource, and removing the old one. I'll do that.

modules/javafx.media/src/main/java/com/sun/media/jfxmediaimpl/NativeMediaManager.java
johanvos: I added a boolean "load" to the loadLibraryFromResource method, which is true by default, but false when loading dependencies. In that case, the resource is simply unpacked.

johanvos:
I added checksum checks. For the file checksum, I used code that is also in the CssStyleHelper in order to be consistent with how it is done elsewhere in the code.
It will still fail in this case: an app is running, a new local build is created with the same version, and a new app is launched. In this case, the library for the running app will be deleted, causing that app to crash. But I think this is totally acceptable (it is similar to removing Java libraries while running Java applications in the exact same version)

johanvos:
I only checked on Linux so far, hence I don't know if it will fail on Windows, in case the dependencies are not preloaded.

kevinrushforth:
> I only checked on Linux so far, hence I don't know if it will fail on Windows, in case the dependencies are not preloaded.

I will check on Mac and Windows. There shouldn't be any dependencies on those platforms, though.

kevinrushforth:
It turns out there are dependencies on both Mac and Windows as well.

Also, on Windows I ran into two other problems with loading libraries. One is that the javafx.version problem is not valid early enough -- causing the Microsoft Windows DLLs to be unpacked in the cache/versionless directory. The second is that some of the Microsoft files are missing (not unpacked anywhere).

Everything works on both platforms if I manually copy all of the libs, so that proves the concept. We just need to add the dependencies for both platforms, resolve the version number issue (that should be easy to set the version property before the code that loads the libraries), and figure out why some of the libraries are missing.

johanvos:
I added the media libraries that are built for MacOSX as dependencies.
I don't have immediate access to a Windows machine, but I think the approach can be similar: once the libraries that are otherwise in the JavaFX SDK are in the "cache directory", loading libs should behave the same as in the case of the JavaFX SDK?

kevinrushforth:
Yes, once everything ends up in the caches directory, it will work.

Btw, one more issue I noticed. This patch causes a deprecation warning from gradle:

```
$ gradle --info --warning-mode=all
...
As part of making the publishing plugins stable, the 'deferred configurable' behavior of the 'publishing {}' block is now deprecated. Please add 'enableFeaturePreview('STABLE_PUBLISHING')' to your settings file and do a test run by publishing to a local repository. If all artifacts are published as expected, there is nothing else to do. If the published artifacts change unexpectedly, please see the migration guide for more details: https://docs.gradle.org/4.8/userguide/publishing_maven.html#publishing_maven:deferred_configuration. In Gradle 5.0 the flag will be removed and the new behavior will become the default.
	at build_60z30efsfsqoyqyd9no2utm1d.addMavenPublication(/localhome/kcr/javafx/github/johanvos-jfx/build.gradle:1521)
	(Run with --stacktrace to get the full stack trace of this deprecation warning.)
```

modules/javafx.media/src/main/java/com/sun/media/jfxmediaimpl/NativeMediaManager.java
```diff
@@ -124,6 +124,11 @@ protected NativeMediaManager() {
                     dependencies.add("avplugin-ffmpeg-56");
                     dependencies.add("avplugin-ffmpeg-57");
                 }
+                if (HostUtils.isMacOSX()) {
+                    dependencies.add("fxplugins");
+                    dependencies.add("glit-lite");
```
kevinrushforth: should be `glib-lite`

modules/javafx.media/src/main/java/com/sun/media/jfxmediaimpl/NativeMediaManager.java
```diff
@@ -124,6 +124,11 @@ protected NativeMediaManager() {
...
+                    dependencies.add("jfxmedia_av");
```
kevinrushforth: should be `jfxmedia_avf`

modules/javafx.media/src/main/java/com/sun/media/jfxmediaimpl/NativeMediaManager.java
```diff
@@ -124,6 +124,11 @@ protected NativeMediaManager() {
...
+                }
```
kevinrushforth: Windows will also need `fxplugins` and `glib-lite`

kevinrushforth:
The fix for the problem of the Microsoft libraries being loaded before `javafx.version` is set, should be as simple as swapping the order of the following two blocks:

```
        // This loading of msvcp140.dll and vcruntime140.dll (VS2017) is required on Windows platforms
        if (PlatformUtil.isWindows()) {
            loadMSWindowsLibraries();
        }

        AccessController.doPrivileged((PrivilegedAction<Object>) () -> {
            // Get the javafx.version and javafx.runtime.version from a preconstructed
            // java class, VersionInfo, created at build time.
            VersionInfo.setupSystemProperties();
            return null;
        });
```

build.gradle
```diff
@@ -1535,31 +1573,24 @@ void addMavenPublication(Project project, boolean hasNative, List<String> projec
...
+                            parent.appendNode("version", MAVEN_VERSION)
+
                             Node dependencies = asNode().appendNode("dependencies")
                             projectDependencies.each { dep ->
                                 Node dependency = dependencies.appendNode("dependency")
                                 dependency.appendNode("groupId", MAVEN_GROUP_ID)
                                 dependency.appendNode("artifactId", "javafx.$dep")
                                 dependency.appendNode("version", MAVEN_VERSION)
-                                dependency.appendNode("classifier", t.name)
+                                dependency.appendNode("classifier", "\${javafx.platform}")
```
saudet: I'm pretty sure we can access the "os.name" and "os.arch" system properties directly from Gradle without Maven...

saudet: Yeah, but Maven doesn't give you any sort of control over those values, so it's pretty limiting. I'll eventually be changing that in ND4J as well. Here's what is look like in JavaCPP:
https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/Loader.java#L83-L112

johanvos: We have the additional constraint that javafx.graphics should be a single module, hence a single jar. That module contains the NativeLibLoader, so that is part of the module (but exists in different flavors, one per module).
I hope that in the not-too-far future maven and gradle have better support for platform specific classifiers *and* modules.
So far, the current solution seems to work (tested on mac/linux/windows). 
A cool feature of nd4j/javacpp which we don't have yet is the possibility of cross-compilation (e.g. you can specify javacpp.platform=ios on a mac). Would be great to have this later as well.

For now, the main goal is to make sure we can provide the module functionality with artifacts uploaded to maven central so that developers can leverage module-info statements in their projects.
That seems to work now.

saudet: I'm pretty sure that the dependency doesn't work from anything else than Maven though, only Maven supports profiles.

johanvos: probably. But I think the build process should make sure the correct code (Java classes and native libs) are selected, instead of deciding this at runtime. With the move towards AOT this sounds plausible. But it's an interesting discussion point.

saudet: It will always need to happen at runtime, we're talking WORA after all, unless we don't and go the JMOD/app way only, in which case it happens at build time for each end-user app.

johanvos: I don't see the link with WORA. WORA is about writing your *code* once. Hence, putting build/dependency logic in build systems (which is more or less their reason of existence) doesn't violate WORA.

saudet: We frequently end up repackaging artifacts in a "uber JAR", and if that JAR file contains only binaries for Mac because the developer is on Mac, it's not going to run on Linux.

Anyway, there's lots of other corner cases as well. You're far from being out of the woods... I do hope people eventually understand the value of standardizing all that in something like JavaCPP.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -25,18 +25,42 @@
...
+import java.security.NoSuchAlgorithmException;
 import java.security.PrivilegedAction;
+import java.util.Arrays;
 import java.util.HashSet;
+import java.util.List;
 
 public class NativeLibLoader {
 
     private static final HashSet<String> loaded = new HashSet<String>();
+    private static final String tmpdir = System.getProperty("java.io.tmpdir");
```
kevinrushforth: `tmpdir` is unused and can be removed

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -130,6 +154,10 @@ private static void loadLibraryInternal(String libraryName) {
                             + libraryName + ") succeeded");
                 }
             } catch (UnsatisfiedLinkError ex2) {
+                // if the library is available in the jar, copy it to /tmp and load it from there
```
kevinrushforth: Maybe comment could say "copy it to cache" rather than `/tmp` ?

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -150,6 +178,113 @@ private static void loadLibraryInternal(String libraryName) {
...
+                        System.err.println("Loaded library " + reallib + " from resource");
+                    }
+                } else if (verbose) {
+                    System.err.println("Unpacked library " + reallib + " from resource");
+                }
+                return true;
+            }
+        } catch (Throwable t) {
+            // we should only be here if the resource exists in the module, but 
+            // for some reasons it can't be loaded.
```
kevinrushforth: Trailing whitespace on the above two lines.

modules/javafx.media/src/main/java/com/sun/media/jfxmediaimpl/NativeMediaManager.java
```diff
@@ -268,7 +289,6 @@ public static MetadataParser getMetadataParser(Locator locator) {
     public MediaPlayer getPlayer(Locator locator) {
         // FIXME: remove this
         initNativeLayer();
-
```
kevinrushforth: Maybe best to revert this change (i.e., restore the blank line) lest someone think that the FIXME comment applies to the creation of the player?

kevinrushforth:
Is there an RFE in JBS for this? If not, can you file one, since I think this is almost ready to to go in.

kevinrushforth:
The two recent changes looks fine to me, and I can run two concurrent apps on Windows now.

Are there other changes you plan to make? I only have one final test I'd like to do before completing my review. If you are ready for final review prior to merging this fix, maybe you can drop the "[WIP]" from the title of the PR and send a review request to openjfx-dev?

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
saudet: No, actually we need to fall back on "java.io.tmpdir" when "user.home" is not writable, for whatever reason, it happens. BTW, it would also be nice to let users set the cache directory.

saudet:
I see at least one more thing missing from this PR. It doesn't provide a locking mechanism to synchronize 2 or more JVM processes trying to modify the cache at the same time...

johanvos:
There are probably many edge cases that need to be fixed. I think we should create a JEP for this and target that to OpenJDK. Having it standardized in the JDK would be nicer than each project having to maintain a solution with all edge-cases.

kevinrushforth:
I did some additional testing and a 0 byte .so file is written in the case where the file exists but the checksum does not match. You might need to reopen or rewind the input stream after computing the checksum.

saudet:
@johanvos Yes, exactly my point! But the guys at Oracle are not interested, so a JEP is pointless for now. Unless you have a better idea, I'll just keep trying to make a point with JavaCPP... @kevinrushforth or anyone else, tips welcome as to how to go about this.

johanvos:
the inputstream obtained from class.getResourceAsStream() doesn't support mark/reset so I have to recreate it, and slightly modified the signature of a method. 
Tested it on linux now, by manually manipulating one library, and that works.
I also improved the read performance for the digestInputStream by using a 4K buffer.

saudet:
In my tests I've found a buffer size of 64 K to be optimal, especially on Windows, which is really slow with small buffer sizes like 1 or 4 K, but a channel as used by JNR is probably best: 
https://github.com/jnr/jffi/blob/master/src/main/java/com/kenai/jffi/internal/StubLoader.java#L367

build.gradle
```diff
@@ -552,6 +551,7 @@ if (HUDSON_JOB_NAME == "not_hudson") {
 defineProperty("RELEASE_SUFFIX", relSuffix)
 defineProperty("RELEASE_VERSION_SHORT", "${RELEASE_VERSION}${RELEASE_SUFFIX}")
 defineProperty("RELEASE_VERSION_LONG", "${RELEASE_VERSION_SHORT}+${PROMOTED_BUILD_NUMBER}${relOpt}")
+defineProperty("MAVEN_VERSION", "$RELEASE_VERSION_LONG")
```
kevinrushforth: This will be "11-ea+19" for the next build. Will this work? Or does it need to be something more like "11.0.0-ea-19"?

Here are the settings that were used by our Hudson job for build 18:

```
PROMOTED_BUILD_NUMBER: 18
RELEASE_VERSION: 11
RELEASE_SUFFIX: -ea
RELEASE_VERSION_SHORT: 11-ea
RELEASE_VERSION_LONG: 11-ea+18
RELEASE_VERSION_PADDED: 11.0.0.0
```

javafx.pom
```diff
@@ -11,10 +11,6 @@
     <description>OpenJFX JavaFX</description>
     <properties>
         <javafx.version>11.0.0-SNAPSHOT</javafx.version>
```
kevinrushforth: This this OK to hard-code to the SNAPSHOT version?

shruda:
@johanvos,
@kevinrushforth asked you some days ago about the empty jars but I found now answer at the comments.

We did some tests with your published SNAPSHOTS's and we had to force an exclude of the empty jars at the dependencies. 
Otherwise e.g. Eclipse shows a warning that the module name is instable because of the "auto-generated" module name in case of the empty jars.

Best Regards,
Steve

tiainen:
The empty jars are required because we'd like to support both maven and gradle. The base requirement is that we can declare dependencies between the javafx artifacts (i.e. javafx-controls depends on javafx-graphics, which in turn depends on javafx-base). This will allow maven and gradle to automatically download the correct artifacts.

_Case 1_

We remove the empty jars. We define the dependencies without specifying the platform classifier. This will cause maven to try and find a javafx-graphics jar without classifier which doesn't exist. We only have artifacts with a classifier. During dependency resolution, maven will fail with the following message.

    [ERROR] Failed to execute goal on project hellosphere: Could not resolve dependencies for project org.projavafx.jfx3d:hellosphere:jar:1.0-SNAPSHOT: The following artifacts could not be resolved: org.openjfx:javafx-graphics:jar:11-internal+0-2018-07-13-144331, org.openjfx:javafx-base:jar:11-internal+0-2018-07-13-144331: Failure to find org.openjfx:javafx-graphics:jar:11-internal+0-2018-07-13-144331 in https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced -> [Help 1]

_Case 2_

We remove the empty jars. To make it work in maven, we define the dependencies by adding the classifier element to each dependency. For maven, we define a property named `${javafx.platform}` that gets filled in automatically by the parent javafx pom definition (see `javafx.pom`). This works fine in maven, but gradle doesn't process the pom files in the same way. It doesn't resolve the maven property and will therefor fail when resolving the dependencies with the following message:

    * What went wrong:
    Could not resolve all files for configuration ':compileClasspath'.
    > Could not find javafx-graphics-${javafx.platform}.jar (org.openjfx:javafx-graphics:11-internal+0-2018-07-13-140316).
    > Could not find javafx-base-${javafx.platform}.jar (org.openjfx:javafx-base:11-internal+0-2018-07-13-140316).

In the end, the solution was to introduce an empty jar file.

johanvos:
support for modules is still in flux with maven/gradle, so this will be a moving target, especially because there are so many combinations of {repositories, build systems, IDE's}.

I suggest we continue this discussion on the mailinglist instead of on this PR?

shruda:
Thanks @tiainen  for the explanation.
@johanvos , it's okay for me.

PR #87, opened 2018-05-24
Fix JDK-8202743: Dashed Stroke randomly painted incorrectly, may freeze application

bourgesl:
This PR corresponds to the jfx backport of the JDK-8202580 bug fix:
https://bugs.openjdk.java.net/browse/JDK-8202580

Notes: 
- (D)Dasher changes corresponds to MarlinFX 0.8.2 (upgrade to 0.9.2 later) 
- test adapted to javafx API using scene.snapshot()

### DIFF
--- a/tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
+++ b/tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
@@ -95,6 +95,7 @@ public class DashedRectTest {
         public void start(Stage primaryStage) throws Exception {
             this.stage = primaryStage;
 
+            stage.setScene(new Scene(new Group()));
             stage.setTitle("DashedRectTest");
             stage.show();
 
@@ -130,8 +131,6 @@ public class DashedRectTest {
     public void TestDashedPath() throws InterruptedException {
 
         final int size = MAX * 2;
-
-        final WritableImage img = new WritableImage(size, size);
 
         Util.runAndWait(() -> {
 
@@ -177,14 +176,14 @@ public class DashedRectTest {
             final SnapshotParameters sp = new SnapshotParameters();
             sp.setViewport(new Rectangle2D(0, 0, size, size));
 
-            scene.getRoot().snapshot(sp, img);
+            final WritableImage img = scene.getRoot().snapshot(sp, new WritableImage(size, size));
+
+            // Check image on few pixels:
+            final PixelReader pr = img.getPixelReader();
+
+            // 10, 5 = blue
+            checkPixel(pr, 10, 5, BLUE_PIXEL);
         });
-
-        // Check image on few pixels:
-        final PixelReader pr = img.getPixelReader();
-
-        // 10, 5 = blue
-        checkPixel(pr, 10, 5, BLUE_PIXEL);
     }
 
     private static void checkPixel(final PixelReader pr,
--- a/tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
+++ b/tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
@@ -1,0 +1,217 @@
+/*
+ * Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package test.com.sun.marlin;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import javafx.application.Application;
+import javafx.application.Platform;
+import javafx.collections.ObservableList;
+import javafx.geometry.Rectangle2D;
+import javafx.scene.Group;
+import javafx.scene.Scene;
+import javafx.scene.SnapshotParameters;
+import javafx.scene.image.PixelReader;
+import javafx.scene.image.WritableImage;
+import javafx.scene.paint.Color;
+import javafx.scene.shape.ClosePath;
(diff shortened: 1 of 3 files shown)


johanvos:
Discussed and approved for 2d-dev in http://mail.openjdk.java.net/pipermail/2d-dev/2018-May/009183.html.

johanvos:
@bourgesl About your note: Do you recommend applying this PR, or would it be better to upgrade to MarlinFX 0.9.2 instead (which has more changes)?

kevinrushforth:
I think it would be better to apply this PR and do the upgrade to 0.9.2 separately (easier to review).

I'll run the unit test today (which should be the only difference between this and the Java2D fix), and finish my review.

bourgesl:
@johanvos I deliberately made this PR on current jdk11 (Marlin 0.8.2) before preparing the Marlin 0.9.2 patch  (4K LOC) to be integrated asap (regression)

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
```diff
@@ -0,0 +1,217 @@
...
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package test.com.sun.marlin;
+
+import java.io.File;
+import java.io.IOException;
```
kevinrushforth: No need to use File / IOException

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * @test
+ * @summary verify that dashed rectangle is properly rasterized
+ * @bug 8202743
+ */
+public class DashedRectTest {
+
+    static final int BLUE_PIXEL = 0xff0000ff;
+
+    static final File OUTPUT_DIR = new File(".");
```
kevinrushforth: This is unused and should be removed.

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
```diff
@@ -0,0 +1,217 @@
...
+
+            // Corrupt Marlin Dasher.dash cached array:
+            final Path path1 = new Path();
+            path1.getElements().addAll(
+                    new MoveTo(20, 20),
+                    new HLineTo(70),
+                    new VLineTo(70),
+                    new HLineTo(20),
+                    new ClosePath()
+            );
```
kevinrushforth: Minor suggestion: Normally we would populate the content of the scene graph in the test method and not in the Application start method, which is called from the static `@BeforeClass` method.

Since you only have a single test method, it isn't crucial to fix this...think of it more like a best practice. It's up to you whether to make this change.

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
```diff
@@ -0,0 +1,217 @@
...
+
+    @Test(timeout = 10000)
+    public void TestDashedPath() throws InterruptedException, IOException {
+
+        final int size = MAX * 2;
+
+        final WritableImage img = new WritableImage(size, size);
+
+        done = false;
+        Platform.runLater(() -> {
```
kevinrushforth: Util.runAndWait is preferred, since it takes care of both the waiting and the propagation of exceptions.

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
```diff
@@ -0,0 +1,217 @@
...
+            wait();
+        }
+    }
+
+    private static void checkPixel(final PixelReader pr,
+                                   final int x, final int y,
+                                   final int expected) {
+
+        final int rgb = pr.getArgb(x, y);
+        if (rgb != expected) {
```
kevinrushforth: Have you run this on Mac? I will try it, but I note that we've had problems doing exact color comparisons on Mac with the color space conversion (although snapshot might be unaffected...I'll check).

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
```diff
@@ -0,0 +1,217 @@
...
+            throw new IllegalStateException("bad pixel at (" + x + ", " + y
+                    + ") = " + rgb + " expected: " + expected);
```
kevinrushforth: Minor suggestion: use assertEquals or else call "fail" rather than throwing ISE.

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
kevinrushforth: Answering my own question, this runs fine on Mac, so no need to worry.

bourgesl:
I fixed the test in the latest commit, hope it is OK now.

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
bourgesl: fixed

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
bourgesl: fixed

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
```diff
@@ -97,6 +95,46 @@ public void init() {
         public void start(Stage primaryStage) throws Exception {
             this.stage = primaryStage;
 
+            stage.setTitle("DashedRectTest");
```
kevinrushforth: I recommend setting a dummy Scene before showing the Stage, otherwise it can render garbage (at least on Linux it does), which could lead to someone thinking there is something wrong:

```
        stage.setScene(new Scene(new Group()));
```

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
```diff
@@ -133,84 +171,29 @@ public void start(Stage primaryStage) throws Exception {
...
-
-        done = false;
-        Platform.runLater(() -> {
             final SnapshotParameters sp = new SnapshotParameters();
             sp.setViewport(new Rectangle2D(0, 0, size, size));
 
-            myApp.stage.getScene().getRoot().snapshot(sp, img);
-
-            signalDone();
+            scene.getRoot().snapshot(sp, img);
```
kevinrushforth: I just noticed that you aren't using the return value from snapshot. According to spec, you should not assume that it will always use the image that is passed in (it will use it if you can). This should be:

```
    img = scene.getRoot().snapshot(sp, img);
```

If you put the check into the runAndWait as well, then you don't need to make img final.

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
bourgesl: fixed

tests/system/src/test/java/test/com/sun/marlin/DashedRectTest.java
bourgesl: fixed

bourgesl:
I will update the proposed webrev...

bourgesl:
I added the proposed patch (webrev) to https://bugs.openjdk.java.net/browse/JDK-8202743

bourgesl:
Kevin, could you push in jfx-dev (hg) for me as I am leaving for the week-end ?

kevinrushforth:
I, too, am out today. I can merge this into the develop branch, but pushing to openjfx repo on hg.openjdk will need to be Monday.

bourgesl:
Thanks, kevin.
I will push myself into hg.openjdk later

Issue #90, opened 2018-06-06
Automatic merge from master into develop branch not run


kevinrushforth:
I merged pull request #88 into the master branch yesterday. I expected that this would get merged into develop soon after that, but it didn't (and still hasn't about 15 hours later).

@tiainen Is the sync job still running regularly? What triggers it?

kevinrushforth:
I see that it was run just now. Was it kicked off manually, or on a timer?

tiainen:
The job is run every hour around the 50 minute mark. It currently fails whenever changes are pushed to the master branch on github. The job executes the following script: https://github.com/javafxports/openjdk-jfx/blob/develop/.github/sync_upstream.sh

It fails at line 27 when trying to push to github, with the following message:

    Push master (github)
    To git@github.com:javafxports/openjdk-jfx.git
     ! [rejected]        master -> master (fetch first)
    error: failed to push some refs to 'git@github.com:javafxports/openjdk-jfx.git'
    hint: Updates were rejected because the remote contains work that you do
    hint: not have locally. This is usually caused by another repository pushing
    hint: to the same ref. You may want to first integrate the remote changes
    hint: (e.g., 'git pull ...') before pushing again.
    hint: See the 'Note about fast-forwards' in 'git push --help' for details.

Currently, I fix this issue manually but as a permanent solution, I think we just need to pull in the changes from github/master before we do the push? This normally only fails when there are merge conflicts between the mercurial repository and the master branch on the github repository.

kevinrushforth:
OK, thanks for the explanation. I think your proposed solution sounds fine. There shouldn't ever be merge conflicts as long as we follow the rule that commits pushed directly to master must only touch files that are not tracked in the upstream hg repo (e.g., .github/** or the Travis / appveyor CI files).

Until then, I'm happy to fix this manually on the rare occasions that I merge a PR into master.

tiainen:
About fixing it manually, that's another issue. The job is currently run on a jenkins environment that is not publicly accessible. To fix it manually requires logging in to that machine and manually bring the local master branch up-to-date with the one on github.

For now I'll add a line in the script to perform a pull (fetch + merge) before pushing to github. That should get us a lot further already.

PR #91, opened 2018-06-06
Merge changes from github/master before pushing

tiainen:
As discussed in #90. Pull any changes that might have been pushed directly to the master branch on github before pushing to github. This minimizes manual interaction during the automatic synchronization process.


.github/sync_upstream.sh
```diff
@@ -23,6 +23,9 @@ git fetch "imports/openjfx/jfx-dev/rt/root"
 echo "Merge (root)"
 git merge "imports/openjfx/jfx-dev/rt/root/master" -m "Merge from (root)" --no-ff
 
+echo "Pull master (github)"
+git pull github master
```
kevinrushforth: Should this be done before the previous git merge? That matches what is done for develop (it probably doesn't matter, but might be cleaner).

tiainen: The order does indeed not matter, but I've moved it after the checkout making the steps analogous to how develop branch is handled.

kevinrushforth:
@tiainen I think you will need to manually fix this up one last time, since the existing script won't pull the new changeset from master (because the fix isn't there yet).

tiainen:
I thought so as well, but luckily the jenkins job is configured to download the script from master first, and then execute it:

    wget https://raw.githubusercontent.com/javafxports/openjdk-jfx/master/.github/sync_upstream.sh -O sync_upstream.sh
    chmod +x sync_upstream.sh
    ./sync_upstream.sh

So the new flow is already active.

kevinrushforth:
And I see it worked. Nice.

PR #92, opened 2018-06-06
Fix JDK-8201553: Update FX build to use gradle 4.8

kevinrushforth:
This is a proposed fix for [JDK-8201553](https://bugs.openjdk.java.net/browse/JDK-8201553) to update our build to use gradle 4.8. Note that this does not change the minimum required version of gradle for developers. That can be done separately.


kevinrushforth:
I have done local testing on all three platforms as well as the CI testing that was triggered by this PR. This is now ready to be reviewed and considered for merging.

johanvos:
Looks ok to me, and it works on Linux. +1 for merging this.

PR #96, opened 2018-06-08
upgrade MarlinFX 0.9.2

bourgesl:
Here is the PR providing MarlinFX 0.9.2 to run it against CI (testing).

Formal review will happen on webrev (openjfx RFR) as this patch is large (5.8K LOC) but 95% identical to Marlin2D (integrated in OpenJDK 11).

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/marlin/Curve.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/marlin/Curve.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2007, 2017, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2007, 2018, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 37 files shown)


modules/javafx.graphics/src/main/java/com/sun/prism/impl/shape/DMarlinPrismUtils.java
```diff
@@ -44,31 +44,29 @@
...
-
-    static final boolean DO_CLIP = MarlinProperties.isDoClip();
-    static final boolean DO_CLIP_FILL = true;
+    // slightly slower ~2% if enabled stroker clipping (lines) but skipping cap / join handling is few percents faster in specific cases
+    static final boolean DISABLE_2ND_STROKER_CLIPPING = true;
 
     static final boolean DO_TRACE_PATH = false;
 
+    static final boolean DO_CLIP = MarlinProperties.isDoClip();
+    static final boolean DO_CLIP_FILL = true;
```
johanvos: This was there before as well, but what is the purpose? It seems to be only used on a single place. Should this be a boolean that can be set via a property?

modules/javafx.graphics/src/main/java/com/sun/prism/impl/shape/MarlinPrismUtils.java
```diff
@@ -44,31 +44,29 @@
...
-
-    static final boolean DO_CLIP = MarlinProperties.isDoClip();
-    static final boolean DO_CLIP_FILL = true;
+    // slightly slower ~2% if enabled stroker clipping (lines) but skipping cap / join handling is few percents faster in specific cases
+    static final boolean DISABLE_2ND_STROKER_CLIPPING = true;
 
     static final boolean DO_TRACE_PATH = false;
 
+    static final boolean DO_CLIP = MarlinProperties.isDoClip();
+    static final boolean DO_CLIP_FILL = true;
```
johanvos: see above -- there are a number of those.

modules/javafx.graphics/src/main/java/com/sun/marlin/Curve.java
```diff
@@ -33,86 +33,94 @@
...
+        } else if (type == 4) {
+            set(points[0], points[1],
+                points[2], points[3]);
+        } else {
             set(points[0], points[1],
                 points[2], points[3],
                 points[4], points[5]);
-            return;
-        default:
-            throw new InternalError("Curves can only be cubic or quadratic");
```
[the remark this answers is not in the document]
bourgesl: Good catch but this code is always called with type = 4/6... so I simplified.

modules/javafx.graphics/src/main/java/com/sun/prism/impl/shape/DMarlinPrismUtils.java
bourgesl: I like constants that are inlined well by hotspot. Moreover the clip fill flag helped me in Marlin 0.8 to handle properly the even/odd rule.
It could be removed but I think it is more obvious like this.

bourgesl:
Will push directly in jfx-dev asap

PR #97, opened 2018-06-09
Travis-CI: use integrated jdk constants

mwkroening:
Travis-CI integrated install-jdk.sh into their platform: 
https://sormuras.github.io/blog/2018-05-31-jdk-matrix.html

Utilizing this we can reduce the complexity of our Travis-CI procedure.
It also allows for adding additional jdk versions like openjdk-ea very easily.

It does not support OracleJDK on OSX jet, but I don't think this is a problem, is it?


kevinrushforth:
@brcolow should comment on this.

Also, since this PR touches only CI files and no files that exist in HG on the upsgtream openjfx/jfx-dev/rt repo, then it should be merged into master, if approved. As such, it should be rebased against master (it is the one exception to the rule that PRs are done against develop).

brcolow:
We want to specifically use OpenJDK for all platforms, not Oracle JDK. If `install-jdk.sh` was added, we have to make sure that we are still using OpenJDK.

mwkroening:
Yes, this is still using OpenJDK.
I only pointed that OSX/OracleJDK thing out because I was not sure, if that was important to this project.

brcolow:
LGTM.

johanvos:
Looks good. Since the checks pass, I assume that by definition it works. 
I wonder where travis gets the openjdk distribution from, and what happens when there is an update?

mwkroening:
Before 2018-05-31 there was no official way to use any Java version above 8 in Travis.
A solution avoiding any manual installation fiddling has been [install-jdk.sh](https://github.com/sormuras/bach#install-jdksh) ([Source](https://github.com/sormuras/bach/blob/master/install-jdk.sh)) since 2017-12-08. We are [using this method for Linux](https://github.com/javafxports/openjdk-jfx/pull/97/files#diff-354f30a63fb0907d4ad57269548329e3L50) only ([after downloading it](https://github.com/javafxports/openjdk-jfx/pull/97/files#diff-36c2be5516485318c75883ee12afcff3L17)), as [OSX support](https://sormuras.github.io/blog/2018-05-29-jdk-matrix.html) just came on 2018-05-29. Therefore our CI on OSX still involves manual installation fiddling ([here](https://github.com/javafxports/openjdk-jfx/pull/97/files#diff-36c2be5516485318c75883ee12afcff3L5) and [here](https://github.com/javafxports/openjdk-jfx/pull/97/files#diff-945aa645b7b1356e9a1037a7a7033e54L4)).
Now Travis integrated the use of install-jdk.sh. That means supplying the build constant `jdk: openjdk10` install-jdk.sh is automatically triggered on [Linux](https://travis-ci.org/javafxports/openjdk-jfx/jobs/390184797#L407) and [OSX](https://travis-ci.org/javafxports/openjdk-jfx/jobs/390184798#L21) and will then download the OpenJDK build from [jdk.java.net](http://jdk.java.net/).
It is also possible to provide `openjdk-ea` as an additional build constant to always test with the latest and greatest early-access build as well, if you are interested.

kevinrushforth:
Since @johanvos and @brcolow are happy with this, I can squash / merge it later today (I'll wait in case anyone has additional thoughts on this).

Issue #98, opened 2018-06-11
[swt] Dragging a Window with FXCanvas from HiDPI to normal DPI display results into empty screen


tomsontom:
This is a clone of https://bugs.openjdk.java.net/browse/JDK-8204676. The problem is that if the screens you drag them SWT-Shell have a different renderScale getPixel() returns null until the next time the PresentableState.update(float,float,float) is called

PR #99, opened 2018-06-11
WIP: synchronize sceneState on scale-factor change

tomsontom:
refs #98

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/EmbeddedScene.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/EmbeddedScene.java
@@ -33,11 +33,12 @@ import javafx.scene.input.InputMethodEvent;
 import javafx.scene.input.InputMethodRequests;
 import javafx.scene.input.InputMethodTextRun;
 import javafx.scene.input.MouseEvent;
-import javafx.scene.input.ScrollEvent;
 import javafx.scene.image.PixelFormat;
 import java.nio.IntBuffer;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import com.sun.javafx.cursor.CursorFrame;
 import com.sun.javafx.embed.AbstractEvents;
 import com.sun.javafx.embed.EmbeddedSceneDTInterface;
@@ -47,6 +48,7 @@ import com.sun.javafx.scene.input.KeyCodeMap;
 import com.sun.javafx.scene.traversal.Direction;
 import com.sun.javafx.sg.prism.NGNode;
 import com.sun.javafx.tk.TKClipboard;
+import com.sun.javafx.tk.TKPulseListener;
 import com.sun.javafx.tk.Toolkit;
 import com.sun.prism.paint.Color;
 import com.sun.prism.paint.Paint;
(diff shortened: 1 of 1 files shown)


modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/EmbeddedScene.java
```diff
@@ -155,6 +155,7 @@ public void setPixelScaleFactors(float scalex, float scaley) {
         renderScaleX = scalex;
         renderScaleY = scaley;
         entireSceneNeedsRepaint();
+        sceneState.update();
```
kevinrushforth: This is not thread-safe. This method must only be called on the JavaFX event thread, and with the render lock held. You will need a different solution to the problem you are trying to solve.

kevinrushforth (changes requested):

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/EmbeddedScene.java
tomsontom: ok I see in SWT this is called from the correct Thread but in Swing is is called from EDT. The problem is that sceneState update is ONLY called when the scene is dirty and there's not even internal API to call.

kevinrushforth: even if called on the correct thread it will need the renderlock, and I don't know if this is an OK place to acquire that lock (it might be, but that would need to be explored).

tomsontom:
I looked a bit in the codebase and WindowStage calls GlassScene.updateSceneState() (eg in setBounds) in a similar context and I don't see them getting a lock. If not mistaken this one gets called by a TkPulseListener registered in Window at line 1081.

In contrast to that the dispose-method in EmbeddedScene calls it with a lock.

kevinrushforth:
I'll take a closer look when I get time.

tomsontom:
I've pushed a different solution using a TkPulseListener (borrowing the scheme from WindowStage/Window)

kevinrushforth:
Conceptually that seems like the right idea (and one I was thinking to suggest), so I'll review it.

kevinrushforth:
This will need to be tested with both Swing interop (JFXPanel) and SWT interop (FXCanvas). Two questions:

1. Can you say what testing you have done?
2. Have you confirmed that the listener doesn't cause a memory leak?

I'll do some testing (might not be this week, since I am reviewing more urgent fixes with a tighter deadline). I'll also review the changes in more detail.

kevinrushforth:
Also, can you file a JBS bug for this?

tomsontom:
On 1. I‘ve tested FXCanvas moving it between HiDPI and normal DPI screens on OS X and Windows - I‘ll try JFXPanel

On 2. I remove the listener upon dispose so I can‘t see how I could introduce a memory leak with this change

tomsontom:
Need to correct - I only tested on OS X - as the hidpi fix on windows was not yet available

kevinrushforth:
JBS bug: [JDK-8204676](https://bugs.openjdk.java.net/browse/JDK-8204676)

kevinrushforth:
Have you had a chance to test on Windows yet? I can review it later this week if you are ready. Please send a review request to openjfx-dev so others can see it, too.

tomsontom:
Unfortunately no I need to bring back up my windows toolchain after changing laptops

kevinrushforth:

Issue #71, opened 2018-04-22
JavaFX and MathML.


scientificware:
When I discovered that JavaFX could manage MathML, I also discovered that it was buggy. Becauseof MathML is rendered with WebKit, I compared with other WebKit based web browsers. I also noticed that since 2016 that those web browser make good progress with MaltML display. First I thought that the differences were due to outdated WebKit version used by JavaFX but last releases, with an uptodate WebKit version, show that the problem is the same.

I contacted F. Wang from Igalia who worked to port or improve MathML support in FireFox and later in WebKit. He confirms that it was a bug in JavaFX not in WebKit witch displays MathML correctly.

As he suggested me, I wrote to the JavaFX mailing list and had a short discussion about this issue on OpenJDK.Java.net :

http://mail.openjdk.java.net/pipermail/openjfx-dev/2017-December/021112.html
https://bugs.openjdk.java.net/browse/JDK-8147476

So, as suggested by mainteners, I'm trying to find by myself what's wrong with MathML in javaFX HTMLEditor.

- As far as I could read in javaFX java code,
  - I understand that HTMLEditor and WebView only manage and display a cached page.
  - I know that JavaFX use WebKit to display a webpage.
- So I don't understand why, 
  - when I tested webkit with Safari on OSX and IOS last versions, in an Apple Store and with my Ipod (with an oudated IOS 9), the display is pretty good :
![ipod_ios9_screencapture](https://user-images.githubusercontent.com/19194678/39093477-8d57983c-4620-11e8-9878-f98f945149c4.PNG)
  -  And with JavaFX HTMLEditor the display is pretty bad (https://github.com/scientificware/javafxmathml) : 
![screen_javafx_mathml](https://user-images.githubusercontent.com/19194678/39093773-2aea4de8-4625-11e8-8cce-407416796205.png)

I tested all applications with Mozilla MathML Torture Test 
(https://developer.mozilla.org/fr/docs/Mozilla/MathML_Project/MathML_Torture_Test) 

Could someone help me ?
- To find in which code level is the problem.
- I don't think the problem is in the JavaFX java code, I read it several times.
- I tried to find what wrong in the native code but, I don't know where to start !
- So where I should start to work ?

scientificware:
I think that the problem is related only with the vertical size of each element :
I notice that the size of the cursor is right outside the mathematics text but when we move it through the mathematics text, the cursor size is far much lower than text size.
[...]

PR #102, opened 2018-06-13
JavaFX and MathML. #71

scientificware:
HI !
Please could you review this patch.
The remaining problems are due to the missing OpenMathData table support .
A quick solution would be :
- to include a font file containing the right glyph alignment.
- and modify the mathml.css file to force MathML renderer to use them first.
Best regards.


kevinrushforth:
Since this is the initial review request, please squash your commits into a single commit before we start to review it.

kevinrushforth:
Here is some quick feedback that you can address at the same time you squash your commits:

1. You need to remove the gradle.properties file. It should not be checked in.

2. Your test will need to be reworked to be a proper unit test. Also, if you need to display a window, it will need to go into tests/system instead of modules/javafx.web.

kevinrushforth:
For future reference, you can push a squashed commit to an in-progress PR, although if you think it's cleaner, then opening up a new PR is always an option.

scientificware:
- [X] Since this is the initial review request, please squash your commits into a single commit before we start to review it.
- [X] Your test will need to be reworked to be a proper unit test.
  - **Could you point me a reference unit test ?**
- [X] You need to remove the gradle.properties file. It should not be checked in.
- [X] For future reference, you can push a squashed commit to an in-progress PR, although if you think it's cleaner, then opening up a new PR is always an option.
  - **I have no preference, I going to reopen it and push a squashed commit.**

PR #103, opened 2018-06-13
JavaFX and MathML. #71

scientificware:
HI !
Please could you review this patch.
The remaining problems are due to the missing OpenMathData table support .
A quick solution would be :
- to include a font file containing the right glyph alignment.
- and modify the mathml.css file to force MathML renderer to use them first.
Best regards.


kevinrushforth:
Please squash these commits into a single commit to make it easier for us to review.

scientificware:
Sorry, I thought having choose the right options for my repository "Allow squash".
In fact, I have to commit from the master branch too.

scientificware:
I'm not succeed. With the online tools, I can squash in my fork but not outside.
Really sorry, I will see that in the weekend.

kevinrushforth:
You will likely need to do this using local git tools.

PR #120, opened 2018-07-03
init bcm host to allow dispman calls

imario42:
I had to add this init method to make the dispman call working.

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/monocle/dispman/DispmanCursor.c
+++ b/modules/javafx.graphics/src/main/native-glass/monocle/dispman/DispmanCursor.c
@@ -78,6 +78,10 @@ JNIEXPORT void JNICALL Java_com_sun_glass_ui_monocle_DispmanCursor__1initDispman
     cursor.element = 0;
     cursor.currentCursor = 0;
     cursor.isVisible = 0;
+
+    // init bcm host to allow dispman calls
+    (*wr_bcm_host_init)();
+
     addDispmanxElement();
     updateCursor();
 }


kevinrushforth:
@imario42 Your OCA has been recorded. This PR can be evaluated now.

johanvos:
@imario42 What toolchain do you use, and where did you encounter the error if the bcm_host_init() call is not called? Would be nice to have an issue with a log that fails before and succeeds after.
Reason for asking is that I assume this fix is needed starting from a specific lib version, and we should somehow be able to document the requirements, or build different versions for different rendering systems.

johanvos:
I looked into this more detailed. Running older versions of JavaFX (8ux) on newer raspian distributions will fail now, because bcm_host_init() is not called. Hence, it should at least somehow be called. 
I'm not sure the cursorInit is the best place to do it though. 
However, the whole monocle part needs more attention, so I think we accept this PR and come back to the monocle structure in 12+

johanvos:
Some more info: this is required on the RPI:
https://elinux.org/Raspberry_Pi_VideoCore_APIs has 

> "The Raspberry Pi requires that the bcm_host_init() function is called first before any GPU calls can be made."

This does not mean however that this call is required when building for a different target.

johanvos:
Adding more thoughts: if bcm_host_init is not called yet, it most likely means DispamnAcceleratedScreen.c is not called. This happens when for some reason the prism-es2-monocle pipeline is not initialized, and there can be many reasons for this, including not supplying the required properties or missing/wrong EGL and GLES libraries on the device.
In those cases, I think it should be made very clear that there is a fallback to the sw rendering, and in that case the call to bcm_host_init should happen somewhere else (probably triggered via something like prism-sw-monocle).

johanvos:
@imario42 can you see if the embedded SDK build with explanation at http://docs.gluonhq.com/embedded/ works for you?
That build is created *without* the additional bcm host init.

imario42:
@johanvos : Yes, that way it worked as described on your homepage. Thanks!

PR #121, opened 2018-07-03
[WIP] Try building javafx.web module

arajkumar:
Newly upgraded WebKit 606.1 supports unified source build which significantly reduces the build time.

### DIFF
--- a/.ci/before_install.sh
+++ b/.ci/before_install.sh
@@ -2,9 +2,9 @@
 set -euo pipefail
 
 if [[ "${TRAVIS_OS_NAME}" == osx ]]; then
-  brew install ccache
   brew update
   brew install ant
+  brew install ccache
   brew install findutils
   brew unlink python # fixes 'run_one_line' is not defined error in backtrace
 fi
--- a/.ci/before_install.sh
+++ b/.ci/before_install.sh
@@ -2,6 +2,7 @@
 set -euo pipefail
 
 if [[ "${TRAVIS_OS_NAME}" == osx ]]; then
+  brew install ccache
   brew update
   brew install ant
   brew install findutils
--- a/.ci/script.sh
+++ b/.ci/script.sh
@@ -4,7 +4,15 @@ set -uo pipefail
 echo "which java: $(which java)"
 ulimit -c unlimited -S
 
-sh ./gradlew all test -PCONF=Release -PCOMPILE_WEBKIT=true --no-daemon --stacktrace --info
+if [[ $1 == webkit ]]; then
+  BUILD_WEBKIT=true
+else
+  BUILD_WEBKIT=false
+fi
+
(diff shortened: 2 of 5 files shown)


ararunprasad:
Note: This is for experimental purpose only, just to evaluate whether we could build javafx.web with new unified source changes.

kevinrushforth:
Note that the Travis and Appveyor CI builds don't build webkit.

kevinrushforth:
Please ignore my last (now deleted) comment. :)

scientificware:
I confirm, it's very impressive. Thank you.

scientificware:
I had also get a problem when trying to build WebCore.
My previous config didn't work no more because of ant or a missing of java version.
After having install ant 1.9.12 and Java 8 the build works fine.
Previously, I used the ant version provide by gradle.
One of these solved the problem.

brcolow:
Just an FYI, without `DebugNative` the PDB debugging symbols won't be generated and uploaded on Windows. Currently we are not uploaded the debugging symbols generated on the Travis builds, so that should be okay but it will reduce test logging verbosity which could make working on PRs more difficult (for those that touch native code). What is the time is the `CONF` is kept at DebugNative?

ararunprasad:
@brcolow As of now I'm just experimenting to see whether it is possible to enable WebKit compilation with Travis time constrains. I'm not sure about the time taken to build with CONF=DebugNative, locally it is around 1.5 times of Release build.

ararunprasad:
Linux build proceeds upto 89% and macOS builds upto 95% before getting timed out. Looks like Windows is failing because of missing cygwin. I'm looking for some build optimisation techniques to reduce the time.

brcolow:
cygwin is installed on Appveyor, it may be a PATH issue.

ararunprasad:
With more files(32) on unified bundle, on macOS jfxwebkit library was build and terminates while building DRT.

brcolow:
FWIW I messed around with trying to get ccache to work for the JavaFX build - never could because ccache was never called by the native compilation step in gradle.

johanvos:
Can we separate the job that builds libjfxwebkit so that it gets the full amount of available time?

ararunprasad:
@johanvos. Probably if we cache the generated JNI headers, native module has dependency only on the JNI headers.

johanvos:
Would it make sense (in the longer term) to separate the building of the native library into a different project, where the generated JNI headers are imported as a dependency (generated by the main project)?

ararunprasad:
@johanvos, it would. But still 50minutes time limit imposed by Travis will not be sufficient to build WebKit.

kevinrushforth:
Closing this old experimental PR given that this sandbox repo will be retired soon. See [this message](https://mail.openjdk.java.net/pipermail/openjfx-dev/2019-September/023551.html).

PR #125, opened 2018-07-06
JDK-8207839: [win] Negative glyph_id causes ArrayIndexOutOfBoundsException

nakajima-akira:
[Root cause]
This bug is only on Windows.
This bug is caused by int -> short conversion.
 e.g. 65535(int) -> -1(short)

Correct conversion is -1(short) -> 65535(int).
But now converting as -1(short) -> -1(int)  because of without MASK.

[Solution]
> -            iglyphs[i] = glyphs[j] | slotMask;
> +            iglyphs[i] = ((int)glyphs[j] & SHORTMASK) | slotMask;

[Testing]
Please run TestHandlingBigGlyphID_Win10
 or TestHandlingBigGlyphID_Win7_and_Win10.java.
Before applying this patch, this bug causes ArrayIndexOutOfBoundsException.


kevinrushforth:
@nakajima-akira Welcome to OpenJFX. Please read [CONTRIBUTING.md](https://github.com/javafxports/openjdk-jfx/blob/develop/.github/CONTRIBUTING.md), specifically the parts about needing to sign an Oracle Contributor Agreement and to file a bug in JBS.

Also, you will need to provide a test that can be used to verify your fix -- this should be an automated unit test unless there is a compelling reason why such isn't possible.

Thank you.

nakajima-akira:
Thank you for your advice.

I have signed OCA as named NTT Comware Corporation.
http://www.oracle.com/technetwork/community/oca-486395.html#n
Is this OCA not enough?

nakajima-akira:
Hi Kevin.

Thank you for your advice.

I have a signed OCA as named NTT Comware Corporation.
http://www.oracle.com/technetwork/community/oca-486395.html#n
Is this OCA not enough?


Regards.
--------------------------------------
Company: NTT Comware Corporation
Name: Akira Nakajima
E-Mail: nakajima.akira@nttcom.co.jp
OCA : http://www.oracle.com/technetwork/community/oca-486395.html#n
--------------------------------------

kevinrushforth:
Yes, that is sufficient. I just sent you an email to confirm.

modules/javafx.graphics/src/main/java/com/sun/javafx/font/directwrite/DWGlyphLayout.java
```diff
@@ -138,14 +138,15 @@ public void layout(TextRun run, PGFont font, FontStrike strike, char[] text) {
...
+        final int SHORTMASK = 0x0000ffff;
         boolean missingGlyph = false;
         i = 0; j = rtl ? glyphCount - 1 : 0;
         while (i < glyphCount) {
             if (glyphs[i] == 0) {
                 missingGlyph = true;
                 if (composite) break;
             }
-            iglyphs[i] = glyphs[j] | slotMask;
+            iglyphs[i] = ((int)glyphs[j] & SHORTMASK) | slotMask;
```
kevinrushforth: The cast to `(int)` is redundant: `glyphs[j]` is an int.

tests/system/src/test/java/test/javafx/scene/text/TestHandlingBigGlyphID_Win10.java
```diff
@@ -0,0 +1,38 @@
...
+import static javafx.application.Application.launch;
+import javafx.scene.Group;
+import javafx.scene.Scene;
+import javafx.scene.paint.Color;
+import javafx.scene.text.Font;
+import javafx.scene.text.Text;
+import javafx.scene.text.TextFlow;
+import javafx.stage.Stage;
+
+public class TestHandlingBigGlyphID_Win10 extends Application {
```
kevinrushforth: In order for this test to be run, you need to do the following:

1. The name of the test class must end with exactly `Test` -- so I suggest something like `BigGlyphIDTest` (no need for "Win10" to be in the name).

2. The test class itself cannot extend Application. We have plenty of examples of how this can be done in other tests.

3. In order to be an automated test, you need some sort of assertion checking.

tests/system/src/test/java/test/javafx/scene/text/TestHandlingBigGlyphID_Win10.java
```diff
@@ -0,0 +1,38 @@
...
+    @Override
+    public void start(Stage stage) throws Exception {
+        final String fontName = "ARIALUNI.TTF"; // Arial Unicode MS
+        // download from https://www.wfonts.com/download/data/2015/05/16/arial-unicode-ms/ARIALUNI.TTF
+        //  and place in $(user.home)/fonts/
```
kevinrushforth: An automated test should not depend on an external font that needs to be downloaded.

tests/system/src/test/java/test/javafx/scene/text/TestHandlingBigGlyphID_Win7_and_Win10.java
```diff
@@ -0,0 +1,38 @@
+import javafx.application.Application;
```
kevinrushforth: Same comments for this test. Also, do they really need to be separate tests?

modules/javafx.graphics/src/main/java/com/sun/javafx/font/directwrite/DWGlyphLayout.java
nakajima-akira: glyphs[j] is short.
I now checked following code works.
iglyphs[i] = (glyphs[j] & SHORTMASK) | slotMask;

kevinrushforth: You're right. I was looking at iglyphs[]. With a short, the cast is optional, so it's up to you whether to keep it.

tests/system/src/test/java/test/javafx/scene/text/TestHandlingBigGlyphID_Win10.java
nakajima-akira: This bug occurs only when all of the following conditions are satisfied.
1. on Windows
2. font has glyphID bigger than 32767 (max value of signed short)
3. font has complex character. (isComplexCharCode() return true)

When display characters of above conditions
, Java processes layout() in DWGlyphLayout.java
, then iglyphs[] have minus value by missing MASK.

Windows default does not have such font which is satisfied above conditions.
As far as I know, such font is "Arial Unicode MS" supplied by Microsoft Office.

tests/system/src/test/java/test/javafx/scene/text/TestHandlingBigGlyphID_Win7_and_Win10.java
nakajima-akira: For those who dislikes installing external font to Windows, I have created Test for Win10 which is not need to install font.
Test for Win7 needs to install font.

tests/system/src/test/java/test/javafx/scene/text/TestHandlingBigGlyphID_Win10.java
nakajima-akira: I today checked other tests codes.
TextTest.java and FontTest.java say
"StubFontLoader only knows about Amble."

So We can not load font "Arial Unicode MS" ?
If so, It looks impossible to create automated unit test code.

nakajima-akira: As far as I know, "Arial Unicode MS"  is best font for this tests.
I tried to load font several way in automated test, but I could not success.
Is it OK to modify these test codes and move it to tests/manual/text/ ?

nakajima-akira:
Could you review please?
I made tests/manual/text/BigGlyphIDTest.java using GNU unifont.

kevinrushforth:
Filed in JBS as: [JDK-8207839](https://bugs.openjdk.java.net/browse/JDK-8207839)

kevinrushforth:
@prrace Can you take a look at this, too?

nakajima-akira:
I thought SHORTMASK would be inline expanded, because it's final int.
By your advise, I changed to DWGlyph.SHORTMASK.
I changed (c) year to 2018 on the test code.
Thanks.

nakajima-akira:
I posted commit with prrace's comment applied.
Could you review please?
* Fix bug to handle negative glyph_id on Windows     9f89f0e

johanvos:
This might also fix https://github.com/javafxports/openjdk-jfx/issues/287 and  https://bugs.openjdk.java.net/browse/JDK-8189092

PR #134, opened 2018-07-19
JDK-8207372 Fix Robot mouse wheel implementation for Linux/Mac.

brcolow:
https://bugs.openjdk.java.net/browse/JDK-8207372

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/robot/Robot.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/robot/Robot.java
@@ -225,8 +225,8 @@ public final class Robot {
 
     /**
      * Scrolls the mouse wheel by the specified amount of wheel clicks. A positive
-     * {@code wheelAmt} scrolls the wheel towards the user (up) whereas negative
-     * amounts scrolls the wheel away from the user (down).
+     * {@code wheelAmt} scrolls the wheel towards the user (down) whereas negative
+     * amounts scrolls the wheel away from the user (up).
      *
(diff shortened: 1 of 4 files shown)


modules/javafx.graphics/src/main/java/javafx/scene/robot/Robot.java
```diff
@@ -224,8 +224,9 @@ public void mouseClick(MouseButton... buttons) {
     }
 
     /**
-     * Scrolls the mouse wheel by the specified amount of wheel clicks. Positive
-     * {@code wheelAmt} scrolls up whereas negative scrolls down.
+     * Scrolls the mouse wheel by the specified amount of wheel clicks. A positive
+     * {@code wheelAmt} scrolls the wheel towards the user (up) whereas negative
+     * amounts scrolls the wheel away from the user (down).
```
kevinrushforth: I see that the FX docs were backwards relative to AWT before this change, so "up" and "down" need to be switched to correct this (which is probably why I thought it was Windows that was wrong instead of Linux). It should be:

positive ... towards the user (down)
negative ... away from the user (up)

kevinrushforth:
> I see that this fixes both mouseWheel issues ...
Meaning both [JDK-8207372](https://bugs.openjdk.java.net/browse/JDK-8207372) and [JDK-8207371](https://bugs.openjdk.java.net/browse/JDK-8207371)
I can close JDK-8207371 as a duplicate.

brcolow:
Fixed, sorry about that. Yes, killing two birds with one stone. Thanks for marking the other as a duplicate.

brcolow:
Sure.

PR #135, opened 2018-07-19
JDK-8207370 Fix failing Monocle tests (deprecated assertEquals).

brcolow:
I have not yet confirmed that this fixes all of the failures. In other words, I haven't actually run `gradle -PUNSTABLE_TEST=true` yet.

https://bugs.openjdk.java.net/browse/JDK-8207370


johanvos:
Of the 12 failing tests, 11 are now fixed. The remaining failure is not due to the assertEquals issue, so +1 from me for this patch.

brcolow:
Thanks all.

PR #136, opened 2018-07-20
JDK-8207015: jlink javafx.graphics on Windows fails: PluginException: Duplicate resources

kevinrushforth:
[JDK-8207015](https://bugs.openjdk.java.net/browse/JDK-8207015) : jlink javafx.graphics on Windows fails: PluginException: Duplicate resources

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5589,6 +5589,12 @@ compileTargets { t ->
         def standaloneSdkDir = "${rootProject.buildDir}/${standaloneSdkDirName}"
         def standaloneLegalDir = "${standaloneSdkDir}/legal"
 
+        def excludeDLLs = []
+        if (IS_WINDOWS) {
+            excludeDLLs += targetProperties.VS2017DLLNames
+            excludeDLLs += targetProperties.WinSDKDLLNames
+        }
+
         moduleProjList.each { project ->
             def moduleName = project.ext.moduleName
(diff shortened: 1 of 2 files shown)


kevinrushforth:
This is now ready for review.

build.gradle
```diff
@@ -5589,6 +5589,12 @@ compileTargets { t ->
         def standaloneSdkDir = "${rootProject.buildDir}/${standaloneSdkDirName}"
         def standaloneLegalDir = "${standaloneSdkDir}/legal"
 
+        def excludeDLLs = []
```
johanvos: maybe this should be renamed to excludeDynamicLibs or so, as I think (fear) this issue might surface for other platforms in the future as well.

johanvos:
This doesn't break anything on Linux, will test Mac later.
I have a bigger worry about duplicate resources, but I will post that in the original issue at JBS as I think it relates more to jlink.

johanvos:
works on mac too, so +1

build.gradle
kevinrushforth: OK, I change it to excludeNativeLibs. The code that populates the list is still specific to Microsoft DLLs, but could be generalized later if there is a need.

muralibilla:
Tested in Windows..Changes looks fine..+1

PR #137, opened 2018-07-20
JDK-8208076 : display INVISIBLE_GLYPH_ID as square box on Windows

nakajima-akira:
I just now reported to [bugreport.java.com](bugreport.java.com).
Maybe 2-3days later Bug ID will be published.

I have not confidence whether "checking INVISIBLE_GLYPH_ID in this position" is correct.
-
ng_invisible


nakajima-akira:
BUG-ID was published.
[https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8208076](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8208076)

Even on the same Windows ver, whether reproduce or not depends on environments.
In my environments, It reproduced with Widows7 SP1 64bit and not reproduced with Windows 10.
This does not mean that Win7 always reproduces and Win10 does not.
I do not know what kind of environment is affecting to reproduce.

kevinrushforth:
It has to do with whether the HW pipeline or the SW pipeline is used. That depends on two factors: the graphics accelerator on the system, and whether or not you are using some sort of virtual environment / remote desktop.

You can force the SW pipeline and see the bug on any system:

```
java -Dprism.order=sw ...
```

kevinrushforth:
> BUG-ID was published.
> https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8208076

The direct URL is: https://bugs.openjdk.java.net/browse/JDK-8208076

kevinrushforth:
Please send a review request email to openjfx-dev with a pointer to the JBS bug and this PR.

kevinrushforth:
Targeted to openjfx13.

kevinrushforth:
I applied the patch, but I still see the bug.

prrace:
This is confusing / does not add up.
- Some times the bug is not seen at all. 
- Other times the bug is seen but then presumably fixed with this patch
- And for at least one person (Kevin) it is seen both with  + without the patch

I could explain the first of these as somehow being sometimes an invisible glyph isn't inserted so the problem doesn't manifest, but can't explain the last ... this needs more understanding & analysis than it has been given - I think that needs to come from the fixer.

nakajima-akira:
I forgot to write, this does not work without merging the following patches.
https://github.com/javafxports/openjdk-jfx/pull/125
```
public int getGlyphCode () {
 - return run.glyphIndices;
 + return (run.glyphIndices & SHORTMASK);
 }
```
Since getGlyphCode (before merging the above patch) returns a negative int value (0xffffffff), it does not match the INVISIBLE_GLYPH_ID (0xffff).

```
+ if (g.getGlyphCode () == CharToGlyphMapper.INVISIBLE_GLYPH_ID) {
+     return;
+ }
```

and , please don't forget to run java -Dprism.order=sw

kevinrushforth:
The above changes from #125 were merged a couple months ago. I will double-check whether I was testing with a build that includes those changes.

tests/manual/text/INVISIBLE_GLYPH_IDTest.java
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
```
kevinrushforth: Can you update this to 2019?

nakajima-akira: I have no time today, so I will do it tomorrow.

nakajima-akira: I did re-push to https://github.com/javafxports/openjdk-jfx/commit/d2edbed328a753f536bcb4c92cae874b2097646f.

Only INVISIBLE_GLYPH_IDTest.java is changed.
Copyright 2018 -> 2019
and added replace("\\", "/") to run collectry on Windows10 1809.
`userHome = userHome.replace("\\", "/");`

kevinrushforth:
@nakajima-akira Please do not force-push to a branch once you have published a review. It makes it very difficult to review incremental changes such as this. See [CONTRIBUTING.md](https://github.com/javafxports/openjdk-jfx/blob/develop/.github/CONTRIBUTING.md) for guidelines surrounding force pushes:

> Please adhere to the general guideline that you should never force push to a publicly shared branch. Once you have opened your pull request, you should consider your branch publicly shared. Instead of force pushing you can just add incremental commits; this is generally easier on your reviewers. If you need to pick up changes from develop, you can merge develop into your branch.

There might be a slight delay in reviewing this, since I need to double-check that nothing other than the copyright year has changed (I will also look at the other change you made).

nakajima-akira:
Sorry, I understand.

PR #140, opened 2018-07-21
JDK-8207377 Document behavior of getPixelColor on HiDPI.

brcolow:
https://bugs.openjdk.java.net/browse/JDK-8207377


modules/javafx.graphics/src/main/java/javafx/scene/robot/Robot.java
```diff
@@ -251,7 +251,12 @@ public Color getPixelColor(double x, double y) {
```
kevinrushforth: The fix looks good. There are two getPixelColor methods in Robot.java. Can you also update the docs for the other one?

brcolow: I knew I forgot something, lol.

PR #147, opened 2018-08-04
JDK-8207328 Fix Stylesheet Javadocs

wowselim:
I attempted to improve the javadoc comments a bit to more match the underlying code. While reading some of the code I noticed that for example in the `convertToBinary` some input / output stream might potentially not be closed properly. I would also like to fix that if desired.

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
@@ -67,7 +67,7 @@ public class Stylesheet {
 
     private final String url;
     /** The URL from which the stylesheet was loaded.
-     * @return The URL from which the stylesheet was loaded, or null if
+     * @return A String version of the URL from which the stylesheet was loaded, or null if
      *         the stylesheet was created from an inline style.
      */
     public String getUrl() {
@@ -238,7 +238,7 @@ public class Stylesheet {
     private String[] stringStore;
     final String[] getStringStore() { return stringStore; }
 
-    /** Load a binary stylesheet file from a input stream.
+    /** Loads a binary stylesheet from a URL.
      * @param url the URL from which the stylesheet will be loaded
      * @return the constructed stylesheet
      * @throws java.io.IOException the exception
@@ -293,9 +293,9 @@ public class Stylesheet {
     }
 
     /**
-     * Convert the .css file referenced by urlIn to binary format and write to urlOut.
-     * @param source is the JavaFX .css file to convert
-     * @param destination is the file to which the binary conversion is written
+     * Convert the .css file referenced by source to binary format and write to destination.
+     * @param source the JavaFX .css file to convert
+     * @param destination the file to which the binary version is written
      * @throws IOException the exception
      * @throws IllegalArgumentException if either parameter is null, if source and destination are the same,
      * if source cannot be read, or if destination cannot be written.
--- a/modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
@@ -67,7 +67,7 @@ public class Stylesheet {
 
     private final String url;
     /** The URL from which the stylesheet was loaded.
-     * @return A String version of the URL from which the stylesheet was loaded, or null if
+     * @return A String representation of the URL from which the stylesheet was loaded, or null if
      *         the stylesheet was created from an inline style.
      */
     public String getUrl() {
@@ -131,9 +131,9 @@ public class Stylesheet {
     }
 
     /**
-     * Constructs a Stylesheet using the given URL as the base URI. The
+     * Constructs a Stylesheet interpreting the given String as the base URI. The
      * parameter may not be null.
-     * @param url
+     * @param url the base URI for this stylesheet.
      */
     Stylesheet(String url) {
 
@@ -239,9 +239,10 @@ public class Stylesheet {
     final String[] getStringStore() { return stringStore; }
 
     /** Loads a binary stylesheet from a URL.
-     * @param url the URL from which the stylesheet will be loaded
-     * @return the constructed stylesheet
-     * @throws java.io.IOException the exception
+     * @param url the URL from which the stylesheet will be loaded.
+     * @return the loaded stylesheet.
+     * @throws java.io.IOException if the binary stream corresponds to a more recent binary
+     * css version or if any other IO error occurs while reading from the stream.
      */
     public static Stylesheet loadBinary(URL url) throws IOException {
 
@@ -296,7 +297,7 @@ public class Stylesheet {
      * Convert the .css file referenced by source to binary format and write to destination.
      * @param source the JavaFX .css file to convert
      * @param destination the file to which the binary version is written
-     * @throws IOException the exception
+     * @throws IOException if the destination file can not be created or any other IO error occurs.
      * @throws IllegalArgumentException if either parameter is null, if source and destination are the same,
      * if source cannot be read, or if destination cannot be written.
      */
--- a/modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
@@ -66,7 +66,9 @@ public class Stylesheet {
     final static int BINARY_CSS_VERSION = 6;
 
     private final String url;
-    /** The URL from which the stylesheet was loaded.
+    /**
+     *  The URL from which this {@code Stylesheet} was loaded.
+     *
      * @return A String representation of the URL from which the stylesheet was loaded, or null if
      *         the stylesheet was created from an inline style.
      */
@@ -75,11 +77,17 @@ public class Stylesheet {
     }
 
     /**
-     * True if this style came from user stylesheet, we need to know this so
+     * Specifies the origin of this {@code Stylesheet}. We need to know this so
      * that we can make user important styles have higher priority than
-     * author styles
+     * author styles.
      */
     private StyleOrigin origin = StyleOrigin.AUTHOR;
+
+    /**
+     * The origin of this {@code Stylesheet}.
+     *
+     * @return the origin of this {@code Stylesheet}
+     */
     public StyleOrigin getOrigin() {
         return origin;
     }
@@ -131,9 +139,10 @@ public class Stylesheet {
     }
 
     /**
-     * Constructs a Stylesheet interpreting the given String as the base URI. The
+     * Constructs a Stylesheet using the given URL as the base URI. The
      * parameter may not be null.
-     * @param url the base URI for this stylesheet.
+     *
+     * @param url the base URI for this stylesheet
      */
     Stylesheet(String url) {
 
@@ -141,10 +150,20 @@ public class Stylesheet {
 
     }
 
+    /**
+     * The rules that are defined in this {@code Stylesheet}
+     *
+     * @return a list of rules used by this {@code Stylesheet}
+     */
     public List<Rule> getRules() {
         return rules;
     }
 
+    /**
+     * The font faces used by this {@code Stylesheet}
+     *
+     * @return a list of font faces used by this {@code Stylesheet}
+     */
     public List<FontFace> getFontFaces() {
         return fontFaces;
     }
@@ -238,11 +257,13 @@ public class Stylesheet {
     private String[] stringStore;
     final String[] getStringStore() { return stringStore; }
 
-    /** Loads a binary stylesheet from a URL.
-     * @param url the URL from which the stylesheet will be loaded.
-     * @return the loaded stylesheet.
-     * @throws java.io.IOException if the binary stream corresponds to a more recent binary
-     * css version or if any other IO error occurs while reading from the stream.
+    /**
+     * Loads a binary stylesheet from a URL.
+     *
+     * @param url the URL from which the {@code Stylesheet} will be loaded
+     * @return the loaded {@code Stylesheet}
+     * @throws IOException if the binary stream corresponds to a more recent binary
+     * css version or if an I/O error occurs while reading from the stream
      */
     public static Stylesheet loadBinary(URL url) throws IOException {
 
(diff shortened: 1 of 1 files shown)


modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -238,10 +238,11 @@ final void readBinary(int bssVersion, DataInputStream is, String[] strings)
     private String[] stringStore;
     final String[] getStringStore() { return stringStore; }
 
-    /** Load a binary stylesheet file from a input stream.
-     * @param url the URL from which the stylesheet will be loaded
-     * @return the constructed stylesheet
-     * @throws java.io.IOException the exception
+    /** Loads a binary stylesheet from a URL.
+     * @param url the URL from which the stylesheet will be loaded.
```
nlisker: No need to add a `.`  because this is a phrase (@param and @return are usually in this form).

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -238,10 +238,11 @@ final void readBinary(int bssVersion, DataInputStream is, String[] strings)
...
+     * @return the loaded stylesheet.
```
nlisker: No need for a `.`.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -238,10 +238,11 @@ final void readBinary(int bssVersion, DataInputStream is, String[] strings)
...
+     * @throws java.io.IOException if the binary stream corresponds to a more recent binary
```
nlisker: No need for a fully qualified name: `IOException` is enough.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -131,9 +131,9 @@ public void setOrigin(StyleOrigin origin) {
     }
 
     /**
-     * Constructs a Stylesheet using the given URL as the base URI. The
+     * Constructs a Stylesheet interpreting the given String as the base URI. The
```
nlisker: This is not public API and the constructor comment about `null` input is not exactly right. I suggest to not touch this at all until changes to the actual code of the class are made.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -293,10 +294,10 @@ public static Stylesheet loadBinary(URL url) throws IOException {
     }
 
     /**
-     * Convert the .css file referenced by urlIn to binary format and write to urlOut.
-     * @param source is the JavaFX .css file to convert
-     * @param destination is the file to which the binary conversion is written
-     * @throws IOException the exception
+     * Convert the .css file referenced by source to binary format and write to destination.
```
nlisker: "Convert" should be "Converts".
"write" should be "writes".

nlisker: No need for a `.` in front of `css`.

nlisker: "source" should be `{@code source}` since it's a method argument. Same for "destination".

nlisker: Add an empty line between the description and the tags.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -293,10 +294,10 @@ public static Stylesheet loadBinary(URL url) throws IOException {
...
+     * @param source the JavaFX .css file to convert
```
nlisker: No need for a `.` in front of `css`.

nlisker: Maybe "the JavaFX *compliant* css file" is a better phrasing?

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -293,10 +294,10 @@ public static Stylesheet loadBinary(URL url) throws IOException {
...
+     * @param destination the file to which the binary version is written
```
nlisker: "binary formatted data" is probably better than "binary version" because the latter can be mistaken for a version number

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -293,10 +294,10 @@ public static Stylesheet loadBinary(URL url) throws IOException {
...
+     * @throws IOException if any IO error occurs.
```
nlisker: "if an I/O error occurs" is the standard phrasing.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -293,10 +294,10 @@ public static Stylesheet loadBinary(URL url) throws IOException {
...
      * @throws IllegalArgumentException if either parameter is null, if source and destination are the same,
```
nlisker: `null`, `source` and `destination` should use `@code`.

nlisker:
> While reading some of the code I noticed that for example in the `convertToBinary` some input / output stream might potentially not be closed properly. I would also like to fix that if desired.

That would require a new JBS issue. The class design could be improved as well:

* `loadBinary` should probably not accept `null` (throw an NPE).
* The empty constructor makes no sense.
* The `String` accepts `null` even if it says it doesn't.
* Class-level documentation on how to work with the class could then be added.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -238,10 +238,11 @@ final void readBinary(int bssVersion, DataInputStream is, String[] strings)
     private String[] stringStore;
     final String[] getStringStore() { return stringStore; }
 
-    /** Load a binary stylesheet file from a input stream.
-     * @param url the URL from which the stylesheet will be loaded
-     * @return the constructed stylesheet
-     * @throws java.io.IOException the exception
+    /** Loads a binary stylesheet from a URL.
```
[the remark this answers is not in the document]
kevinrushforth: Also, as long as you are touching this code, the `/**` should be on a line by itself.

kevinrushforth:
@wowselim Please also send a review request to openjfx-dev@openjdk.java.net with a pointer to the JBS bug and this PR so that it can be pushed to jfx-dev once the review is done.

kevinrushforth:
@wowselim are you still interested in fixing this bug for openjfx 11?

wowselim:
Sorry guys, this whole process just took way longer than I initially anticipated. I tried to improve on all the points that you have mentioned. Additionally I changed most occurences of `Stylesheet` to `{@code Stylesheet}` since I saw the same thing in the Stage class

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -238,11 +257,13 @@ final void readBinary(int bssVersion, DataInputStream is, String[] strings)
...
-     * @return the loaded stylesheet.
-     * @throws java.io.IOException if the binary stream corresponds to a more recent binary
-     * css version or if any other IO error occurs while reading from the stream.
+    /**
+     * Loads a binary stylesheet from a URL.
+     *
+     * @param url the URL from which the {@code Stylesheet} will be loaded
+     * @return the loaded {@code Stylesheet}
+     * @throws IOException if the binary stream corresponds to a more recent binary
+     * css version or if an I/O error occurs while reading from the stream
```
nlisker: Missing a `.` at the end of `@throws` description.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -294,12 +315,14 @@ public static Stylesheet loadBinary(URL url) throws IOException {
...
-     * @throws IllegalArgumentException if either parameter is null, if source and destination are the same,
-     * if source cannot be read, or if destination cannot be written.
+     * Convert the css file referenced by source to binary format and write to destination.
+     *
+     * @param source the JavaFX compliant css file to convert
+     * @param destination the file to which the binary formatted data is written
+     * @throws IOException if the destination file can not be created or if an I/O error occurs
+     * @throws IllegalArgumentException if either parameter is null, if {@code source} and
+     * {@code destination} are the same, if {@code source} cannot be read, or if {@code destination}
+     * cannot be written
```
nlisker: Missing `.`  at the end of `@throws` description.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -294,12 +315,14 @@ public static Stylesheet loadBinary(URL url) throws IOException {
     }
 
     /**
-     * Convert the .css file referenced by source to binary format and write to destination.
-     * @param source the JavaFX .css file to convert
-     * @param destination the file to which the binary version is written
-     * @throws IOException if any IO error occurs.
-     * @throws IllegalArgumentException if either parameter is null, if source and destination are the same,
-     * if source cannot be read, or if destination cannot be written.
+     * Convert the css file referenced by source to binary format and write to destination.
```
nlisker: "Convert" should be "Converts".
"write" should be "writes".

nlisker: "source" should be `{@code source}` since it's a method argument. Same for "destination".

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -294,12 +315,14 @@ public static Stylesheet loadBinary(URL url) throws IOException {
...
+     *
+     * @param source the JavaFX compliant css file to convert
+     * @param destination the file to which the binary formatted data is written
+     * @throws IOException if the destination file can not be created or if an I/O error occurs
+     * @throws IllegalArgumentException if either parameter is null, if {@code source} and
```
nlisker: `null` should use `@code` style.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -238,11 +257,13 @@ final void readBinary(int bssVersion, DataInputStream is, String[] strings)
...
 
-    /** Loads a binary stylesheet from a URL.
-     * @param url the URL from which the stylesheet will be loaded.
-     * @return the loaded stylesheet.
-     * @throws java.io.IOException if the binary stream corresponds to a more recent binary
-     * css version or if any other IO error occurs while reading from the stream.
+    /**
+     * Loads a binary stylesheet from a URL.
+     *
+     * @param url the URL from which the {@code Stylesheet} will be loaded
```
nlisker: `URL` should use `@code`.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
kevinrushforth: Typically we don't end with a period for `@param`, `@return`, or `@throws`. I recommend leaving this one as is.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
kevinrushforth: Same comment as above (recommend to leave it alone).

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -66,20 +66,28 @@
...
-     * True if this style came from user stylesheet, we need to know this so
+     * Specifies the origin of this {@code Stylesheet}. We need to know this so
      * that we can make user important styles have higher priority than
-     * author styles
+     * author styles.
      */
     private StyleOrigin origin = StyleOrigin.AUTHOR;
+
+    /**
+     * The origin of this {@code Stylesheet}.
```
kevinrushforth: The initial sentence of a method's description should usually start with a verb. In this case, something like `Gets the origin ...` or `Returns the origin ...` or `Retrieves the origin ...` would be fine.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -133,18 +141,29 @@ public void setOrigin(StyleOrigin origin) {
...
+     * @param url the base URI for this stylesheet
      */
     Stylesheet(String url) {
 
         this.url = url;
 
     }
 
+    /**
+     * The rules that are defined in this {@code Stylesheet}
```
kevinrushforth: Same comment as for getOrigin about starting with a verb. Also, the description should end with a period.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -133,18 +141,29 @@ public void setOrigin(StyleOrigin origin) {
...
+     *
+     * @return a list of rules used by this {@code Stylesheet}
+     */
     public List<Rule> getRules() {
         return rules;
     }
 
+    /**
+     * The font faces used by this {@code Stylesheet}
```
kevinrushforth: Start with a verb and end with a period.

kevinrushforth:
@wowselim Do you plan to finish this PR soon?
The deadline for getting doc changes in for openjfx 11 is Monday, 27-Aug-2018. Otherwise it will be for openjfx 12.

kevinrushforth:
> @wowselim Please also send a review request to openjfx-dev@openjdk.java.net with a pointer to the JBS bug and this PR so that it can be pushed to jfx-dev once the review is done.

The above still needs to be done. The subject can be something like:

Subject: [11] RFR: 8207328: API docs for javafx.css.Stylesheet are inaccurate / wrong

kevinrushforth:
I'll finish my review later today.

nlisker:
+1 from me as well.

PR #152, opened 2018-08-09
JDK-8148129: Implement Accelerated composition for WebView (only using NGCamera)

ararunprasad:
This is an another implementation of Accelerated composition with minimal changes to javafx.graphics module.

Though it works as expected, I'm not satisfied with the workarounds made around WCCamera to propagate perspectiveTransform to respective 3D contexts.


ararunprasad:
This pull request is meant only for bookkeeping purpose. It is a duplicate of #51

PR #153, opened 2018-08-10
Add a powershell script for Windows dev, use on Appveyor.

brcolow:
The script `tools\scripts\build.ps1` works for locally building OpenJFX without any additional environment setup as long as one has the [choco](https://chocolatey.org/) package manager installed. It automatically finds the location of the latest installed versions of Visual Studio and the Windows SDK and sets the necessary environment variables so that the gradle build works out-of-the-box without requiring somewhat tedious manual setup/configuration.

### DIFF
--- a/tools/scripts/build.ps1
+++ b/tools/scripts/build.ps1
@@ -1,0 +1,15 @@
+choco install ant
+choco install vswhere
+choco install zip
+$env:WINSDK_DIR = (Get-ItemProperty -Path "HKLM:\SOFTWARE\Wow6432Node\Microsoft\Windows Kits\Installed Roots").KitsRoot10
+$env:VCINSTALLDIR = "$(vswhere -legacy -latest -property installationPath)\VC\Auxiliary\Build"
+$msvcToolsVer = Get-Content "$env:VCINSTALLDIR\Microsoft.VCToolsVersion.default.txt"
+if ([string]::IsNullOrWhitespace($msvcToolsVer)) {
+  # The MSVC tools version file can have txt *or* props extension.
+  $msvcToolsVer = Get-Content "$env:VCINSTALLDIR\Microsoft.VCToolsVersion.default.props"
+}
+$env:MSVC_VER = $msvcToolsVer
+$env:VS150COMNTOOLS = $env:VCINSTALLDIR
+$env:VSVARS32FILE = "$env:VCINSTALLDIR\vcvars32.bat"
+refreshenv
+.\gradlew all test -PCOMPILE_WEBKIT=false -PCONF=DebugNative --stacktrace -x :web:test --info
--- a/tools/scripts/build.ps1
+++ b/tools/scripts/build.ps1
@@ -12,4 +12,9 @@ $env:MSVC_VER = $msvcToolsVer
 $env:VS150COMNTOOLS = $env:VCINSTALLDIR
 $env:VSVARS32FILE = "$env:VCINSTALLDIR\vcvars32.bat"
 refreshenv
-.\gradlew all test -PCOMPILE_WEBKIT=false -PCONF=DebugNative --stacktrace -x :web:test --info
+if ($env:APPVEYOR -eq "true") {
+  .\gradlew all test -PCOMPILE_WEBKIT=false -PCONF=DebugNative --stacktrace -x :web:test --info --no-daemon
+} else {
+  .\gradlew all test -PCOMPILE_WEBKIT=false -PCONF=DebugNative --stacktrace -x :web:test --info
+}
+
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,36 +4,18 @@ image: Visual Studio 2017
 
 environment:
   JAVA_HOME: C:\jdk10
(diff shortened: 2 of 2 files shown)


tools/scripts/build.ps1
```diff
@@ -0,0 +1,15 @@
+choco install ant
```
kevinrushforth: This would be the first time a "github-only" file showed up in this directory. I'd like to limit the 'github-only' files to as few locations as possible. This raises an interesting question of whether these sort of build files should eventually be committed to the HG openjdk repo, in which case it wouldn't matter.

Pending that more global discussion, perhaps there is a better place to put it? In a new appveyor dir maybe? Not sure (would like to think about this for more time than I have today).

Maybe one of the Gluon folks have some thoughts on this?

kevinrushforth: This would be the first time a "github-only" file showed up in this directory. I'd like to limit the 'github-only' files to as few locations as possible. This raises an interesting question of whether these sort of build files should eventually be committed to the HG openjdk repo, in which case it wouldn't matter.

Pending that more global discussion, perhaps there is a better place to put it? In a new appveyor dir maybe? Not sure (would like to think about this for more time than I have today).

Maybe one of the Gluon folks have some thoughts on this?

johanvos: It's a non-trivial question. In general, I am leaning towards separating code and build logic, since there might be multiple usages of the same source.
In this case however, the file is intended to be added for test purposes, hence I think it makes sense to do it here.
It is also not a github specific file, so I think it makes sense adding it to the hg repository as well.

kevinrushforth:
In that case, I am OK with this as long as we can keep the changes that do not currently go back to the HG repo on OpenJDK (the `appveyor.yml` script) separate from those that do (files the `tools/scripts` dir). Can you file a JBS bug to add the `build.ps1` script?

This brings up a point I've been thinking about for a while. We really should eliminate (or at least minimize) any differences between the HG repo and the `master` branch on GitHub.

I'll file a JBS issue to consider unifying them.

brcolow:
https://bugs.openjdk.java.net/browse/JDK-8209765

kevinrushforth:
Thanks for filing the bug. That's fine about mentioning appveyor (I also added a comment for those looking at it indicating that the appveyor.yml file will be excluded from the openjfx/jfx-dev/rt changeset).

I fixed up a couple things, assigned it to you, and linked it to this PR (by adding the `github-bug` label and adding an issue Link to the URL of this PR).

Can you send a review request to openjfx-dev with a pointer to this PR so any other interested party can also review it?

kevinrushforth:
As long as the Appveyor build is passing with the new script, it looks fine to me.

tools/scripts/build.ps1
```diff
@@ -0,0 +1,15 @@
...
+$msvcToolsVer = Get-Content "$env:VCINSTALLDIR\Microsoft.VCToolsVersion.default.txt"
+if ([string]::IsNullOrWhitespace($msvcToolsVer)) {
+  # The MSVC tools version file can have txt *or* props extension.
+  $msvcToolsVer = Get-Content "$env:VCINSTALLDIR\Microsoft.VCToolsVersion.default.props"
+}
+$env:MSVC_VER = $msvcToolsVer
+$env:VS150COMNTOOLS = $env:VCINSTALLDIR
+$env:VSVARS32FILE = "$env:VCINSTALLDIR\vcvars32.bat"
+refreshenv
+.\gradlew all test -PCOMPILE_WEBKIT=false -PCONF=DebugNative --stacktrace -x :web:test --info
```
kevinrushforth: I recommend adding `--no-daemon` which was present in the appveyor.yml file.

brcolow: For CI builds I wanted to use `--no-daemon`. For local builds it makes sense to use the daemon IMO. So in the latest commit we check if `APPVEYOR` environment variable is set. If so, use `--no-daemon`. Otherwise, don't. Is that okay?

brcolow: I broke it, let me fix it.

brcolow: Actually, strange, it seems that the downloading of OpenJDK 10 is what failed. Let me see.

kevinrushforth:
I see the following from the Appveyor log (looks like it is a dependency of ant), but it doesn't look like this is new:

```
jdk8 v8.0.181 [Approved]
jdk8 package files install completed. Performing other installation steps.
Downloading JDK from http://download.oracle.com/otn-pub/java/jdk/8u181-b13/96a7b8442fe848ef90c96a2fad6ed6d1/jdk-8u181-windows-x64.exe
Installing jdk8...
jdk8 has been installed.
PATH environment variable does not have C:\Program Files\Java\jdk1.8.0_181\bin in it. Adding...
  jdk8 may be able to be automatically uninstalled.
Environment Vars (like PATH) have changed. Close/reopen your shell to
 see the changes (or in powershell/cmd.exe just type `refreshenv`).
 The install of jdk8 was successful.
  Software installed to 'C:\Program Files\Java\jdk1.8.0_181\'
```

In any event, it isn't using JDK 8, so it's not a problem. I'll merge the PR now.

PR #155, opened 2018-08-20
JDK-8209764: JavaFX/Monocle - Partial Screen Capture Broken

brcolow:
When capturing a partial region of the screen the `data` array elements have the wrong positions. This leads to an ArrayIndexOutOfBoundsException.

I introduced this error in the robot public API pull request.

This PR fixes it so that the `data` array indices match the correct pixels from the screen buffer.

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/monocle/MonocleRobot.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/monocle/MonocleRobot.java
@@ -214,7 +214,7 @@ class MonocleRobot extends GlassRobot {
             int colStop = Math.min(x + width, scrWidth);
             for (int row = y; row < rowStop; row++) {
                 for (int col = x; col < colStop; col++) {
-                    data[row * scrWidth + col] = buffer.get(row * scrWidth + col);
+                    data[(row - y) * (colStop - x) + (col - x)] = buffer.get(row * scrWidth + col);
                 }
             }
         }


kevinrushforth:
Is there a JBS bug filed? If not, can you file one and then link it to this PR?

I presume that this only affects Monocle, and that the other glass platforms do not have this bug?

brcolow:
I will file one and link to the PR.

This only affects Monocle.

kevinrushforth:
Thanks for filing the bug. I fixed up a couple things, assigned it to you, and linked it to this PR (by adding the `github-bug` label and adding an issue Link to the URL of this PR).

Can you send a review request to openjfx-dev with a pointer to this PR so any other interested party can also review it?

johanvos:
seems a correct fix indeed.

kevinrushforth:
It's already there: [http://hg.openjdk.java.net/openjfx/jfx-dev/rt/rev/d8330260200b](http://hg.openjdk.java.net/openjfx/jfx-dev/rt/rev/d8330260200b)

Or were you asking about openjfx11? It was too late for 11 (it was the final week of RDP2 and we were only looking at looking at critical P2 bugs).

brcolow:
Will there be a "sub-repo" (?) opened for 12, as there is currently for the other versions? That way I can get a tag (say, 12+1)?

kevinrushforth:
We won't fork a 12 repo until some time during rampdown when we go to stabilize the release (presuming we follow a slightly-modified OpenJDK 12 schedule, that will likely be in December or January). Prior to that we will just do builds of 12 from the mainline repo, which is what we did for 11 until about 5 weeks ago.

The reason there hasn't been a build of openjfx 12, along with the "12+1" tag, is that we haven't started doing any early access builds yet. It seems like it might be about time to start doing that now that we aren't doing weekly builds of 11 any more (there will be one more "final" build of 11, but nothing else is planned).

brcolow:
Alright, great. Thanks for the info.

PR #162, opened 2018-08-22
Add Automatic-Module-Name to empty jars

### LINKED ISSUE #161: empty jars should be modules
see https://bugs.openjdk.java.net/browse/JDK-8209836

johanvos:
This fixes #161


kevinrushforth:
Go ahead and 'Squash & Merge' this PR as soon as you are satisfied with it.

kevinrushforth:
Since this needs to be pushed to HG openjfx/11-dev/rt anyway, there is probably no need to actually merge this PR into develop (it will be automatically synced there soon after). It's up to you.

Issue #151, opened 2018-08-08
JDK-8207942: Add new protected VirtualFlow methods for subclassing


eugener:
Related JBS issues:
https://bugs.openjdk.java.net/browse/JDK-8207942
https://bugs.openjdk.java.net/browse/JDK-8207957

From JDK 9 onward it is not possible to subclass VirtualFlow and many libraries/applications cannot produce adequate releases. This can finally be fixed in OpenJFX 12.

Maxoudela:
I can provide a pull request if you like.

kevinrushforth:
What we need first is a discussion on the API. A [WIP] pull request might be a starting point for that discussion, but since this will be new feature with new public API, it will need a [CSR](https://wiki.openjdk.java.net/display/csr/Main). I think @eugener plans to do this, so please wait for him to weigh in on this.

eugener:
In my view there are two approaches possible: 

1. Make some of the existing methods protected to allow for direct extension of `VirtualFlow` as Samir suggests
2. Change the `VirtualFlow` to use a pluggable implementations, i.e define APIs based on interfaces which can be plugged into `VirtualFlow`, so it can call them when required. 

Samir (@Maxoudela) is the one who filed the ttps://bugs.openjdk.java.net/browse/JDK-8207942, So I would be ok if we start the discussion with his PR, provided that approach 1 is deemed to be more favorable.

kevinrushforth:
Approach 1 might be be a quicker path, and it might be nice to see what such an API would look like, so if Samir wants to generate a PR as a discussion point, that seems fine.

@eugener - are you interested in exploring option 2? If so, that might be better done over email rather than a prototype, but even that might possibly be done as a [WIP] PR. One question though: you say "change" VirtualFlow. You mean compatibly, I hope?

eugener:
@kevinrushforth As much as it is possible it should be fully compatible. The internals of it will simply call the interface methods. This is all hypothetical of course. reality might turn out to be different :)

I will try to experiment with it, as long as you think it might be somewhat viable approach.

kevinrushforth:
@eugener If you think there might be value in it, and if the resulting API would be something that would "fit" with the existing API, then you might spend a little time prototyping it to see what it would look like.

Maxoudela:
I have not forgotten you guys. I'll make the PR as soon as possible as a discussion point. 
I just need to install and build OpenJFX, and also make the PR the right way. 
[...]

PR #163, opened 2018-08-22
JDK-8207942: Add protected API to VirtualFlow and other related classes for sublclassing

Maxoudela:
Here is a first draft of the changes I mentioned in https://bugs.openjdk.java.net/browse/JDK-8207942


kevinrushforth:
This PR is for discussion related to issue #151

kevinrushforth:
@JonathanGiles if you want to weigh in on this, your insight would be appreciated.

kevinrushforth:
One method caught me eye:

```
    /**
     * This method is an experts-only method - if the requested index is not
     * already an existing visible cell, it will create a cell for the
     * given index and insert it into the sheet. From that point on it will be
     * unmanaged, and is up to the caller of this method to manage it.
     *
     * @param index
     * @return
     */
    protected T getPrivateCell(int index)  {
```

The implications to sub-classes needs to be thought through and documented.

In any case, the next steps are documenting all of the new public / protected methods (as Eugene indicated), getting agreement that the new API is what we want, and then filing a CSR once it is close to what we want and fully documented.

Thanks.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1735,10 +1735,10 @@ protected int getCellIndex(T cell){
      *                                                                         *
      **************************************************************************/
 
-    final VirtualScrollBar getHbar() {
+    protected final VirtualScrollBar getHbar() {
```
jperedadnr: `com.sun.javafx.scene.control.VirtualScrollBar` is private, so unless I'm missing something, adding `protected` here won't be of any use.

Either you move the control to public `javafx.scene.control.VirtualScrollBar`, or you return the superclass `protected final ScrollBar getHbar()`?

I guess the latter is good enough, as usually there is no need to modify the current implementation.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
```diff
@@ -314,9 +314,15 @@ final BooleanProperty reorderingProperty() {
     private final ReadOnlyObjectProperty<NestedTableColumnHeader> rootHeaderProperty() {
         return rootHeader.getReadOnlyProperty();
     }
-    final NestedTableColumnHeader getRootHeader() {
+
+    /**
+     * @return The root header that is actually just one NestedTableColumnHeader that spans
+     * the entire width.
+     */
+    protected final NestedTableColumnHeader getRootHeader() {
```
jperedadnr: Shouldn't you also remove `final`? Otherwise it can't be subclassed.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
```diff
@@ -816,18 +820,24 @@ private void refreshView() {
...
 
-    // Handles the horizontal scrolling when the selection mode is cell-based
-    // and the newly selected cell belongs to a column which is not totally
-    // visible.
+    /**
+     * Handles the horizontal scrolling when the selection mode is cell-based
+     * and the newly selected cell belongs to a column which is not totally
+     * visible.
+     */
     void scrollHorizontally() {
```
jperedadnr: This method should be `protected`, and so should be `onFocusPreviousCell`, `onFocusNextCell`, `onSelectPreviousCell`, and `onSelectNextCell`.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
```diff
@@ -314,9 +329,15 @@ final BooleanProperty reorderingProperty() {
     private final ReadOnlyObjectProperty<NestedTableColumnHeader> rootHeaderProperty() {
         return rootHeader.getReadOnlyProperty();
     }
-    final NestedTableColumnHeader getRootHeader() {
+
+    /**
+     * @return The root header that is actually just one NestedTableColumnHeader that spans
+     * the entire width.
+     */
+    public final NestedTableColumnHeader getRootHeader() {
```
Maxoudela: I initially made this protected but I'm using it elsewhere so I believe it should be made public.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
```diff
@@ -688,7 +704,7 @@ int onScrollPageDown(boolean isFocusDriven) {
      * Function used to scroll the container up by one 'page', although
      * if this is a horizontal container, then the scrolling will be to the left.
      */
-    int onScrollPageUp(boolean isFocusDriven) {
+    protected int onScrollPageUp(boolean isFocusDriven) {
```
Maxoudela: Actually, all these methods should be protected if anyone wants to implement the behavior. If we look at TableViewSkin, it does the following :
/ init the behavior 'closures'
        behavior.setOnFocusPreviousRow(() -> onFocusPreviousCell());
        behavior.setOnFocusNextRow(() -> onFocusNextCell());
        behavior.setOnMoveToFirstCell(() -> onMoveToFirstCell());
        behavior.setOnMoveToLastCell(() -> onMoveToLastCell());
        behavior.setOnScrollPageDown(isFocusDriven -> onScrollPageDown(isFocusDriven));
        behavior.setOnScrollPageUp(isFocusDriven -> onScrollPageUp(isFocusDriven));
        behavior.setOnSelectPreviousRow(() -> onSelectPreviousCell());
        behavior.setOnSelectNextRow(() -> onSelectNextCell());
        behavior.setOnSelectLeftCell(() -> onSelectLeftCell());
        behavior.setOnSelectRightCell(() -> onSelectRightCell());

so if we want to do the same, all these methods should be protected.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
```diff
@@ -837,7 +837,7 @@ private void refreshView() {
      * and the newly selected cell belongs to a column which is not totally
      * visible.
      */
-    protected void scrollHorizontally() {
+    public void scrollHorizontally() {
```
Maxoudela: This method is also called from various place and should be public IMHO.

kevinrushforth:
As a general comment, each new protected or public method that you are proposing needs to be documented and needs an `@since 12` javadoc tag.

Given that the newly-added protected and public methods will become part of the supported API, we need good justification for each new method.

Maxoudela:
I'll add the @since 12 tag for each method. I tried to document each of them, any help is appreciated regarding the comments.

I finally managed to build and test my JDK version against the SpreadsheetView. I'll add a comment on each method explaining why it should be made public/protected and add some links to SpreadsheetView code for better comprehension.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
Maxoudela: There are several methods gravitating around the TableView headers. 

One should understand that the TableView header is composed of one TableHeaderRow, that contains a single NestedTableColumnHeader that spans the whole width, and then several NestedTableColumnHeader that will represent each column.

Anyone who truly wants to tweak the TableView header will have to access to any methods that are linking any of these objects together. That includes getTableHeaderRow() that allow a custom TableColumnHeader to retrieve the TableHeaderRow is is contained in.

It also includes the getRootHeader() method that allow a TableColumnHeader to retrieve the root NestedTableCOlumnHeader containing all tableColumnHeader.
For example, getRootHeade() allow to add specific listener when the header is clicked to trigger a selection : https://bitbucket.org/controlsfx/controlsfx/src/1334a58d91017c1a1c4c13d9d2bb780d2fa3223c/controlsfx/src/main/java/impl/org/controlsfx/spreadsheet/HorizontalHeader.java?at=default&fileviewer=file-view-default#HorizontalHeader.java-185

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
```diff
@@ -378,9 +399,10 @@ private final void setRootHeader(NestedTableColumnHeader value) {
         return snappedTopInset() + headerPrefHeight + snappedBottomInset();
     }
 
-    // used to be protected to allow subclasses to modify the horizontal scrolling,
-    // but made private again for JDK 9
-    void updateScrollX() {
+       /**
+     * This method is called whenever a scroll is happening on the horizontal bar.
+     */
+    protected void updateScrollX() {
```
Maxoudela: Each time a scroll is happening, this method is called in order to shift the TableColumnHeaders. 
If anyone has overriden the TableHeader to provide another functionality, it must also be warned that a scroll is happening.

For example, you can add some icons above each TableColumnHeader, or even remove the columnHeader and simply have some icons. But you would want them to slide when the scrollBar is touched, thus having access to updateScrollX().

pickers

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
```diff
@@ -816,18 +832,24 @@ private void refreshView() {
...
         TC col = getFocusedCell().getTableColumn();
         scrollHorizontally(col);
     }
 
-    void scrollHorizontally(TC col) {
+    /**
+     * Programmatically scroll to the given column.
+     * @param col
+     */
+    protected void scrollHorizontally(TC col) {
```
Maxoudela: The scrollHorizontally() and scrollHorizontally(TC col) methods are called to respectively scroll to the selected column, or the given column.
This can be used at various places but some examples are directly linked to the methods above. If onSelectPreviousCell or onSelectNextCell is called, we may want to force the scroll to adapt in order to make the newly selected cell visible.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1735,10 +1735,19 @@ protected int getCellIndex(T cell){
...
+    protected final ScrollBar getHbar() {
         return hbar;
     }
-    final VirtualScrollBar getVbar() {
+
+    /**
+     *
+     * @return The scroll bar used for scrolling vertically.
+     */
+    protected final ScrollBar getVbar() {
```
Maxoudela: getHbar() and gVbar() should be made protected because anyone overriding the VirtualFlow will need to control them. I have several examples to support that:
- I have implemented a zoom property in the SpreadhsheetView that shrink or enhance the Grid. And as I am doing that, I still want the ScrollBar to have the same size (like Excel). Therefore, I must compensate the zoom factor applied on the grid, on the ScrollBar.
- I have added a Rectangle object on my Grid to provide a visual black Rectangle indicating the selected cells (just like Excel). This Rectangle is added on top of everything and I need to clip it if I don't want it to overlap the scrollBars. Therefore, I need to have access to the scrollBars and especially their width and height property in this case.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1894,8 +1917,11 @@ T getLastVisibleCellWithinViewPort() {
         return null;
     }
 
-    // Returns first visible cell whose bounds are entirely within the viewport
-    T getFirstVisibleCellWithinViewPort() {
+    /**
+     *
+     * @return first visible cell whose bounds are entirely within the viewport
+     */
+    protected T getFirstVisibleCellWithinViewPort() {
```
Maxoudela: These two methods getFirstVisibleCellWithinViewPort() and getLastVisibleCellWithinViewPort() are very useful when someone manually insert some rows into the VirtualFlow. One may needs to know which index are actually visible to the end user in the viewport.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1867,12 +1880,22 @@ private void resizeCellSize(T cell) {
...
 
-    private List<T> getCells() {
+    /**
+     * @return The list of cells representing those cells which actually make up the
+     * * current view. The cells are ordered such that the first cell in this
+     * * list is the first in the view, and the last cell is the last in the
+     * * view. When pixel scrolling, the list is simply shifted and items drop
+     * * off the beginning or the end, depending on the order of scrolling.
+     */
+    protected List<T> getCells() {
```
Maxoudela: The primary intent of overriding the VirtualFlow is to gain access to the rows displayed and tweak them. Therefore, this getter should be protected. 
For example, I use it to sort the rows that allow me to strictly control how the rows will be displayed, and which row will overlap another row when some cell are spanning.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -2110,22 +2136,34 @@ boolean addTrailingCells(boolean fillEmptyCells) {
...
+    protected void rebuildCells() {
         needsRebuildCells = true;
         requestLayout();
     }
 
-    void requestCellLayout() {
+    /**
+     * Inform the VirtualFlow that a layout should be done and only the cell layout will be triggered.
+     */
+    protected void requestCellLayout() {
```
Maxoudela: All the above methods should be made protected so that the developper can inform the VirtualFlow of some change, and trigger a layout. 
This is part of the basic toolkit when custom rows are inserted and may have some impact that the original VirtualFlow is not aware of.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
```diff
@@ -424,15 +448,15 @@ protected NestedTableColumnHeader createRootHeader() {
...
+    protected TableColumnHeader getReorderingRegion() {
         return reorderingRegion;
     }
 
     void setReorderingColumn(TableColumnBase rc) {
         dragHeaderLabel.setText(rc == null ? "" : rc.getText());
     }
 
-    void setReorderingRegion(TableColumnHeader reorderingRegion) {
+    protected void setReorderingRegion(TableColumnHeader reorderingRegion) {
```
Maxoudela: I am personally not using the reorderingRegion property. Only the reorderingProperty (see above) is useful to me when a reorder is taking place. I'll let @jperedadnr explain the needs.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
```diff
@@ -291,13 +291,28 @@ public TableHeaderRow(final TableViewSkinBase skin) {
...
+    public final boolean isReordering() {
         return reordering.get();
     }
-    final BooleanProperty reorderingProperty() {
+
+    /**
+     *
+     * @return A BooleanProperty indicating whether a reorder is going on.
+     */
+    public final BooleanProperty reorderingProperty() {
```
Maxoudela: If anyone has overriden the TableHeaderRow, he may wants to be notified when a reorder is taking place. He also may wants to programmatically trigger a reorder. These methods should therefore be made protected.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
```diff
@@ -455,7 +455,11 @@ void installTableColumnStyleClassListener() {
     NestedTableColumnHeader getNestedColumnHeader() { return nestedColumnHeader; }
     void setNestedColumnHeader(NestedTableColumnHeader nch) { nestedColumnHeader = nch; }
 
-    TableHeaderRow getTableHeaderRow() { return tableHeaderRow; }
+    /**
+     *
+     * @return The @link {@link TableHeaderRow} associated with this TableColumnHeader.
+     */
+    protected TableHeaderRow getTableHeaderRow() { return tableHeaderRow; }
```
Maxoudela: If anyone is tweaking the tableView header, it will tweak all the classes linked to it (TableHeaderRow created in TableViewSkinBase, NestedTableColumnHeader for rootHeader and NestedTableCOlumnHeader for actual column header).

If a custom TableHeaderRow has been provided in createTableHeaderRow() of TableViewSkinBase, it is only logical that its access should be protected in TableColumnHeader so that custom TableColumnHeader have access to the TableHeaderRow they're inside of.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
```diff
@@ -477,7 +481,11 @@ private void updateTableSkin() {
         }
     }
 
-    TableViewSkinBase<?,?,?,?,?> getTableSkin() {
+    /**
+     *
+     * @return The {@link TableViewSkinBase} in which this TableColumnHeader is inserted.
+     */
+    protected TableViewSkinBase<?,?,?,?,?> getTableSkin() {
```
Maxoudela: IMHO, this method is only used when interacted with TableSkinUtils. Since the TableSkin is created way before this class, it is not compulsory to open this method. But I believe that if the original TableCOlumnHeader is using it so much, anyone extending the TableColumnHeader will also need a convenient access to the TableSkin.

johanvos:
Is this now ready for review?

Maxoudela:
Yes, I have documented and explained every method change. Since 12 tags have been added.

Maxoudela:
Thank you very much for the review @nlisker !
Few things, when you suggest to use `@code`. Is it meant to replace the `{@link }` of simply the places where `{@link }` is not worth it and we simply want to highlight the word?

How would you change "This method is called when the width is changing."? By "Is  called when the width is changing."

I've quickly look and haven't found a method without description? Are you talking about the test package `VirtualFlowShim.java` ? There were no description before so I guessed it was ok to leave it like that.

I'll wait for your response and your thorough review before updating the PR. Thanks.

nlisker:
@Maxoudela My comments are following the [style guide](https://www.oracle.com/technetwork/articles/java/index-137868.html), which I recommend to take a look at (under the header "A Style Guide").

> Few things, when you suggest to use `@code`. Is it meant to replace the `{@link }` of simply the places where `{@link }` is not worth it and we simply want to highlight the word?

Yes. `@link` should be used when the extra information from the link is relevant, and not used excessively for the same type (usually once per doc page). So, add `@code` for all the places where there is none, and replace `@link` with it in all the places where there is no real need for `@link`.

> How would you change "This method is called when the width is changing."? By "Is called when the width is changing."

For the first sentence in the description I would write what the method does, not so much how it works. So for `updateTableWidth()` I would write something like (I didn't actually look to see what the method does): "Updates the table width when a resize operation occurs. The method is called when the table width is changing [is it being called continuously during changing, or once when the change starts?]. Overriding this method in a subclass allows to specify a special resizing behavior [do you have to call `super()`?]."

> I've quickly look and haven't found a method without description?

A few quick examples:

* TableColumnHeader.java: `getTableHeaderRow()`, `getTableSkin()`
* TableHeaderRow.java: `isReordering()`, `reorderingProperty()`, `getRootHeader()`. `getReorderingRegion()` is missing a `@return` description. `setReorderingRegion(TableColumnHeader reorderingRegion)` is missing `@param` description.

An API method *must* have at least: a description whose first sentence is its summary, a `@return` with description if the return type is not `void`, `@param`s with description for each argument if there are any. You can already go over the changed methods and fill in the holes.

eugener:
@kevinrushforth I never filled the CSR before. But can do it with your help.

Maxoudela:
@nlisker Thank you for the interesting link. I've tried to follow the guidelines and your input. I have made another commit that attempts to fix the issues. 
If you have some spare time this week-end to review a bit more, that would be much appreciated. 

I'll then send an email next week to openjfx-dev inviting comment on the API.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -2485,13 +2544,19 @@ private void releaseCell(T cell) {
...
 
+
     /**
-     * This method is an experts-only method - if the requested index is not
-     * already an existing visible cell, it will create a cell for the
-     * given index and insert it into the sheet. From that point on it will be
-     * unmanaged, and is up to the caller of this method to manage it.
+     * Creates and returns a new cell for the given index.
+     * <p>
+     * This method is an experts-only method - if the requested index is not already an existing visible cell, it will
```
nlisker: "This method is an experts-only method" Not sure how fitting this remark is. If it's supposed to be a warning, write the warning itself (that is, why is it "expert-only?). If there are pitfalls, explain them (and maybe how to deal with them).

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -2485,13 +2544,19 @@ private void releaseCell(T cell) {
...
+     * create a cell for the given index and insert it into the sheet. From that point on it will be unmanaged, and is
+     * up to the caller of this method to manage it.
+     *
+     * @param index the cell index
+     * @return A cell for the given index inserted in the VirtualFlow.
```
nlisker: * "A" -> "a"
* "VirtualFlow" -> "VirtualFlow container" (or similar)

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -2485,13 +2523,17 @@ private void releaseCell(T cell) {
...
     /**
      * This method is an experts-only method - if the requested index is not
      * already an existing visible cell, it will create a cell for the
      * given index and insert it into the sheet. From that point on it will be
      * unmanaged, and is up to the caller of this method to manage it.
+     *
+     * @param index
+     * @return
      */
-    T getPrivateCell(int index)  {
```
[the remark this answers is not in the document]
nlisker: A good example of usage.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
kleopatra: definitely needed in custom column headers, mainly to configure itself based on the table/skin (been there and suffered ;) There's a catch, though: in the "early" stages of the header's life-cycle - that is after being created and before the header row has been set - the skin is null. Plus there is no clean (nor any dirty) way to know when exactly the skin is available. Historically, its safe (read: not null) access got lost in a bug fix in NestedTableColumnHeader which also removed the skin as parameter in the constructor.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
jperedadnr: Sorry, I missed to reply to this comment.
About the region, it makes sense having direct access to it. For instance, for TableView2 we [override](https://bitbucket.org/controlsfx/controlsfx/src/09355e6b74836a8e2281fc4591e449ab36ce30fa/controlsfx/src/main/java/impl/org/controlsfx/tableview2/TableHeaderRow2.java?at=default#TableHeaderRow2.java-262) the setter to adjust the `dragHeader` region when reordering occurs.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1867,12 +1886,27 @@ private void resizeCellSize(T cell) {
         }
     }
 
-    private List<T> getCells() {
+    /**
+     * Returns the list of cells displayed in the current view.
+     * <p>
+     * The cells are ordered such that the first cell in this list is the first in the view, and the last cell is the
+     * last in the view. When pixel scrolling, the list is simply shifted and items drop off the beginning or the end,
```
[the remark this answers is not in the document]
nlisker: Could just be my lack of familiarity with this API. I was thinking that if I have a `TableView`, there are cells arranged in a 2D grid (rows and columns). If these are the cells that this methods returns in a (1D) list, how is the reduction happening? If the cells of `getCells` are rows (of cells) then it's fine, though I hope that this concept of virtual flow is known, otherwise calling "rows" "cells" is very confusing.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1735,10 +1735,27 @@ protected int getCellIndex(T cell){
...
+     */
+    protected final ScrollBar getHbar() {
         return hbar;
     }
-    final VirtualScrollBar getVbar() {
+
+    /**
+     * Returns the scroll bar used for scrolling vertically. A developer who needs to be notified when a scroll is
+     * happening could attach a listener to the {@link ScrollBar#valueProperty()}. The {@link ScrollBar#getWidth()} is
+     * also useful when adding a component over the {@code TableView} in order to clip it so that it don't overlap the
```
nlisker: "it don't" -> "it doesn't"

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1855,7 +1872,16 @@ private void positionCell(T cell, double position) {
         }
     }
 
-    private void resizeCellSize(T cell) {
+    /**
+     * Resizes the given cell. For example, if {@link #isVertical()} is set to {@code true}, the cell width will be
+     * maximum between the viewport width and the sum of all the cells {@code prefWidth}. The cell height will be
+     * computed by the cell itself unless {@code fixedCellSizeEnabled} is set to true, then {@link #getFixedCellSize()}
+     * is used.
+     *
```
nlisker: * "will be maximum" -> "will be the maximum"
* "the sum of all the cells" needs apostrophe: cells'
* "is set to true" needs `@code` for `true`
* "For example, if `{@link #isVertical()}` is set to `{@code true}`" if there are only 2 cases (vertical and horizontal), you can drop "for example" and add at the end that for `isVertical` set to `false` the width and height calculations are reversed.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1867,12 +1893,28 @@ private void resizeCellSize(T cell) {
...
+    protected List<T> getCells() {
         return cells;
     }
 
-    // Returns last visible cell whose bounds are entirely within the viewport
-    T getLastVisibleCellWithinViewPort() {
+    /**
+     * Returns the last visible cell whose bounds are entirely within the viewport. When manually inserting some rows,
+     * one may need to know which index are actually visible to the end user in the viewport.
+     *
```
nlisker: * "inserting some rows" remove "some"
* "which index are actually visible to the end user in the viewport" -> "which cell indices are visible in the viewport"

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1894,8 +1936,14 @@ T getLastVisibleCellWithinViewPort() {
         return null;
     }
 
-    // Returns first visible cell whose bounds are entirely within the viewport
-    T getFirstVisibleCellWithinViewPort() {
+    /**
+     * Returns the first visible cell whose bounds are entirely within the viewport. When manually inserting some rows,
+     * one may need to know which index are actually visible to the end user in the viewport.
+     *
```
nlisker: Same as above.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -2110,22 +2158,45 @@ boolean addTrailingCells(boolean fillEmptyCells) {
         return filledWithNonEmpty;
     }
 
-    void reconfigureCells() {
+    /**
+     * Informs the {@code VirtualFlow} that a layout pass should be done and cell contents (For example, in the {@code
+     * TableView}, if no cells has been added inside a row) are the same.
+     *
```
nlisker: * "the `{@code VirtualFlow}`" -> "the `{@code VirtualFlow}` container/region/handler/system...", it has to be something.
* "and cell contents" comma before "and" and "and that [the? or any cells?] cell contents"
* "For example" -> "for example"
* "cells has been" -> "cells have been"
* I still don't understand: cell contents are the same as what?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -2110,22 +2158,45 @@ boolean addTrailingCells(boolean fillEmptyCells) {
...
+    protected void reconfigureCells() {
         needsReconfigureCells = true;
         requestLayout();
     }
 
-    void recreateCells() {
+    /**
+     * Informs the {@code VirtualFlow} that a layout pass should be done and cell factory has changed. All cells are
+     * trashed and recreated to be replaced in the viewport.
+     *
```
nlisker: Same as above.

* comma before "and"
* "and cell factory" -> "and that a/the cell factory"
* What does "trashed" mean? Removed, deleted, dereferenced...?
* "recreated to be replaced" is odd. Maybe "recreated with the intent to be replaced"? I'm not sure what you're trying to say here.
* If we're talking about the cells in the viewport, then rearrange the position of "in the viewport" to "All cells in the viewport are trashed and recreated to be replaced".

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -2110,22 +2158,45 @@ boolean addTrailingCells(boolean fillEmptyCells) {
...
+    protected void recreateCells() {
         needsRecreateCells = true;
         requestLayout();
     }
 
-    void rebuildCells() {
+    /**
+     * Informs the {@code VirtualFlow} that a layout pass should be done and cell contents have changed. All cells are
+     * removed then added properly in the viewport.
+     *
```
nlisker: Same as above.

* "removed then" -> "removed and then"

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -2110,22 +2158,45 @@ boolean addTrailingCells(boolean fillEmptyCells) {
...
+     */
+    protected void rebuildCells() {
         needsRebuildCells = true;
         requestLayout();
     }
 
-    void requestCellLayout() {
+    /**
+     * Informs the {@code VirtualFlow} that a layout pass should be done and only the cell layout will be requested.
+     *
```
nlisker: Same as above.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -2485,13 +2556,22 @@ private void releaseCell(T cell) {
...
-     * This method is an experts-only method - if the requested index is not
-     * already an existing visible cell, it will create a cell for the
-     * given index and insert it into the sheet. From that point on it will be
-     * unmanaged, and is up to the caller of this method to manage it.
+     * Creates and returns a new cell for the given index.
+     * <p>
+     * If the requested index is not already an existing visible cell, it will create a cell for the given index and
+     * insert it into the sheet. If the index is existing, simply returns the visible cell. From that point on it will
+     * be unmanaged, and is up to the caller of this method to manage it.
+     * <p>
```
nlisker: * "sheet" is the table? if so, use that instead of a new term, otherwise explain what "sheet" is.
* "index is existing" -> "index exists"
* "From that point on" comma after "on"

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -2485,13 +2556,22 @@ private void releaseCell(T cell) {
...
+     * This is useful if you need a row that should not be visible (a row that always stick to the top for example). It
+     * allows you to easily creates it and be sure it's correctly initialized and inserted in the {@code VirtualFlow}.
+     *
```
nlisker: Avoid "you": "useful in the case of a row", or "useful if one needs a row", or "useful if a row... is desired" or something similar.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
```diff
@@ -378,9 +404,17 @@ private final void setRootHeader(NestedTableColumnHeader value) {
...
-    // but made private again for JDK 9
-    void updateScrollX() {
+    /**
+     * Called whenever the value of the horizontal scrollbar changes in order to request layout changes - shifting the
+     * {@code TableColumnHeaders}.
+     * <p>
+     * For example, if you have added custom components around a {@code TableColumnHeader} (such as icons above), you
+     * will also want to shift them. When overriding, calling super() is required to shift the {@code
+     * TableColumnHeaders} and it's up to the developer to notify its own custom components.
+     *
```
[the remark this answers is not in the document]
Maxoudela: Regarding your last bullet, I don't really understand. We stated that this method was called whenever the value of the horizontal scrollbar changes. 
So "it's up to the developer to notify its own custom components" should be sufficient in that case no? We can repeat that "the developer can notify its own custom components that a change to the horizontal scroll value has been changed and the location of the column headers have shifted". But IMHO this is stated right above.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
```diff
@@ -573,48 +579,96 @@ void horizontalScroll() {
         tableHeaderRow.updateScrollX();
     }
 
-    void onFocusPreviousCell() {
-        TableFocusModel<M,?> fm = getFocusModel();
+    /**
+     * Called when the focus is set on previous cell in order to scroll to it to make it visible. This usually happens
+     * when the focus is set on the cell above the current focused cell.
+     *
```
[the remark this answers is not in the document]
Maxoudela: I said "on the cell above the current focused cell". I really don't see how you managed to see it on the left? 
We are dealing with the TableView that lays its cells into rows. There is no orientation that can be chosen on the TableView regarding the VirtualFlow. The VirtualFlow orientation will always be vertical meaning the Virtualization will be on rows.

I believe above is a pretty straight forward when you consider the TableView layout :
cell11 cell12 cell13
cell21 cell22 cell23
cell31 cell32 cell33

If I'm considering cell22, above is refering to cell12. I don't see how it could be mistaken by cell21. Or if it is, I'm lacking words to describe it more accurately.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
Maxoudela: I don't really know what the previous developer intended when he used the "cell contents are the same". My deduction base on other methods is that the number of cells inside a row has not changed. 
Don't forget that the VirtualFlow applied to the TableView means that the "cells" we are refering in the VirtualFlow are in fact "rows". So I guess that "cell content are the same" means the numbers of cells inside the row has not changed and we do not need to remove them and re-insert them. But again, I only seen it called at one place and it was pretty confusing.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
Maxoudela: ""recreated to be replaced" : The cell factory (aka the rowFactory for the TableView) has changed so every rows and cells are incorrect. So we must delete them all, and call the factory again to re-create new rows filled with new cells, then insert them in the viewport.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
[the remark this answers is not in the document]
nlisker: > So "it's up to the developer to notify its own custom components" should be sufficient in that case no?

Could be, but I think it's clearer with something as simple as "notify its own custom components of this change".

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
nlisker: Maybe the confusing part is "This *usually* happens when", it implies that there are cases where the cell is not the one above.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
nlisker: Then we require clarification from someone who is familiar with the implementation. @kevinrushforth, can you help clarify?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
nlisker: OK, so I would write something like:
"... and that the cell factory has changed. All cells in the viewport are recreated with the new cell factory."

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
Maxoudela: In fact, I'm not really sure about the documentation. The TableView has a Behavior that will execute the selection and the focus. All these methods are called like that :
```
behavior.setOnFocusPreviousRow(() -> onFocusPreviousCell());
        behavior.setOnFocusNextRow(() -> onFocusNextCell());
        behavior.setOnMoveToFirstCell(() -> onMoveToFirstCell());
        behavior.setOnMoveToLastCell(() -> onMoveToLastCell());
        behavior.setOnScrollPageDown(isFocusDriven -> onScrollPageDown(isFocusDriven));
        behavior.setOnScrollPageUp(isFocusDriven -> onScrollPageUp(isFocusDriven));
        behavior.setOnSelectPreviousRow(() -> onSelectPreviousCell());
        behavior.setOnSelectNextRow(() -> onSelectNextCell());
        behavior.setOnSelectLeftCell(() -> onSelectLeftCell());
        behavior.setOnSelectRightCell(() -> onSelectRightCell());
```

So basically, the onFocusNextCell and onFocusPreviousCell are doing the exact same thing. They scroll to make the cell visible. I used the word "usually" because I do not like the method naming. I believe there should only be one method called "scrollToFocusedCell". And then the description would be easier. 
Right now we have a method "onFocusNextCell" that is called when the focus is changed, but anyone could call it to simply scroll to the focused index..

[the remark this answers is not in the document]
nlisker: So I suggest you discuss this in a comment here or in the mailing list. Maybe `onFocusPreviousCell` and `onFocusNextCell` can be combined to one. `onSelectPreviousCell` and `onSelectNextCell` are a similar case.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
Maxoudela: All of these methods are doing a layout, with some little differences in order to optimise the layout I guess.
Regarding reconfigureCells(), it's only called by the TableViewSkin when its own `needCellsReconfigured` variable is set to true.

This variable is set to true at two places :
```
if (newCount != oldCount) {
            // FIXME updateItemCount is called _a lot_. Perhaps we can make rebuildCells
            // smarter. Imagine if items has one million items added - do we really
            // need to rebuildCells a million times? Maybe this is better now that
            // we do rebuildCells instead of recreateCells.
            requestRebuildCells();
        } else {
            needCellsReconfigured = true;
        }
```
That made me add the comment about the number of cells inside a row.

And 

```
 private InvalidationListener widthListener = observable -> {
        // This forces the horizontal scrollbar to show when the column
        // resizing occurs. It is not ideal, but will work for now.

        // using 'needCellsReconfigured' here rather than 'needCellsRebuilt'
        // as otherwise performance suffers massively (RT-27831)
        needCellsReconfigured = true;
        if (getSkinnable() != null) {
            getSkinnable().requestLayout();
        }
    };
```

If we look inside the VirtualFlow at the impact of this variable :
```
setMaxPrefBreadth(-1);
            lastWidth = -1;
            lastHeight = -1;
```

My guess is that this method is requesting a layout when the Table width has changed and that nothing has changed internally inside the rows. But I'm no expert on the inner behaviour of the VirtualFlow and the subtleties between all these methods.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
Maxoudela: If we decide to merge them as one, we loose the information whether the method was called because the TableViewBehavior has selected the previous or the next cell. 
In my personal case, I don't care since I simply scroll to make the cell visible. So does the TableView. Maybe people are interested in that information?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
```diff
@@ -638,10 +692,17 @@ Region getColumnReorderLine() {
...
 
     /**
-     * Function used to scroll the container down by one 'page', although
-     * if this is a horizontal container, then the scrolling will be to the right.
+     * Scrolls the container down by one 'page'. The direction is determined by the orientation ({@link
+     * VirtualFlow#isVertical()}) of the {@code VirtualFlow} (down for vertical, right for horizontal).
+     * <p>
+     * If the last visible cell is selected (or f{@code isFocusDriven} is {@code true}), it will be made visible at the
+     * very top.
+     *
```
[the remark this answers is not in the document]
nlisker: Then ask Kevin or Ajit for an explanation.

If your assumption is true, then something like:
"Scrolls the container [maybe "content" is better?] down so that the last cell becomes the first cell" should be good.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
nlisker: Maybe, this is something to discuss with everyone. In any case, I think that "previous" and "next" are not the best descriptors. "up"/"down" or "above"/"below" are more concrete in my opinion.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
Maxoudela: Ok I've read this "Move the selection down on the last fully visible item on the current page. If the last fully visible item is selected, move the selection on the last fully visible item on the next page." from https://wiki.openjdk.java.net/display/OpenJFX/TableView+User+Experience+Documentation

And that concurs what I've deduced from the code. This seems quite odds but the specification are what they are. 
So "Scrolls the content down so that the last cell becomes the first cell. If the last visible cell is selected, scrolls the content down so that the last cell on the next page because visible"? That's rather odd because we're using the definition of a page inside the definition of a page...

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
Maxoudela: They are called by the behavior when "setOnSelectPreviousRow" is called. "Previous" row makes sense. But the "previous cell" does not really indeed. Especially that the behavior methods are mapped to the "up" and "down" keyboard buttons. 
What's the best way to collect opinions? The mailing list or here?

kevinrushforth:
Removing the [WIP] label since this is now under active review. I won't have time to look at this week, but it seems to be getting close to being ready for the next step. As I mentioned above, once the API is generally agreed upon, you can file a CSR.

@Maxoudela : I asked above whether @eugener might be able to do that, but I forgot that you also have JBS access, so you can file it yourself.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
nlisker: Probably the mailing list.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
nlisker: I don't see the definition of a page there, but it really seems like a "page" is the cell/row content of the viewport. Can you test empirically with a `TableView` how it behaves?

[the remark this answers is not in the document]
Maxoudela: Allright I've been able to test on a TableView. When pressing "page down", the selection is set on the last fully visible cell/row. So a page is definitely the content of the viewport. Then if I press again "page down", it will scrolll so that the last visible cell that was selected is displayed on top.
In the description, I would empathized the fact that this method is called when the selection is changing.
Maybe something like that :
"Returns the index to select after a page scroll.
The last fully visible cell index on the current viewport is always returned. If this index is already selected, a scroll is happening first, so that it is displayed at the top of the viewport. Then the last fully visible cell index of the viewport is returned.
If {@code isFocusDriven} is set to {@code true}, the focused cell is considered instead of the selected cell."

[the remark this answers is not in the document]
nlisker: Alright, so that's confirmed.

I would rephrase starting from the 2nd sentence to:
"The returned index belongs to the last fully visible cell in the viewport, after scrolling (if needed). Scrolling is done only if that cell is already selected, in which case that cell is made visible at the top of the viewport.
If `{@code isFocusDriven}` is ["set to" is optional] `{@code true}`, the focused cell is used instead of the selected cell."

Does the last sentence apply only to the scrolling, or to the returned result also? That is, if it's `true`, is the first sentence effectively "Returns the index to focus after a page scroll."?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
kleopatra: wrote up [some thoughts](https://github.com/kleopatra/swingempire-fx/wiki/TableSelectionModel) on naming. The executive version:

- let the naming be driven by the naming of the corresponding selection/focusModel methods
- use consistent names across skin/behavior
- be complete, that is let behavior/skin provide api hooks for a action/hook methods for all model navigational methods that are used by them

This would be partly out off scope of the core bug, as it would require changes in behavior as well (change xxPreviousRow to xxAboveCell and add additional methods for taking callback hooks). Could mitigate by defering the completeness to future and align the skin names to the corresponding behavior names - they are acceptable, IMO.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
Maxoudela: I just checked in the code and it's the case. If `isFocusDriven` is `true`, the index will be focused after a page scroll, not selected.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
Maxoudela: Thanks for the feedback, few thoughts:
- The TableView has two selection Mode : Cell based or row based. Maybe the confusion is the TableViewBehavior comes from that. `SelectLeftCell ` is acceptable because it will only be called when the TableView is in cell based selection. 
However, selecting the "previous" is complicate to name. If you are row based, you'll select the previous row (and that make sense because in row based selection, you only have the next row and the previous row). But if you are cell based, you'll select the cell above. So the naming of the method that will be called when the key "UP" is pressed is not trivial.

Another thing I wanted to share is also the fact that there is no "onFocusLeftCell" in the TableSkin. Right now we have in the Behavior :
```
onSelectNextRow
onSelectPreviousRow
onSelectRightCell
onSelectLeftCell
```
But we only have :
```
onFocusPreviousRow
onFocusNextRow
```
And if we look at the TableViewBehaviorBase :
```
protected void focusLeftCell() {
        TableSelectionModel sm = getSelectionModel();
        if (sm == null) return;
        TableFocusModel fm = getFocusModel();
        if (fm == null) return;

        fm.focusLeftCell();

        if (onFocusPreviousRow != null) onFocusPreviousRow.run();
    }
```
So we are calling `onFocusPreviousRow ` when we are focusing the left cell. So I don't know what the final decision will be, but if we want to tackle these naming issue, we may want to address what I've stated above also.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
nlisker: This method is starting to be cumbersome because it can either focus or select, and it can either scroll the content or not (even though it's ca;;ed "onScroll..."). It seems to me like it could be 2 methods and renamed to "selectOnPageUp" and "focusOnPageUp".

In any case, maybe:

> Returns the index of the selected (or focused, if `{@code isFocusDriven}` is `{@true}`) cell after a page scroll operation [added "operation" because actual scrolling doesn't always happen]. If the selected/focused cell is not the last fully visible cell, then the last fully visible cell is selected/focused. Otherwise, the content is scrolled such that the cell is made visible at the top of the viewport (and the new last fully visible cell is selected/focused instead).

I'm afraid to ask what happens if the selected/focused cell is the "real" last cell - the one not fully visible. Does it not behave like the last fully visible cell?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
nlisker: @kleopatra After trying to help with the documentation I'm realizing fast that the API is not standard. Maybe you could help with creating a normal API with the people who are familiar with it.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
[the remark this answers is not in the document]
nlisker: The only good thing about this behavior is that the docs are still accurate...

Honestly, this behavior sounds like a bug to me.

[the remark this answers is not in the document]
Maxoudela: I totally agree with you but it seems like changing some "package-protected" methods to "protected" is a huge deal. I can't even imagine what it will be if I suggest changing the internal code and behavior of the method to be honest.
I think we should document it like we did, and maybe open a bug/improvement ticket that can be resolved later or in a different PR.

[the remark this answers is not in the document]
kleopatra: to put coal into the fire: there's an [open issue concerning scrolling behavior](https://bugs.openjdk.java.net/browse/JDK-8197536) on list/table ;)

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
kleopatra: Not sure where nor how to put this ... but I think there are a couple of more methods that should be exposed/added:

expose:

    /**
     * Called by the parent header when this {@code TableColumnHeader} is removed
     * from its children. This method
     * must implement any logic necessary to clean up itself. 
     */
    protected void dispose() {

Reason: Implementations of custom column headers usually wire custom listeners (or other resources) that require to add custom cleanup. Currently, there's no way to do so. If done, it must be done also in NestedTableColumnHeader.

add: access to changeListenerHandler (could be implemented just the same way as in SkinBase without exposing the internal LambdaMultiplePropertyChangeListenerHandler)

reason: custom header typically add additional changeListeners - doing so by api that's already available would cleanup those wiring automatically in super's dispose. Currently, that can only be done by either re-implementing something like the lambdaHandler (*cough by c&p) or reflectively accessing the internal field (*cough again).

kevinrushforth:
Regarding the [CSR](https://bugs.openjdk.java.net/browse/JDK-8212584), I left a comment that the specification needs to be inline and not a pointer to somewhere else. The Specification section in [JDK-8203409](https://bugs.openjdk.java.net/browse/JDK-8203409) has a good example of this.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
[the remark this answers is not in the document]
kevinrushforth: Agreed. This was one of the methods that worried me, and this description makes it clear.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1855,7 +1872,16 @@ private void positionCell(T cell, double position) {
...
+    /**
+     * Resizes the given cell. If {@link #isVertical()} is set to {@code true}, the cell width will be the maximum
+     * between the viewport width and the sum of all the cells' {@code prefWidth}. The cell height will be computed by
+     * the cell itself unless {@code fixedCellSizeEnabled} is set to {@code true}, then {@link #getFixedCellSize()} is
+     * used. If {@link #isVertical()} is set to {@code false}, the width and height calculations are reversed.
+     *
+     * @param cell the cell to resize
+     * @since 12
+     */
+    protected void resizeCellSize(T cell) {
```
kevinrushforth: I think `resizeCell` would be a better name for this method. The `Size` suffix seems redundant.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1867,12 +1893,28 @@ private void resizeCellSize(T cell) {
...
-    // Returns last visible cell whose bounds are entirely within the viewport
-    T getLastVisibleCellWithinViewPort() {
+    /**
+     * Returns the last visible cell whose bounds are entirely within the viewport. When manually inserting rows, one
+     * may need to know which cell indices are visible in the viewport.
+     *
+     * @return last visible cell whose bounds are entirely within the viewport
+     * @since 12
+     */
+    protected T getLastVisibleCellWithinViewPort() {
```
kevinrushforth: 'viewport' is one word, so this should be renamed to `getLastVisibleCellWithinViewport`. Ditto for getFirst...

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
```diff
@@ -573,48 +579,96 @@ void horizontalScroll() {
...
+    protected void onFocusNextCell() {
+        TableFocusModel<M, ?> fm = getFocusModel();
         if (fm == null) return;
 
         flow.scrollTo(fm.getFocusedIndex());
     }
 
-    void onSelectPreviousCell() {
+    /**
+     * Called when the selection is set on the previous cell in order to scroll to it to make it visible. This      *
```
kevinrushforth: You have a spurious `*` on this line which shows up in the generated javadocs.

Maxoudela:
Thank you @kevinrushforth for the review, I'll modify right away the documentation.

Regarding your comment :

> Can you provide an example of VirtualFlow subclass showing how the new public and protected methods will be used? This will help show that the the new methods are sufficient for an application to be able to subclass VirtualFlow. This could either go in as a new simple application in apps/toys/Hello or in tests/manual.

How could I proceed? I am overriding a little but the VirtualFlow in my application for very specific needs (inserting new rows, sorting them, adding listeners to the ScrollBars). I'm afraid my use case is very dependant on my own custom TableView. But I can provide a basic VirtualFlow Subclass that demonstrates that we can force the layout or change the scrollBars value for example. Would that be sufficient?

kevinrushforth:
> I can provide a basic VirtualFlow Subclass that demonstrates that we can force the layout or change the scrollBars value for example. Would that be sufficient?

It might be. I'd want to see the example and how much of the new API it uses to say for sure.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
```diff
@@ -508,7 +508,11 @@ protected TableHeaderRow createTableHeaderRow() {
      *                                                                         *
      **************************************************************************/
 
-    final TableHeaderRow getTableHeaderRow() {
+    /**
+     *
+     * @return The TableHeaderRow created using {@link #createTableHeaderRow()}.
+     */
+    protected final TableHeaderRow getTableHeaderRow() {
```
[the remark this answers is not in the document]
kevinrushforth: If there is a use case to override then it could be made non-final. Is the "override to narrow the type" something that you see as being likely?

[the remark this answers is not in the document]
kleopatra: yes, I do it whenever the custom class created by the factory method has additional public api (vs. just a bug hacking class). My pattern is something like:

// factory method
protected MyType createType() {
    return new MyType();
}

protected MyType getType() {
    // cast is safe because the factory guarantees at least MyType
    return (MyType) super.getType();
}   

That was fine for fx8 (f.i.) TableHeaderRow TableViewSkin, not possible since fx9. The ugly way around a final getter is to either cast at each usage of any extended api or introduce a custom getter with another name, like

protected MyType getMyType() {
    return (MyType) getType();
}

Note that the pattern of overidable factory methods and final getters is spread all over the skin classes, very nasty, IMO.

kleopatra:
> 
> 
> > I can provide a basic VirtualFlow Subclass that demonstrates that we can force the layout or change the scrollBars value for example. Would that be sufficient?
> 
> It might be. I'd want to see the example and how much of the new API it uses to say for sure.

A good measure of which of the api is needed, are the internal classes that use a VirtualFlow: all what they need (at least the  intersection of all of them), will very likely be needed by custom implementations as well ;)

Maxoudela:
I have removed the `final` modifier for `getTableHeaderRow`. 

I have modified the Javadoc according to the review. I have added two unit test for `getFirstVisibleCellWithinViewport `and `getLastVisibleCellWithinViewport `. Regarding the rest, I have no clue how to uni tests them.

Also regarding an example of a VirtualFlow subclass, I have been thinking for days and I don't quite see how I could produce something meaningful. The only thing that is easily demonstrable is to modify the value of the scrollBar from the class itself, or add some listeners on it. 

But I don't see how I can provide a VirtualFlow subclass and assert that all the methods we exposed are sufficient for subclassing.

kevinrushforth:
The following comment by @kleopatra might be good enough justification for exposing a method:

> A good measure of which of the api is needed, are the internal classes that use a VirtualFlow (or any of the other classes that are considered for opening right now)

This is a good way of looking at it: what methods of VirtualFlow would need to be exposed in order to implement, e.g., a custom replacement for TableView (or pick your other favorite, but that's probably the best example)?

So for the example control, it might be enough to have one that demonstrates something that can now be done using the new protected / public API that can't be done today without dipping into non-exported API using qualified exports.

@eugener do you have any good examples of third-party controls (e.g., from ControlsFX) that need this?

Maxoudela:
Exactly, these two components of `ControlsFX `are not working with OpenJFX 11 right now. And I used them as reference for determining which methods should be opened.

The only thing I worry is that I may have forgotten some methods that would come handy in the future. But some reviewers have given their opinion and asked for more methods that I originally planned so I guess that what we have right now should be sufficient. 

Do you want me to publish a customised version of SpreadsheetView or TableView2 using this version of OpenJFX to demonstrate that the version can be used to compile and run these controls?

kevinrushforth:
In JBS, @kleopatra raised the issue of the naming of the `on{Focus,Select}*Cell` methods of `TableViewSkin`. I wrote up [3 possible directions](https://bugs.openjdk.java.net/browse/JDK-8207942?focusedCommentId=14221550&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14221550) and kleopatra replied with [her recommendation](https://bugs.openjdk.java.net/browse/JDK-8207942?focusedCommentId=14221715&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14221715).

I agree with her recommendation, so unless there are other arguments against doing this, I would like to see this change implemented:

* Rename the existing `on{Focus,Select}{Previous,Next}Cell` methods to `on{Focus,Select}{Above,Below}Cell` 
* Add the missing `onFocus{Left,Right}Cell` methods

kevinrushforth:
Btw, I will do a final review of the API and review the CSR after this last change is resolved. Separately, I will review the recent impl changes and new unit test (but there may be a slight delay in that).

aghaisas:
> 
> The only thing I worry is that I may have forgotten some methods that would come handy in the future. But some reviewers have given their opinion and asked for more methods that I originally planned so I guess that what we have right now should be sufficient.

I got some time to see the conversations and changes today. This is a great collaboration in progress.
It is OK if we discover few more methods that need to be opened up for overriding in future. As of now, let us address what has been identified.

jperedadnr:
I've built JavaFX with the latest changes from this PR and tested an stand-alone version of TableView2 (making the necessary changes to run on Java 11), and it works mostly fine.

I had to get rid of behavior and selection model (expected), but also from `TableViewSkin::resizeColumnToFitContent`, `TableViewSkin::resizeColumn` (both now moved to the private `TableSkinUtils` class, and probably related to this [issue](https://bugs.openjdk.java.net/browse/JDK-8207957)), and `TableViewSkin::getColumnReorderLine`, `TableViewSkin::horizontalScroll` (I could workaround these two though).

Maxoudela:
Thanks all for your thoughts.

Do you have an opinion @kleopatra  on @nlisker ideas? I have no clue on which API is better, but the idea of having one `onSelectMethod(Side/Direction direction)` is appealing.

Whatever version we choose to end up, are we clear that these modifications will impact `TableViewBehaviorBase.java` right?

If I add the `onFocus{Left,Right}Cell` on `TableViewSkinBase`, it seems logical to create the `setOnFocus{Left,Right}Cell` methods in `TableViewBehaviorBase` right? As a reminder, the `focusLeftCell()` of `TableViewBehaviorBase` is calling `onFocusPreviousRow` right now : 
```
protected void focusLeftCell() {
        TableSelectionModel sm = getSelectionModel();
        if (sm == null) return;

        TableFocusModel fm = getFocusModel();
        if (fm == null) return;

        fm.focusLeftCell();
        if (onFocusPreviousRow != null) onFocusPreviousRow.run();
    }
```

kleopatra:
Hi all,

> Do you have an opinion @kleopatra  on @nlisker ideas? I have no clue  
> on which API is better, but the idea of having one  
> `onSelectMethod(Side/Direction direction)` is appealing.

honestly, I don't see how to introduce inconsistent (with models) api  
can be appealing <g>. All it does is to force switch statements into  
the implementations which in turn might delegate to private api to do  
the special task.

So - I'm against those suggestions (didn't go into the xxPage or  
xxScroll, though): stick as near as possible to the model api, that'll  
make it self-explanatory per consistency - it's always the same method  
to use, nothing new to remember (important for my oldish brain :)

>
> Whatever version we choose to end up, are we clear that these  
> modifications will impact `TableViewBehaviorBase.java` right?
>
> If I add the `onFocus{Left,Right}Cell` on `TableViewSkinBase`, it  
> seems logical to create the `setOnFocus{Left,Right}Cell` methods in  
> `TableViewBehaviorBase` right?

exactly: behaviour needs the corresponding setOnXX as well and must  
use whichever is appropriate.


Cheers, Jeanette

Maxoudela:
Allright the modifications have been made.

Should I add the 

```
behavior.setOnFocusLeftCell(() -> onFocusLeftCell());
        behavior.setOnFocusRightCell(() -> onFocusRightCell());
```

code modification into the CSR @kevinrushforth ?

kevinrushforth:
New public methods look good. 

The only question is whether we should also rename the existing behavior methods to match, as @kleopatra suggested, but I'm OK with leaving that for another day.

I have a minor javadoc comment (I'll add it below).

kevinrushforth:
Hmm. The above comments didn't show up with the code. hopefully you can figure out which methods I was referring to by clicking on the commit (sorry about that).

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/VirtualFlowSubClassTest.java
```diff
@@ -0,0 +1,153 @@
+/*
+ * Copyright (c) 2010, 2016, Oracle and/or its affiliates. All rights reserved.
```
kevinrushforth: Minor nit: Since this is a new file, it should be just `Copyright (c) 2018, Oracle ...`

kevinrushforth:
I think the CSR can be updated now (after you fix the two javadoc typos).

kleopatra:
A: follow-up on [my comment](https://github.com/javafxports/openjdk-jfx/pull/163#discussion_r232225424) - that is, with the pattern factory methods/getter:
  
    protected SomeType createSomeType();
    protected SomeType getSomeType();

the getter should not be final to allow covariant overrides of both.

This is changed for TableViewSkinBase.create/getTableHeaderRow. Another occurence of the same pattern which would benefit from a similar change:

VirtualContainerBase.create/getVirtualFlow

There's also TableHeaderRow.create/getRootHeader - probably not changeable because it's the value of a property and making the getter not-final would break the "all-final" rule for properties?

B: follow-up of ?? can't find it right now - was a beware for TableColumnHeader.getTableSkin: the returned skin is null until the columnHeader's setRootHeader(..) is called - which happens at an unspecified time _after_ the constructor returns. That's not entirely obvious from the api so should be documented, at least mentioning the null. Something along the lines:

Returns... [ as is], may be null.
Note: the skin is accessible only after this columnHeader's TableHeaderRow has been set. In 
particular, that's _not_ the case while being in the constructor. There is no specified time at which the skin can be accessed.

Native speakers, please improve :)

Maxoudela:
@kevinrushforth  Actually, I don't see where you find the typo :
capture

kevinrushforth:
> @kevinrushforth Actually, I don't see where you find the typo :

I no longer see this typo, so it may have been something odd with my browser (e.g., maybe a couple words were truncated and I didn't notice). In any case, I see no problems with this.

kevinrushforth:
> A: follow-up on ... the pattern factory-method/getter:
> 
> ```
> protected SomeType createSomeType();
> protected SomeType getSomeType();
> ```
> 
> the getter should not be final to allow covariant overrides of both.
> 
> ... Another occurence of the same pattern which would benefit from a similar change:
> 
> ```
> VirtualContainerBase.create/getVirtualFlow
> ```

This enhancement doesn't otherwise modify VirtualContainerBase, so let's leave this for a follow-on RFE if it is needed / useful. I think we have a good set of modifications that are now self-consistent, so I'd prefer to finish up what we have, so we can get it into openjfx12 soon.

> There's also TableHeaderRow.create/getRootHeader - probably not changeable because it's the value of a property and making the getter not-final would break the "all-final" rule for properties?

Right, since `rootHeader` is a Property, it needs to remain final.

> B: follow-up of ?? can't find it right now - was a beware for TableColumnHeader.getTableSkin: the returned skin is null until the columnHeader's setRootHeader(..) is called - which happens at an unspecified time _after_ the constructor returns. That's not entirely obvious from the api so should be documented, at least mentioning the null. Something along the lines:
> 
> ```
> Returns... [ as is], may be null. <p>
> Note: the return value is null during instantiation. There is no specification as to when
> the skin is accessible, though seems to be safe after this columnHeader's 
> TableHeaderRow has been set. 
> ```
> 
> Native speakers, please improve :)

I like documenting that it might be null. I don't want to put language in the spec that effectively says "not sure, but it ought to be safe enough to assume...". I propose something like this:

```
     * Returns the {@code TableViewSkinBase} in which this {@code TableColumnHeader} is inserted.
     * This will return {@code null} until the {@code TableHeaderRow} has been set.
     *
     * @return the {@code TableViewSkinBase} in which this {@code TableColumnHeader} is inserted,
     * or {@code null}
```

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
```diff
@@ -314,9 +327,22 @@ final BooleanProperty reorderingProperty() {
     private final ReadOnlyObjectProperty<NestedTableColumnHeader> rootHeaderProperty() {
         return rootHeader.getReadOnlyProperty();
     }
-    final NestedTableColumnHeader getRootHeader() {
+
```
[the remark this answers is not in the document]
Maxoudela: IMHO, the `getRootHeader` is sufficient and I don't see a need to observe the change. I suggest we leave it as it for now.

Maxoudela:
@kevinrushforth  I have made the modifications and updated the CSR. Thanks

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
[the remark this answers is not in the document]
kevinrushforth: Sounds good. I'll review the updated version then.

kevinrushforth:
> I have made the modifications and updated the CSR.

I'll review it next week.

Maxoudela:
Thanks for the CSR, I have moved it to the Finalized state. 

I've created https://github.com/javafxports/openjdk-jfx/issues/324 regarding the test part.

PR #165, opened 2018-08-23
Use latest OpenJDK 11 EA build as boot JDK

kevinrushforth:
Modifies the Travis and Appveyor scripts to download and use the latest OpenJDK 11 EA build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 sudo: false
 
-jdk: openjdk10
+jdk: openjdk11
 
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


kevinrushforth:
If we want to get this in prior to the GA of OpenJDK 11, then all that is needed is to change the build logic in `appveyor.yml` to grab the latest EA build from the http://jdk.java.net/11/ page rather than hard-coding build 27 (also, I will revert my temporary change that bumped the minimum boot JDK).

brcolow:
LGTM.

PR #158, opened 2018-08-21
Change .ci/* scripts to be non-executable

kevinrushforth:
fixes #157

PR #166, opened 2018-08-23
Integrate recent develop commits into master

kevinrushforth:
Three commits that touched github-only files were recently merged into the develop branch. The pull requests, in order from oldest to newest, are:

#154 - Check white-space rules as part of Travis CI build
#158 - Change .ci/* scripts to be non-executable
#153 - Add a powershell script for Windows dev, use on Appveyor.

These commits need to be integrated into the master branch.

NOTE: unlike normal pull requests, I propose to keep these as three separate commits so that the history is preserved. This means that, once approved, I will "Rebase and merge" rather than "squash and merge" as is typical.


kevinrushforth:
@brcolow you might double-check these, too.

kevinrushforth:
Perhaps not, but if some other script wanted to use pr-list-files.sh it might be best to keep it.

In any case, I note that changing the contents of any of these files as part of this PR will almost certainly cause a merge conflict when merging master --> develop, so it would be best to address any changes in a subsequent PR if needed. The intent wasn't really to review the changes again (unless integrating them would cause a build failure or similar), but rather review that the three commits in question will bring the master branch in line with develop.

PR #168, opened 2018-08-25
[Appveyor] Fail the build if gradle fails.

brcolow:
I am not positive, but I think after #153 was merged the Appveyor build no longer fails if a test fails. I only confirmed this on my fork with full system tests enabled, and even when tests fail, the build passes. I believe this is because the powershell script executes successfully, and returns a 0 exit code, even though the last exit code from invoking gradle was non-zero. This PR should fix that.


brcolow:
Opened issue: https://bugs.openjdk.java.net/browse/JDK-8209970 and sent RFR.

PR #169, opened 2018-08-25
JDK-8209967: Bump minimum gradle version to 4.8 for JDK 11

kevinrushforth:
Fixes [JDK-8209966](https://bugs.openjdk.java.net/browse/JDK-8209966).

This bumps the minimum gradle version to 4.8, which is the currently supported version (and is already the version used when running gradlew).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1212,14 +1212,14 @@ if (Integer.parseInt(NUM_COMPILE_THREADS.toString()) < 1) {
     NUM_COMPILE_THREADS = 1
 }
 
-// Check for Gradle 4.8, error if < 4.3.
+// Check for Gradle 4.8, error if < 4.8.
 if (gradle.gradleVersion != "4.8") {
     def ver = gradle.gradleVersion.split("[\\.]");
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -1212,14 +1212,14 @@ if (Integer.parseInt(NUM_COMPILE_THREADS.toString()) < 1) {
     NUM_COMPILE_THREADS = 1
 }
 
-// Check for Gradle 4.8, error if < 4.3.
+// Check for Gradle 4.8, error if < 4.8.
 if (gradle.gradleVersion != "4.8") {
```
nlisker: Would it be easier to have a variable for the minimum gradle version and use that everywhere instead of a "magic number" in 3 places?

kevinrushforth: Yes, this could be done...I will note it for the next time we upgrade the version of gradle that we use (not worth it for this simple bumping of the minimum). Note that we would need two variables, since "current" and "minimum" aren't always the same (could be done similarly to boot JDK version).

Issue #164, opened 2018-08-23
Image dimensions sometimes off by 1 pixel when scaling down images with preserve ratio set to true


hjohn:
https://github.com/javafxports/openjdk-jfx/blob/e9e9719ab0abd673d343ea694092a69d9c448ad9/modules/javafx.graphics/src/main/java/com/sun/javafx/iio/common/ImageTools.java#L669-L671

See above code.

When an Image is loaded using `new Image()` you have the option of having it scaled down while preserving the aspect ratio.  However, the calculated dimensions are sometimes off by 1 pixel for images that have the exact same aspect ratio due to a rounding error.

A list of input sizes I tried and their resulting size when asking for Images to have a requested width and height of 108x108 pixels while preserving aspect ratio:

     width=800,height=1200 --> resulted in Image of 72.0x108.0
     width=1000,height=1500 --> resulted in Image of 72.0x107.0
     width=1400,height=2100 --> resulted in Image of 72.0x108.0
     width=1440,height=2160 --> resulted in Image of 72.0x108.0
     width=2000,height=3000 --> resulted in Image of 72.0x107.0

When debugging this for one of the cases that resulted in a y dimension of 107, the intermediate value was calculated for `sourceHeight * scale` was `107.99999`, however the code casts this directly to an `int` and you get 107.

The correct way to do this would be first check which dimension will be able to meet the requested width or height.  This dimension should then just copy the requested value (no need to calculate it).  The 2nd dimension is then calculated, and preferably rounded properly to get as close to the original aspect ratio as possible.

I can provide a PR for this.

hjohn:
A work-around, in case there is interest, is to set preserveRatio to false, and provide your own correctly calculated requestedWidth/Height -- you'll need to obtain the original image size for this though if you donot have it (I used AWT for this).

eugener:
Definitely provide a PR. I think @kevinrushforth and @johanvos would agree. Also make sure to have a test covering this.

kevinrushforth:
Please also read the [CONTRIBUTING](https://github.com/javafxports/openjdk-jfx/blob/develop/.github/CONTRIBUTING.md) doc, which is referred to by project README. Specifically, two things we will need are:

1. A signed OCA
2. A tracking bug filed in JBS

As @eugener mentioned, any proposed fix should be accompanied by a test case.

Thanks.

kevinrushforth:
> 1. A signed OCA

I see that this is already covered.

[...]

PR #170, opened 2018-08-25
JDK-8209968: Fix off by one error in image scale calculation

hjohn:
See issue #164.

Contains additional test for case when dimensions are computed with one of them set to 0, just to confirm that they're working correctly.


modules/javafx.graphics/src/test/java/test/com/sun/javafx/iio/common/ImageToolsTest.java
```diff
@@ -0,0 +1,76 @@
...
+import com.sun.javafx.iio.common.ImageTools;
+import static org.junit.Assert.*;
+import org.junit.Test;
+
+import java.util.Random;
+
+public class ImageToolsTest {
+    @Test
+    public void testIfComputeDimensionsReturnsCorrectValuesWhenAspectRatioIsPreserved() {
+        for (int i = 2; i < 1000; i += 2) {
```
eugener: Lot of "magic" numbers. Having constants with some explanatory names would be desirable.

eugener: Is there any need to have long loops ( 1000s of iterations ) for  this kind of test?

kevinrushforth:
In addition to answering Eugene's comments, you have a white-space error in the newly-added test file (see the Travis build log).

modules/javafx.graphics/src/test/java/test/com/sun/javafx/iio/common/ImageToolsTest.java
hjohn: I don't like this test case very much at all.  I'm having trouble thinking of one that would a) cover all edge cases, and b) does not use the same algorithm in the test itself.  Covering all edge cases is tough since you take 4 int parameters, making this a huge problem space, not just the usual of  covering a couple of branches.

If a test is sufficient that shows the old code breaking and the new code working, then I'll be happy to rewrite it.  As it is, I'm unsure how to make it cover all cases without a couple of loops.

eugener: I have no problem with loops at all. But do we need thousands of iterations?

modules/javafx.graphics/src/test/java/test/com/sun/javafx/iio/common/ImageToolsTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.Test;
+
+import java.util.Random;
+
+public class ImageToolsTest {
+    private static final int RANDOM_SEED = 1;  // A random seed
+    private static final int MIN_SIZE = 1;  // Smallest dimension to generate
+    private static final int MAX_SIZE = 1000;  // Largest dimension to generate
+    private static final int RANGE = MAX_SIZE - MIN_SIZE;  // Generate dimension range
+    private static final int RANDOM_COUNT = 10000;  // Number of random source and target dimensions to generate and verify
```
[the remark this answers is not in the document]
hjohn: It was running for 0.12s, I reduced it to 1000 loops.

I managed to come up with asserts that verify if it is closely matching the intended aspect, instead of just some boundary tests.  This catches the known failure cases (which were also caught already by the `assertTrue(msg, x == tw || y == th)` assert) without having to hard code any specific cases.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/iio/common/ImageToolsTest.java
hjohn: I've simplified the entire test.

hjohn: I've reduced the number of iterations, and added some additional asserts that accurately assert the expected values for the dimension that didn't reach it's maximum size.

modules/javafx.graphics/src/main/java/com/sun/javafx/iio/common/ImageTools.java
```diff
@@ -667,8 +667,8 @@ private static void computeUpdatedPixels(int sourceOffset,
                     finalHeight = (int) ((float) sourceHeight * finalWidth / sourceWidth);
```
kevinrushforth: I think the above two blocks should also round instead of truncate. Otherwise the smaller dimension will be calculated differently depending on whether the requested size is 0 or is a larger-than-needed number.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/iio/common/ImageToolsTest.java
```diff
@@ -0,0 +1,77 @@
...
+public class ImageToolsTest {
+    private static final int RANDOM_SEED = 1;  // A random seed
+    private static final int MIN_SIZE = 1;  // Smallest dimension to generate
+    private static final int MAX_SIZE = 1000;  // Largest dimension to generate
+    private static final int RANGE = MAX_SIZE - MIN_SIZE;  // Generate dimension range
+    private static final int RANDOM_COUNT = 1000;  // Number of random source and target dimensions to generate and verify
+
+    @Test
+    public void testIfComputeDimensionsReturnsValuesInCorrectRangeWhenAspectRatioIsPreserved() {
+        Random rnd = new Random(RANDOM_SEED);
```
kevinrushforth: I still want to see some tests with explicit values, including a few with one dimension being 0. You might create a utility method that takes the 4 values, which can then be called with a few carefully chosen values and then also called in the loop with random values. Especially in light of the comment I made on the fix itself.

As an example, the following will produce different results because the first will (after your fix) be rounded:

```
sx = 98, sy = 97, mx = 40, my= 50 -- result: x = 40, y = 40 (with your fix)
sx = 98, sy = 97, mx = 40, my= 0 -- result: x = 40, y = 39 (with or without your fix)
```

modules/javafx.graphics/src/main/java/com/sun/javafx/iio/common/ImageTools.java
hjohn: Alright, I rounded this as well, and covered these in the unit test.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/iio/common/ImageToolsTest.java
hjohn: I update the test with your suggestions, and included sufficient cases to cover all branches of the entire function.

kevinrushforth:
Since this is close to being ready to go in, can you send a review request email to openjfx-dev@openjdk.java.net referencing the JBS bug and this pull request? That way others can see it and comment if they want.

PR #173, opened 2018-08-27
JDK-8090930: Support mouse forward/back buttons.

brcolow:
This PR resolves https://bugs.openjdk.java.net/browse/JDK-8090930 which
is a request to enhance JavaFX to support additional mouse buttons beyond
Left, Middle, and Right. Specifically, this PR adds support for what are commonly
called the "Back" and "Forward" buttons.

Platform native windows (for Windows, macOS, and Linux) are adapted to
listen for BACK/FORWARD mouse button activity and fire respective
MouseEvents, as is currently done for PRIMARY, MIDDLE, and RIGHT.

We also add support for emulating mouse button press/release with
BACK/FORWARD to the native robot implementations so that e.g.
Robot#mousePress(MouseButton.BACK) will work as intended on Windows,
macOS, and Linux. We also add support to the MonocleRobot.

This new functionality is tested in an integrated way by adding two
tests to test/robot/javafx/scene/RobotTest.java that firstly use the
Robot to press and release the BACK/FORWARD buttons on a scene-graph
Node and then secondly ensure that the onMouseClicked event listener
properly fires a MOUSE_CLICKED event for the respective BACK/FORWARD
button. This ensures that the new mouse buttons are supported not only
on the Java side, but on the native side as well.

Because this PR touches nearby and logically related code, we also take
the opportunity to add mouse-wheel handling to `VNCScreen` as we added
back/forward mouse button support to VNC running via Monocle. There
is no documented convention (that I could find) for assigning buttons past
button 5, so we make the next two BACK/FORWARDS.

Going forwards I would like to add support for querying the system for
the number of mouse buttons supported (and which types are supported,
if possible). If this functionality was deemed worthy to add to JavaFX, I would
like to get some suggestions as to where it should be placed. I would think
adding some sort of static method to `MouseButton` may be appropriate,
as is done for `Screen.getPrimary()` which is similar in intent (querying the system
about its' configuration).


brcolow:
1.) Will work on doing a CSR very soon.

2.) I targeted towards openjfx12, if we can get it in that time-frame, that would be cool but not a big deal.

3.) I plan to add more tests, for sure. The commit message has a decent summary of the tests so far, but basically: adding a tests for the back/forward mouse button to the `RobotTest` to make sure that the correct event is fired when a robot calls `mousePress/mouseRelease(MouseButton.BACK/FORWARD)` and it works correctly on Windows, Mac, Linux, and Headless Monocle. Then I added a couple tests for the `MouseEvent` constructor, and made sure that when it is constructed with `MouseButton.BACK/FORWARD` that `event.is(Back/Forward)ButtonDown()` returns `true`. In other cases, I make sure it returns false. This was added to ` modules/javafx.graphics/src/test/java/test/javafx/scene/input/MouseEventTest.java`.

brcolow:
Preliminary CSR:

```
h2. Summary

Add support for extended mouse buttons MouseButton.BACK
and MouseButton.FORWARD. This means adding support to the
native windows, provided by the underlying operating system,
to listen for these mouse events as well as allowing the
Robot to trigger these events when the MouseButton argument
to Robot#mousePress and Robot#mouseRelease is one of the new
mouse buttons (BACK or FORWARD).

h2. Problem

JavaFX currently supports three mouse-buttons: Left (also known as
Primary), Right (also known as secondary) and Middle (also known
as Other). Any MouseEvents fired in a Scene will therefore only support
these three buttons. There is no way for someone using JavaFX to listen
for MouseEvents that involve other buttons (namely: BACK and FORWARD).
This is partly because the native windows provided by the operating
system do not listen for such mouse events and also because the JavaFX
API does not have a way to represent them (i.e. there are only 3 values
of the MouseButton enumeration).

h2. Solution

* Add MouseButton.BACK and MouseButton.FORWARD to the MouseButton enum.
* Make the native windows supplied by the OS listen to (and propagate
up to the Java API) mouse events for these extended buttons.
* Make the native Robots (Windows, macOS, Linux, and Headless Monocle)
"do the right thing" when mousePress or mouseRelease are called with
either of the new buttons (MouseButton.BACK and MouseButton.FORWARD).

h2. Specification



{code:title=modules/javafx.graphics/src/main/java/javafx/scene/input/MouseEvent.java|language=java}
/**
 * Constructs new MouseEvent event with null source and target.
 * @param eventType The type of the event.
 * @param x The x with respect to the scene.
 * @param y The y with respect to the scene.
 * @param screenX The x coordinate relative to screen.
 * @param screenY The y coordinate relative to screen.
 * @param button the mouse button used
 * @param clickCount number of click counts
 * @param shiftDown true if shift modifier was pressed.
 * @param controlDown true if control modifier was pressed.
 * @param altDown true if alt modifier was pressed.
 * @param metaDown true if meta modifier was pressed.
 * @param primaryButtonDown true if primary button was pressed.
 * @param middleButtonDown true if middle button was pressed.
 * @param secondaryButtonDown true if secondary button was pressed.
 * @param backButtonDown true if back button was pressed.
 * @param forwardButtonDown true if forward button was pressed
 * @param synthesized if this event was synthesized
 * @param popupTrigger whether this event denotes a popup trigger for current platform
 * @param stillSincePress see {@link #isStillSincePress() }
 * @param pickResult pick result. Can be null, in this case a 2D pick result
 *                   without any further values is constructed
 *                   based on the scene coordinates
 * @since 12
 */
public MouseEvent(
        @NamedArg("eventType") EventType<? extends MouseEvent> eventType,
        @NamedArg("x") double x, @NamedArg("y") double y,
        @NamedArg("screenX") double screenX, @NamedArg("screenY") double screenY,
        @NamedArg("button") MouseButton button,
        @NamedArg("clickCount") int clickCount,
        @NamedArg("shiftDown") boolean shiftDown,
        @NamedArg("controlDown") boolean controlDown,
        @NamedArg("altDown") boolean altDown,
        @NamedArg("metaDown") boolean metaDown,
        @NamedArg("primaryButtonDown") boolean primaryButtonDown,
        @NamedArg("middleButtonDown") boolean middleButtonDown,
        @NamedArg("secondaryButtonDown") boolean secondaryButtonDown,
        @NamedArg("backButtonDown") boolean backButtonDown,
        @NamedArg("forwardButtonDown") boolean forwardButtonDown,
        @NamedArg("synthesized") boolean synthesized,
        @NamedArg("popupTrigger") boolean popupTrigger,
        @NamedArg("stillSincePress") boolean stillSincePress,
        @NamedArg("pickResult") PickResult pickResult) {
    this(null, null, eventType, x, y, screenX, screenY, button, clickCount,
            shiftDown, controlDown, altDown, metaDown,
            primaryButtonDown, middleButtonDown, secondaryButtonDown,
            backButtonDown, forwardButtonDown,
            synthesized, popupTrigger, stillSincePress, pickResult);
}
{code}

{code:title=modules/javafx.graphics/src/main/java/javafx/scene/input/MouseButton.java|language=java}
public enum MouseButton {

    /**
     * Represents no button.
     */
    NONE,

    /**
     * Represents primary (button 1, usually the left) mouse button.
     */
    PRIMARY,

    /**
     * Represents middle (button 2) mouse button.
     */
    MIDDLE,

    /**
     * Represents secondary (button 3, usually the right) mouse button.
     */
    SECONDARY,

    /**
     * Represents back (button 4) mouse button.
     *
     * @since 12
     */
    BACK,

    /**
     * Represents forward (button 5) mouse button.
     *
     * @since 12
     */
    FORWARD,
}
{code}
```

brcolow:
Well this is a strange error:

`C:\projects\openjdk-jfx\modules\javafx.graphics\src\main\native-glass\win\GlassClipboard.cpp(86): error C2065: '_HASH_SEED': undeclared identifier`

1.) I haven't changed that file and 2.) it was working on the previous commit.

brcolow:
A ha, very interesting. It seems the Appveyor MSVC version was updated in the hour between the previous commit and the last commit. From the 2nd to last commit build log:

```
Starting process 'command 'C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.14.26428/bin/HostX64/x64/cl.exe''. Working directory: C:\projects\openjdk-jfx\modules\javafx.graphics Command: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.14.26428/bin/HostX64/x64/cl.exe /nologo /W3 /EHsc /c /DINLINE=__inline /DUNICODE /D_UNICODE /DWIN32 /DIAL /D_LITTLE_ENDIAN /DWIN32_LEAN_AND_MEAN /IC:/jdk11/include /IC:/jdk11/include/win32 /MDd /Od /Zi /DDEBUG /FS -IC:\projects\openjdk-jfx\modules\javafx.graphics\build\gensrc\headers\javafx.graphics -IC:\projects\openjdk-jfx\modules\javafx.graphics\src\main\native-glass\win /FoC:\projects\openjdk-jfx\modules\javafx.graphics\build\native\glass\win\GlassApplication.obj C:\projects\openjdk-jfx\modules\javafx.graphics\src\main\native-glass\win\GlassApplication.cpp
Successfully started process 'command 'C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.14.26428/bin/HostX64/x64/cl.exe''
```

Note the version `14.14.26428`

From the latest commit build log:

```
GlassClipboard.cpp
Starting process 'command 'C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.15.26726/bin/HostX64/x64/cl.exe''. Working directory: C:\projects\openjdk-jfx\modules\javafx.graphics Command: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.15.26726/bin/HostX64/x64/cl.exe /nologo /W3 /EHsc /c /DINLINE=__inline /DUNICODE /D_UNICODE /DWIN32 /DIAL /D_LITTLE_ENDIAN /DWIN32_LEAN_AND_MEAN /IC:/jdk11/include /IC:/jdk11/include/win32 /MDd /Od /Zi /DDEBUG /FS -IC:\projects\openjdk-jfx\modules\javafx.graphics\build\gensrc\headers\javafx.graphics -IC:\projects\openjdk-jfx\modules\javafx.graphics\src\main\native-glass\win /FoC:\projects\openjdk-jfx\modules\javafx.graphics\build\native\glass\win\GlassApplication.obj C:\projects\openjdk-jfx\modules\javafx.graphics\src\main\native-glass\win\GlassApplication.cpp
Successfully started process 'command 'C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.15.26726/bin/HostX64/x64/cl.exe''
```

Note the version: 14.15.26726

Therefore I think this is unrelated to this issue, and is quite possibly a bug that occurs when building OpenJFX on a MSVC with version at least 14.15.26726. I will look into that and create a separate PR if it's a real issue.

brcolow:
@kevinrushforth Regarding Swing and SWT interop testing:

SWT interop tests should be possible to write, and I will take a look at doing so soon.

From my understanding, though, it seems Swing does not support any mouse buttons past 3, so I don't think it will be possible to test Swing interop, unless you meant testing that pressing the BACK and FORWARD buttons don't fire any Swing events.

brcolow:
In order to add an SWT interop test I need to first enabled SWT testing. It seems that `-PFULL_TEST=true` and -PSWT_TEST=true` is not enough (at least on macOS) as it is still skipping task `:swt:test`:

```
> Task :swt:test SKIPPED

Skipping task ':swt:test' as task onlyIf is false.

:swt:test (Thread[main,5,main]) completed. Took 0.0 secs.

:createTestArgfiles (Thread[main,5,main]) started.
```

What is the standard way to actually compile and run SWT tests? Thanks.

kevinrushforth:
> What is the standard way to actually compile and run SWT tests? Thanks.

We don't run the SWT tests (or Swing interop tests) on our automated build machines, since they are generally headfull tests. The bigger problem, though, with SWT tests is that they were never made to work with Jigsaw (JDK 9). See the following in build.gradle in the ':swt' project:

```
    test {
        //enabled = IS_FULL_TEST && IS_SWT_TEST
        enabled = false // FIXME: JIGSAW -- support this with modules
        logger.info("JIGSAW Testing disabled for swt")
```

I recommend at least running a manual test for both Swing interop and SWT interop.

brcolow:
Okay, will do. Never used either one before with JavaFX so I will do some research. Will report back with my results (and maybe a commit containing the manual tests) soon.

brcolow:
@kevinrushforth This is kind of a silly question - but do you have a command-line invocation/gradle invocation/build file for running the manual SWT tests (in `tests/manual/swt`)? It's not trivial trying to invoke `javac`, `jar`, and `java` with a class-path that includes OpenJFX as built by this PR *and* SWT jars (which seem to be numerous). In case there isn't an easy way to run any of these manual tests, I was trying to work something up in `build.gradle` as a separate task that doesn't run by default but can be executed manually, something like:

```gradle
    test {
        //enabled = IS_FULL_TEST && IS_SWT_TEST
        enabled = false // FIXME: JIGSAW -- support this with modules
        logger.info("JIGSAW Testing disabled for swt")

        if (IS_MAC) {
            enabled = false
            logger.info("SWT tests are disabled on MAC, because Gradle test runner does not handle -XstartOnFirstThread properly (https://issues.gradle.org/browse/GRADLE-3290).")
        }
    }

    def manualTestCompile = task manualTestCompile (type: JavaCompile) {
        source = fileTree(dir: "${rootProject.projectDir}/tests/manual/swt", include: "**/*Test.java")
        classpath = sourceSets.main.compileClasspath
        classpath += files("${project(":swt").buildDir}/libs/javafx-swt.jar")
        destinationDir = file("$rootProject.projectDir}/tests/manual/swt")
    }

    def manualTest = task manualTest(type: JavaExec) {
        classpath = sourceSets.main.runtimeClasspath
        classpath += files("${project(":swt").buildDir}/libs/javafx-swt.jar")
        main = "FXCanvasMouseButtonEventsTest"
        workingDir = "${rootProject.projectDir}/tests/manual/swt"
    }

    compileTargets { t ->
        manualTest.dependsOn project.tasks.findAll {
            (it.name == "javafxSwt$t.capital")
        }
    }
    manualTest.dependsOn manualTestCompile
}
```

Ideally we would expand this to run more than the new test I added, but just as a proof-of-concept. However this is currently not working (and probably is expected not to, as it's quite hodge-podge):

```
20:59:57.085 [ERROR] [system.err] Error: Could not find or load main class FXCanvasMouseButtonEventsTest
20:59:57.085 [ERROR] [system.err] Caused by: java.lang.ClassNotFoundException: FXCanvasMouseButtonEventsTest
```

I have added a commit that includes the new test file `FXCanvasMouseButtonEventsTest.java`.

brcolow:
This may be a severe corruption of how gradle is intended to be used, but I was able to run the manual test I added with the following:

```gradle
    test {
        enabled = IS_FULL_TEST && IS_SWT_TEST
        // enabled = false // FIXME: JIGSAW -- support this with modules
        logger.info("JIGSAW Testing disabled for swt")

        if (IS_MAC) {
            enabled = false
            logger.info("SWT tests are disabled on MAC, because Gradle test runner does not handle -XstartOnFirstThread properly (https://issues.gradle.org/browse/GRADLE-3290).")
        }
    }

    def manualTestCompile = task manualTestCompile (type: JavaCompile) {
        source = fileTree(dir: "${rootProject.projectDir}/tests/manual/swt", include: "**/*Test.java")
        classpath = sourceSets.main.compileClasspath
        classpath += files("${project(":swt").buildDir}/libs/javafx-swt.jar")
        destinationDir = file("${rootProject.projectDir}/tests/manual/swt")
    }

    def manualTestJar = task manualTestJar(type: Jar) {
        from ("${rootProject.projectDir}/tests/manual/swt") {
            include "**/*Test.class"
        }
        from (zipTree("${project(":swt").buildDir}/libs/javafx-swt.jar"))
        sourceSets.main.compileClasspath.each { f ->
            from (zipTree(f))
        }
        destinationDir = file("${rootProject.projectDir}/build/manual/swt")
        archiveName = "swt-manual-test.jar"
        manifest {
            attributes "Main-Class": "FXCanvasMouseButtonEventsTest"
        }
        exclude 'META-INF/*.RSA', 'META-INF/*.SF', 'META-INF/*.DSA'
    }

    def manualTest = task manualTest(type: JavaExec) {
        main = "-jar"
        args = ["swt-manual-test.jar"]
        classpath = sourceSets.main.runtimeClasspath
        classpath += sourceSets.main.compileClasspath
        classpath += files("${project(":swt").buildDir}/libs/javafx-swt.jar")
        workingDir = "${rootProject.projectDir}/build/manual/swt"
    }

    compileTargets { t ->
        manualTest.dependsOn project.tasks.findAll {
            (it.name == "javafxSwt$t.capital")
        }
    }
    manualTest.dependsOn manualTestCompile
    manualTest.dependsOn manualTestJar
```

brcolow:
@kevinrushforth I've been able to verify, at a minimum, that the extended mouse buttons don't cause any errors/crashes with SWT interop. So far my attempts at trying to get the mouse events for the extended buttons to fire in JavaFX from SWT haven't been successful. As I noted above, trying to run a manual test is real tricky because of all the command-line arguments that must be passed in. So far, compiling a fat JAR with the SWT dependencies and running it with `_JAVA_OPTIONS` as `--module-path="C:\Users\brcolow\dev\openjdk-jfx\build\sdk\lib" --add-modules=javafx.controls -Djava.library.path="C:/Users/brcolow/dev/openjdk-jfx/build/modular-sdk/modules_libs/javafx.graphics;C:/Users/brcolow/dev/openjdk-jfx/build/modular-sdk/modules_libs/javafx.media;C:/Users/brcolow/dev/openjdk-jfx/build/modular-sdk/modules_libs/javafx.web;" --patch-module="javafx.base=C:/Users/brcolow/dev/openjdk-jfx/build/modular-sdk/modules/javafx.base" --patch-module="javafx.controls=C:/Users/brcolow/dev/openjdk-jfx/build/modular-sdk/modules/javafx.controls" --patch-module="javafx.fxml=C:/Users/brcolow/dev/openjdk-jfx/build/modular-sdk/modules/javafx.fxml" --patch-module="javafx.graphics=C:/Users/brcolow/dev/openjdk-jfx/build/modular-sdk/modules/javafx.graphics" --patch-module="javafx.media=C:/Users/brcolow/dev/openjdk-jfx/build/modular-sdk/modules/javafx.media" --patch-module="javafx.swing=C:/Users/brcolow/dev/openjdk-jfx/build/modular-sdk/modules/javafx.swing" --patch-module="javafx.web=C:/Users/brcolow/dev/openjdk-jfx/build/modular-sdk/modules/javafx.web" --add-reads=javafx.graphics=jdk.unsupported --add-exports=javafx.base/com.sun.javafx.logging=javafx.controls --add-exports=javafx.base/com.sun.javafx.logging=javafx.fxml --add-exports=javafx.base/com.sun.javafx.logging=javafx.swing --add-modules=javafx.graphics --add-modules=javafx.graphics --add-modules=javafx.fxml'` seems to be a working approach.

As I mentioned above, Swing interop testing won't be possible (manually or automatically) as Swing does not support mouse buttons past 3.

I am assuming no one has actually run any of the SWT manual tests for a while, or maybe there is an internal build file/script at Oracle for doing so? If not, I can open a PR for the gradle tasks I showed above. The idea is that the `:swt:manualTest` task would never be invoked unless explicitly called (i.e. it would not run as part of the normal build). I am not sure if what I have above works in that way, though.

To summarize: testing SWT interop has been a serious headache for me. I have at least contributed a test file (in `tests/manual/swt`) that would demonstrate the mouse button aspect of JavaFX/SWT interoperation. Sorry in advance for making this more complicated than it should be.

brcolow:
@kevinrushforth If I have written too much and made this too complicated for you to advise on let me know and I will take another stab at it.

kevinrushforth:
@brcolow The level of detail is fine. I haven't had time to look at it is all. As a quick note, it's probably best if manual tests are done outside of gradle for SWT, especially since this was never really finished for JDK 9.

kevinrushforth:
I get the following new test failure on Linux in RobotTest:

```
test.robot.javafx.scene.RobotTest > testMouseWheel FAILED
    java.lang.AssertionError: Timeout while waiting for button.onScroll().
        at org.junit.Assert.fail(Assert.java:91)
        at test.robot.javafx.scene.RobotTest.waitForLatch(RobotTest.java:416)
        at test.robot.javafx.scene.RobotTest.testMouseWheel(RobotTest.java:281)
```

If I run just that one test, it works fine. Further testing shows that the test that clicks the "FORWARD" button interferes with the mouse wheel in some way as to make it not work properly. Without more analysis I can't say whether this is a test issue or an actual issue with the Linux code to implement the new mouse buttons.

It runs fine on Mac and Windows (which suggests, but doesn't prove, that it's an implementation bug on Linux rather than a test bug).

modules/javafx.graphics/src/main/native-glass/win/FullScreenWindow.cpp
```diff
@@ -323,6 +323,9 @@ LRESULT FullScreenWindow::WindowProc(UINT msg, WPARAM wParam, LPARAM lParam)
         case WM_MBUTTONDOWN:
         case WM_MBUTTONUP:
         case WM_MBUTTONDBLCLK:
+        case WM_XBUTTONDOWN:
```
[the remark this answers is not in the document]
brcolow: Yes. X1 and X2 are both grouped together in terms of the window manager events (there are only WM_XBUTTONDOWN and WM_XBUTTONUP).

modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/SwingEvents.java
```diff
@@ -81,6 +81,10 @@ public static int mouseButtonToEmbedMouseButton(int button, int extModifiers) {
             abstractButton = AbstractEvents.MOUSEEVENT_MIDDLE_BUTTON;
         } else if ((extModifiers & MouseEvent.BUTTON3_DOWN_MASK) != 0) {
             abstractButton = AbstractEvents.MOUSEEVENT_SECONDARY_BUTTON;
+        } else if ((extModifiers & MouseEvent.getMaskForButton(4)) != 0) {
+            abstractButton = AbstractEvents.MOUSEEVENT_BACK_BUTTON;
+        } else if ((extModifiers & MouseEvent.getMaskForButton(5)) != 0) {
+            abstractButton = AbstractEvents.MOUSEEVENT_FORWARD_BUTTON;
         }
```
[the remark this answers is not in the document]
brcolow: Good catch - I think I may have missed that.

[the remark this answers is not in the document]
brcolow: Ah, I remember now. AWT's `MouseEvent.java` does not define any mouse buttons above 3 (i.e. AWT does not support the back/forward mouse buttons). So I think this is the best we can do.

brcolow:
I addressed the in-line comments. I will investigate the failure on Linux.

.travis.yml
```diff
@@ -60,8 +60,6 @@ matrix:
 before_script:
   - echo $JAVA_OPTS
   - sh ./gradlew --version
-  # run on xvfb screen (for linux).
-  - if [ "${TRAVIS_OS_NAME}" == "linux" ]; then export DISPLAY=:99.0 && sh -e /etc/init.d/xvfb start; fi
```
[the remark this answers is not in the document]
kevinrushforth: There is still a blank line diff in `.travis.yml` which should be reverted (so that it doesn't show up the patch at all).

brcolow:
What in the world am I missing...how is it that `isBackButtonDown` and `isForwardButtonDown` are both returning true when using the full constructor:

https://github.com/javafxports/openjdk-jfx/blob/bbd5ca2e42ba3713e6138a04b59c76e8977d8d97/modules/javafx.graphics/src/test/java/test/javafx/scene/input/MouseDragEventTest.java#L266-L280

This seems impossible... `MouseDragEvent` calls `MouseEvent`'s constructor which just sets `backButtonDown` and `forwardButtonDown` member variables. Am I missing something super obvious?

brcolow:
I forgot to answer your two general questions.

1.) I do have access to a 5-button mouse and I will run the tests on all 3 platforms during the Thanksgiving holidays.
2.) I have manually tested that mouse forward/back buttons work on a real JavaFX application on Windows. I will verify this with Linux and macOS also during the break.

brcolow:
Building JavaFX on my new PC is *so* much faster. Will make developing much easier going forward. I am able to reproduce the test failure you got on Linux (where `testMouseWheel` is failing). Your tip about the forward button interfering in some way will be very helpful going forward. I will do some research and see if I can fix it.

brcolow:
It doesn't seem to be an issue of the JavaFX window losing focus because after clicking `MouseButton.FORWARD` or `MouseButton.BACK` the JavaFX application still responds to keyboard key pressed (for example with a VBox containing a `Button` and `TextField`, a down arrow-key press after clicking with FORWARD/BACK on the `Button` will cycle focus to the `TextField`. It seems more like after a FORWARD/BACK button press X11 is not sending scroll events until a mouse click with left/right happens. It's hard to think it's a problem with the JavaFX event handling because macOS and Windows are working correctly. I will see if anyone ever reported anything like this about X11 before and may even try to create a native X11 test application to demo it (hopefully I won't have to go that far, lol).

brcolow:
I added debugging statements to all the `returns` of the `process_events` handler to see if when it is in the "locked" state where it is not sending scroll events if there are any early outs in the returns before the large event type switch statement. There are no early outs. Therefore I am concluding that GTK is actually not calling the event handler with mouse scroll events at all after a FORWARD or MOUSE button click UNTIL one clicks with left/right mouse (and in fact one doesn't even need to click on the JavaFX application itself, just anywhere on the screen).

I also want to note one further thing. After clicking mouse FORWARD/BACK and it is in the "locked" state, hovering over the GNOME toolbar icons (that default to the left side of the screen) causes them not to change to the mouse-over state (it's sort of like a highlight that happens for responsiveness). Just using weird descriptions now but it is like the JavaFX window is "super-focused".

baedert:
Um...? What is this about? Am I supposed to read all the comments?

brcolow:
@baedert Sorry, I should have been more specific. Basically the last 4 comments by me should be a good overview but here's a quick summary:

This PR is about adding support for mouse button 4 and 5 to JavaFX. On Linux JavaFX uses GTK3 to handle mouse events. We register the following event handler with GTK's `gdk_event_handler_set`:

```cpp
static void process_events(GdkEvent* event, gpointer data)
{
    GdkWindow* window = event->any.window;
    WindowContext *ctx = window != NULL ? (WindowContext*)
        g_object_get_data(G_OBJECT(window), GDK_WINDOW_DATA_CONTEXT) : NULL;

    if ((window != NULL)
            && !is_window_enabled_for_event(window, ctx, event->type)) {
        return;
    }

    if (ctx != NULL && ctx->hasIME() && ctx->filterIME(event)) {
        return;
    }

    glass_evloop_call_hooks(event);

    if (ctx != NULL && dynamic_cast<WindowContextPlug*>(ctx) && ctx->get_gtk_window()) {
        WindowContextPlug* ctx_plug = dynamic_cast<WindowContextPlug*>(ctx);
        if (!ctx_plug->embedded_children.empty()) {
            // forward to child
            ctx = (WindowContext*) ctx_plug->embedded_children.back();
            window = ctx->get_gdk_window();
        }
    }

    if (is_in_drag()) {
        process_dnd_source(window, event);
    }

    if (ctx != NULL) {
        EventsCounterHelper helper(ctx);
        try {
            switch (event->type) {
                case GDK_PROPERTY_NOTIFY:
                    ctx->process_property_notify(&event->property);
                    gtk_main_do_event(event);
                    break;
                case GDK_CONFIGURE:
                    ctx->process_configure(&event->configure);
                    gtk_main_do_event(event);
                    break;
                case GDK_FOCUS_CHANGE:
                    ctx->process_focus(&event->focus_change);
                    gtk_main_do_event(event);
                    break;
                case GDK_DESTROY:
                    destroy_and_delete_ctx(ctx);
                    gtk_main_do_event(event);
                    break;
                case GDK_DELETE:
                    ctx->process_delete();
                    break;
                case GDK_EXPOSE:
                case GDK_DAMAGE:
                    ctx->process_expose(&event->expose);
                    break;
                case GDK_WINDOW_STATE:
                    ctx->process_state(&event->window_state);
                    gtk_main_do_event(event);
                    break;
                case GDK_BUTTON_PRESS:
                case GDK_BUTTON_RELEASE:
                    ctx->process_mouse_button(&event->button);
                    break;
                case GDK_MOTION_NOTIFY:
                    ctx->process_mouse_motion(&event->motion);
                    gdk_event_request_motions(&event->motion);
                    break;
                case GDK_SCROLL:
                    ctx->process_mouse_scroll(&event->scroll);
                    break;
                case GDK_ENTER_NOTIFY:
                case GDK_LEAVE_NOTIFY:
                    ctx->process_mouse_cross(&event->crossing);
                    break;
                case GDK_KEY_PRESS:
                case GDK_KEY_RELEASE:
                    ctx->process_key(&event->key);
                    break;
                case GDK_DROP_START:
                case GDK_DRAG_ENTER:
                case GDK_DRAG_LEAVE:
                case GDK_DRAG_MOTION:
                    process_dnd_target(ctx, &event->dnd);
                    break;
                case GDK_MAP:
                    ctx->process_map();
                    // fall-through
                case GDK_UNMAP:
                case GDK_CLIENT_EVENT:
                case GDK_VISIBILITY_NOTIFY:
                case GDK_SETTING:
                case GDK_OWNER_CHANGE:
                    gtk_main_do_event(event);
                    break;
                default:
                    break;
            }
        } catch (jni_exception&) {
        }
    } else {

        if (window == gdk_screen_get_root_window(gdk_screen_get_default())) {
            if (event->any.type == GDK_PROPERTY_NOTIFY) {
                if (event->property.atom == gdk_atom_intern_static_string("_NET_WORKAREA")
                        || event->property.atom == gdk_atom_intern_static_string("_NET_CURRENT_DESKTOP")) {
                    screen_settings_changed(gdk_screen_get_default(), NULL);
                }
            }
        }

        //process only for non-FX windows
        if (process_events_prev != NULL) {
            (*process_events_prev)(event, data);
        } else {
            gtk_main_do_event(event);
        }
    }
}
```

For some reason, after a mouse button 4 or 5 press (also known as BACK/FORWARD mouse buttons) GTK stops sending mouse scroll wheel events to the event handler until a left or right mouse click happens (it does not even need to happen in the JavaFX application but anywhere on the screen). None of the early-out returns in the above event handler are being executed so it doesn't seem to be on "our" end. Maybe there is some sort of default event handling/filter that is occurring? Additionally, after a mouse button 4 or 5 press, the GNOME toolbar icons (that default to the left side of the screen) no longer respond to a mouse hover event (i.e. there is no hover highlight) so it is in some sort of "locked" or "super-focus" mode. I was just checking in with you to see if any of this rings any bells with you - if not no worries and I will continue to research it.

johanvos:
Good catch about the `event->state` being `0`.
Your patch should fix that, but I'm worried about why the state is 0 to start with. I don't think your patch has an impact on performance, as most cases are probably handled by the first `if` statement, exactly how they would use to be handled.

brcolow:
Any thoughts on if I should add the drag-drop tests to `RobotTest` or not?

kevinrushforth:
> Test adding drag/drop tests. 

Nice. I think these might be the first real automated tests of mouse drag (we use it in the TabDragPolicyTest, but it will be good to verify that the events are generated correctly).

tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
```diff
@@ -251,6 +256,54 @@ public void testMouseClickForwardButton() {
...
+        testMouseDrag(MouseButton.FORWARD);
+    }
+
+    @Test
+    public void testMouseDragBack() {
+        testMouseDrag(MouseButton.BACK);
+    }
+
+    public void testMouseDrag(MouseButton mouseButton) {
+        CountDownLatch mouseDragLatch = new CountDownLatch(1);
```
kevinrushforth: I don't see where you wait for this latch.

kevinrushforth:
So to answer your question about whether it would be good to add mouse drag tests, I think testing that a MouseDrag event is delivered would be a good thing, but I wouldn't actually test Drag And Drop here.

brcolow:
I was messing with the commits I think when you saw it, here is the current test:

```java
   @Test
    public void testMouseDragPrimary() {
        testMouseDrag(MouseButton.PRIMARY);
    }

    @Test
    public void testMouseDragSecondary() {
        testMouseDrag(MouseButton.SECONDARY);
    }

    @Test
    public void testMouseDragMiddle() {
        testMouseDrag(MouseButton.MIDDLE);
    }

    @Test
    public void testMouseDragForward() {
        testMouseDrag(MouseButton.FORWARD);
    }

    @Test
    public void testMouseDragBack() {
        testMouseDrag(MouseButton.BACK);
    }
    public void testMouseDrag(MouseButton mouseButton) {
        CountDownLatch mouseDragLatch = new CountDownLatch(1);
        CountDownLatch setSceneLatch = new CountDownLatch(1);
        Label label = new Label("Source");
        InvalidationListener invalidationListener = observable -> setSceneLatch.countDown();
        Util.runAndWait(() -> {
            label.setOnMouseDragged(event -> {
                if (event.getButton() == mouseButton) {
                    mouseDragLatch.countDown();
                }
            });
            scene = new Scene(new HBox(label));
            stage.sceneProperty().addListener(observable -> {
                setSceneLatch.countDown();
                stage.sceneProperty().removeListener(invalidationListener);
            });
            stage.setScene(scene);
        });
        waitForLatch(setSceneLatch, 5, "Timeout while waiting for scene to be set on stage.");
        Util.runAndWait(() -> {
            int mouseX = (int) (scene.getWindow().getX() + scene.getX() +
                    label.getLayoutX() + label.getLayoutBounds().getWidth() / 2);
            int mouseY = (int) (scene.getWindow().getY() + scene.getY() +
                    label.getLayoutY() + label.getLayoutBounds().getHeight() / 2);
            robot.mouseMove(mouseX, mouseY);
            robot.mousePress(mouseButton);
            for (int i = 1; i <= 50; i ++) {
                robot.mouseMove(mouseX + i, mouseY);
            }
            robot.mouseRelease(mouseButton);
        });
        waitForLatch(mouseDragLatch, 5, "Timeout while waiting for button.onMouseDragged().");
    }
```

brcolow:
I will rebase this tomorrow now that #215 has been merged.

kevinrushforth:
Great. I'll take a final look early next week.

brcolow:
I have addressed all of the comments. I removed the trailing periods on `@param` annotations and will improve the grammar in a follow-up PR, if that's okay. That's strange that the drag tests failed on macOS for you - they worked locally for me. Is this a merge-stopper?

kevinrushforth:
> That's strange that the drag tests failed on macOS for you - they worked locally for me. Is this a merge-stopper?

I'd at least like to know why they fail (they do so consistently for me). I'm running macOS 10.12.6 on a retina display (although for the drag test, the latter should not matter).

brcolow:
I'm running non-retina 10.13.6 (kind of oddball). Let me confirm that the tests succeed again locally for me,  will update you.

brcolow:
No I was wrong, sorry about that. I didn't properly rebase the latest changes - they fail for me on my mac as well. Will try to investigate.

kevinrushforth:
> they fail for me on my mac as well. Will try to investigate.

OK

brcolow:
Adding a print debug so that the mouse events are printed for the drag tests I get:

```sh
event: MouseEvent [source = Label@5067c765[styleClass=label]'Source', target = Text[text="Source", x=0.0, y=0.0, alignment=LEFT, origin=BASELINE, boundsType=LOGICAL_VERTICAL_CENTER, font=Font[name=System Regular, family=System, style=Regular, size=13.0], fontSmoothingType=LCD, fill=0x333333ff], eventType = MOUSE_DRAGGED, consumed = false, x = 72.0, y = 8.0, z = 0.0, button = PRIMARY, primaryButtonDown, pickResult = PickResult [node = null, point = Point3D [x = 72.0, y = 8.0, z = 0.0], distance = 31.72243186433546]
event: MouseEvent [source = Label@69a78910[styleClass=label]'Source', target = Text[text="Source", x=0.0, y=0.0, alignment=LEFT, origin=BASELINE, boundsType=LOGICAL_VERTICAL_CENTER, font=Font[name=System Regular, family=System, style=Regular, size=13.0], fontSmoothingType=LCD, fill=0x333333ff], eventType = MOUSE_DRAGGED, consumed = false, x = 72.0, y = 8.0, z = 0.0, button = SECONDARY, secondaryButtonDown, pickResult = PickResult [node = null, point = Point3D [x = 72.0, y = 8.0, z = 0.0], distance = 31.72243186433546]
event: MouseEvent [source = Label@7366185b[styleClass=label]'Source', target = Text[text="Source", x=0.0, y=0.0, alignment=LEFT, origin=BASELINE, boundsType=LOGICAL_VERTICAL_CENTER, font=Font[name=System Regular, family=System, style=Regular, size=13.0], fontSmoothingType=LCD, fill=0x333333ff], eventType = MOUSE_DRAGGED, consumed = false, x = 72.0, y = 8.0, z = 0.0, button = BACK, backButtonDown, pickResult = PickResult [node = null, point = Point3D [x = 72.0, y = 8.0, z = 0.0], distance = 31.72243186433546]
event: MouseEvent [source = Label@6b14fcd6[styleClass=label]'Source', target = Text[text="Source", x=0.0, y=0.0, alignment=LEFT, origin=BASELINE, boundsType=LOGICAL_VERTICAL_CENTER, font=Font[name=System Regular, family=System, style=Regular, size=13.0], fontSmoothingType=LCD, fill=0x333333ff], eventType = MOUSE_DRAGGED, consumed = false, x = 72.0, y = 8.0, z = 0.0, button = FORWARD, forwardButtonDown, pickResult = PickResult [node = null, point = Point3D [x = 72.0, y = 8.0, z = 0.0], distance = 31.72243186433546]

```

It seems like no drag event is emitted for the middle button and then the that makes the others fail. Maybe that's because I don't actually have a middle button on the mac? If this is the case it relates to a question I asked in the original post for this PR about possible future work in using native JavaFX to query the system and ask what mouse buttons are supported.

kevinrushforth:
> It seems like no drag event is emitted for the middle button and then the that makes the others fail. Maybe that's because I don't actually have a middle button on the mac?

I just tested that theory, by connecting a 3-button mouse to my Mac, but it still fails. I then ran a manual test program, and it does generate MouseDrag events for the middle button. Maybe the problem is in the robot implementation for buttons other than primary and secondary?

In any case, I recommend to file a new bug and skip those three failing tests on Mac.

brcolow:
Must be the robot implementation, indeed. I have done as suggested (filed a new bug (https://bugs.openjdk.java.net/browse/JDK-8215376 - can't seem to get JIRA code formatting to take) and ignored them on macOS).

.idea/misc.xml
```diff
@@ -8,4 +8,3 @@
     <output url="file://$PROJECT_DIR$/.idea/out" />
   </component>
 </project>
-
```
kevinrushforth: This file still shows up in the aggregate patch as being modified (the file in your PR has a blank line removed vs the original).

brcolow: Sorry about that - should have checked.

kevinrushforth:
The changes look good now.

However, I was doing one last sanity check on Linux and I now get multiple failures. Since nothing significant changed between the last passing run and the current failing run, it might be an order-of-execution issue (the tests are run in an unpredictable order). This suggests that some of the tests are interfering with others based on the state that they leave the system in (e.g., if any of the buttons or keys are left pressed, that could affect subsequent tests). I'll try one quick test to validate this theory, but you might want to take a look at it.

Here are the failing tests on my most recent Linux run:

```
test.robot.javafx.scene.RobotTest > testMouseClickForward FAILED
    java.lang.AssertionError: Timeout while waiting for button.onMouseClicked().
        at org.junit.Assert.fail(Assert.java:91)
        at test.robot.javafx.scene.RobotTest.waitForLatch(RobotTest.java:812)
        at test.robot.javafx.scene.RobotTest.testMouseAction(RobotTest.java:417)
        at test.robot.javafx.scene.RobotTest.testMouseClickForward(RobotTest.java:337)

test.robot.javafx.scene.RobotTest > testMouseClickBack FAILED
    java.lang.AssertionError: Timeout while waiting for button.onMouseClicked().
        at org.junit.Assert.fail(Assert.java:91)
        at test.robot.javafx.scene.RobotTest.waitForLatch(RobotTest.java:812)
        at test.robot.javafx.scene.RobotTest.testMouseAction(RobotTest.java:417)
        at test.robot.javafx.scene.RobotTest.testMouseClickBack(RobotTest.java:342)

test.robot.javafx.scene.RobotTest > testMouseDragPrimary FAILED
    java.lang.AssertionError: Timeout while waiting for button.onMouseDragged().
        at org.junit.Assert.fail(Assert.java:91)
        at test.robot.javafx.scene.RobotTest.waitForLatch(RobotTest.java:812)
        at test.robot.javafx.scene.RobotTest.testMouseDrag(RobotTest.java:551)
        at test.robot.javafx.scene.RobotTest.testMouseDragPrimary(RobotTest.java:494)

test.robot.javafx.scene.RobotTest > testMouseDragSecondary FAILED
    java.lang.AssertionError: Timeout while waiting for button.onMouseDragged().
        at org.junit.Assert.fail(Assert.java:91)
        at test.robot.javafx.scene.RobotTest.waitForLatch(RobotTest.java:812)
        at test.robot.javafx.scene.RobotTest.testMouseDrag(RobotTest.java:551)
        at test.robot.javafx.scene.RobotTest.testMouseDragSecondary(RobotTest.java:499)

test.robot.javafx.scene.RobotTest > testMouseDragMiddle FAILED
    java.lang.AssertionError: Timeout while waiting for button.onMouseDragged().
        at org.junit.Assert.fail(Assert.java:91)
        at test.robot.javafx.scene.RobotTest.waitForLatch(RobotTest.java:812)
        at test.robot.javafx.scene.RobotTest.testMouseDrag(RobotTest.java:551)
        at test.robot.javafx.scene.RobotTest.testMouseDragMiddle(RobotTest.java:505)

test.robot.javafx.scene.RobotTest > testMouseDragForward FAILED
    java.lang.AssertionError: Timeout while waiting for button.onMouseDragged().
        at org.junit.Assert.fail(Assert.java:91)
        at test.robot.javafx.scene.RobotTest.waitForLatch(RobotTest.java:812)
        at test.robot.javafx.scene.RobotTest.testMouseDrag(RobotTest.java:551)
        at test.robot.javafx.scene.RobotTest.testMouseDragForward(RobotTest.java:511)

test.robot.javafx.scene.RobotTest > testMouseDragBack FAILED
    java.lang.AssertionError: Timeout while waiting for button.onMouseDragged().
        at org.junit.Assert.fail(Assert.java:91)
        at test.robot.javafx.scene.RobotTest.waitForLatch(RobotTest.java:812)
        at test.robot.javafx.scene.RobotTest.testMouseDrag(RobotTest.java:551)
        at test.robot.javafx.scene.RobotTest.testMouseDragBack(RobotTest.java:517)
```

kevinrushforth:
A couple additional findings.

* I think the drag tests are unstable, not sure why. I was able to get the drag test for primary button passing on Linux only after breaking the test up into multiple Util.runAndWait blocks. Maybe `@ignore` them pending investigation?

* `testMouseAction` doesn't release all mouse buttons. It should be:

```
Util.runAndWait(() -> robot.mouseRelease(MouseButton.PRIMARY, MouseButton.SECONDARY, MouseButton.MIDDLE, MouseButton.BACK, MouseButton.FORWARD));
```

* Related to the above, if one test fails it can cause additional failures because the cleanup will be skipped. Maybe consider an `@After` method to release all buttons / keys / whatever other cleanup you might need?

brcolow:
Okay. I added a `@After` annotated cleanup method that releases all mouse buttons and the A key (the only key used in the tests). I also `@Ignore`d the drag tests with the reason "Flaky".

brcolow:
I see that Joe Darcy made two comments in the JIRA https://bugs.openjdk.java.net/browse/JDK-8210895. I don't seem to have the permissions to make comments (the `Add` button is disabled). Here was my response to his questions/comments:

```
No opinion about documenting the existing constructors - I will do as instructed.

In terms of the enum values ending with a comma that was the style that was used in MouseButton.java (the last value, which was SECONDARY) ended in a comma, so I stuck with that style. I believe it's done that way sometimes so that if new values are added in the future the diffs look better - but that's a very trivial concern. I am find with changing it to a semi-colon.
```

tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
```diff
@@ -490,28 +486,33 @@ public void testMouseClickThrowsNPEForNullArgument() {
     }
 
     @Test
+    @Ignore("Flaky")
     public void testMouseDragPrimary() {
         testMouseDrag(MouseButton.PRIMARY);
     }
 
     @Test
+    @Ignore("Flaky")
```
kevinrushforth: If you want, you can file a follow-up bug to address this, but not necessary.

brcolow: I will be attempting a fix when I look at https://bugs.openjdk.java.net/browse/JDK-8215376

kevinrushforth: Great. In that case, maybe add a comment (or add JDK-8215376 to the `@Ignore` string?)

brcolow:
Awesome, thanks. I added a comment, updated the CSR, and pushed a fixup commit.

kevinrushforth:
The update to the test looks good.

I added a comment to the CSR regarding adding the note about the state of back and forward buttons to the existing MouseDragEvent constructors (to match what you did for MouseEvent). After doing that, you can move the CSR to the Finalize state.

brcolow:
In the case of `MouseDragEvent` the diff looks nasty:

```diff
@@ -171,35 +235,43 @@ public MouseDragEvent(@NamedArg("source") Object source, @NamedArg("target") Eve
     /**
      * Constructs new MouseDragEvent event with null source and target.
      *
-     * @param eventType The type of the event.
-     * @param x The x with respect to the scene.
-     * @param y The y with respect to the scene.
-     * @param screenX The x coordinate relative to screen.
-     * @param screenY The y coordinate relative to screen.
+     * Both {@link #isBackButtonDown()}  and {@link #isForwardButtonDown()}
+     * will return {@literal false} for this event.
+     *
+     * @param eventType The type of the event
+     * @param x The x with respect to the scene


+     * @param y The y with respect to the scene
+     * @param screenX The x coordinate relative to screen
+     * @param screenY The y coordinate relative to screen
      * @param button the mouse button used
      * @param clickCount number of click counts
-     * @param shiftDown true if shift modifier was pressed.
-     * @param controlDown true if control modifier was pressed.
-     * @param altDown true if alt modifier was pressed.
-     * @param metaDown true if meta modifier was pressed.
-     * @param primaryButtonDown true if primary button was pressed.
-     * @param middleButtonDown true if middle button was pressed.
-     * @param secondaryButtonDown true if secondary button was pressed.
+     * @param shiftDown true if shift modifier was pressed
+     * @param controlDown true if control modifier was pressed
+     * @param altDown true if alt modifier was pressed
+     * @param metaDown true if meta modifier was pressed
+     * @param primaryButtonDown true if primary button was pressed
+     * @param middleButtonDown true if middle button was pressed
+     * @param secondaryButtonDown true if secondary button was pressed
      * @param synthesized if this event was synthesized
      * @param popupTrigger whether this event denotes a popup trigger for current platform
      * @param pickResult pick result. Can be null, in this case a 2D pick result
      *                   without any further values is constructed
      *                   based on the scene coordinates
-     * @param gestureSource source object of the ongoing gesture.
+     * @param gestureSource source object of the ongoing gesture
      * @since JavaFX 8.0
      */
     public MouseDragEvent(@NamedArg("eventType") EventType<MouseDragEvent> eventType,
-            @NamedArg("x") double x, @NamedArg("y") double y, @NamedArg("screenX") double screenX, @NamedArg("screenY") double screenY,
-            @NamedArg("button") MouseButton button, @NamedArg("clickCount") int clickCount,
-            @NamedArg("shiftDown") boolean shiftDown, @NamedArg("controlDown") boolean controlDown, @NamedArg("altDown") boolean altDown, @NamedArg("metaDown") boolean metaDown,
-            @NamedArg("primaryButtonDown") boolean primaryButtonDown, @NamedArg("middleButtonDown") boolean middleButtonDown, @NamedArg("secondaryButtonDown") boolean secondaryButtonDown,
-            @NamedArg("synthesized") boolean synthesized, @NamedArg("popupTrigger") boolean popupTrigger, @NamedArg("pickResult") PickResult pickResult,
-            @NamedArg("gestureSource") Object gestureSource) {
+                          @NamedArg("x") double x, @NamedArg("y") double y,
+                          @NamedArg("screenX") double screenX, @NamedArg("screenY") double screenY,
+                          @NamedArg("button") MouseButton button, @NamedArg("clickCount") int clickCount,
+                          @NamedArg("shiftDown") boolean shiftDown, @NamedArg("controlDown") boolean controlDown,
+                          @NamedArg("altDown") boolean altDown, @NamedArg("metaDown") boolean metaDown,
+                          @NamedArg("primaryButtonDown") boolean primaryButtonDown,
+                          @NamedArg("middleButtonDown") boolean middleButtonDown,
+                          @NamedArg("secondaryButtonDown") boolean secondaryButtonDown,
+                          @NamedArg("synthesized") boolean synthesized, @NamedArg("popupTrigger") boolean popupTrigger,
+                          @NamedArg("pickResult") PickResult pickResult,
+                          @NamedArg("gestureSource") Object gestureSource) {
         this(null, null, eventType, x, y, screenX, screenY, button, clickCount,
                 shiftDown, controlDown, altDown, metaDown, primaryButtonDown,
                 middleButtonDown, secondaryButtonDown, synthesized, popupTrigger,
```

Should I just paste in the new constructor signature for these?

kevinrushforth:
The CSR looks fine. I still need to do one final test pass for this.

Also, this needs a second reviewer. I've tagged @arapte so hopefully he'll be able to take a look.

arapte:
A general comment: For all modified files, please update the year in copyright header.
Added two specific comment for GlassWindow.cpp
yet to complete the review of all the files.

kevinrushforth:
> A general comment: For all modified files, please update the year in copyright header.

Normally, this is a good thing to remember, but since I am going to run my "copyright update" script tomorrow or Thursday over the entire repo to address [JDK-8214793](https://bugs.openjdk.java.net/browse/JDK-8214793), I recommend to skip this for this time.

modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
```diff
@@ -510,6 +510,9 @@ LRESULT GlassWindow::WindowProc(UINT msg, WPARAM wParam, LPARAM lParam)
         case WM_RBUTTONDBLCLK:
         case WM_MBUTTONUP:
         case WM_MBUTTONDBLCLK:
+        case WM_XBUTTONDOWN:
```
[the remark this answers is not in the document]
brcolow: That's a really good question and good catch. I think that it should indeed to moved so that the window will grab focus on BACK/FORWARD mouse button clicks. I'll make that change now.

tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
```diff
@@ -714,6 +807,15 @@ public static void exit() {
         Platform.exit();
     }
 
+    @After
+    public void cleanup() {
+        Util.runAndWait(() -> {
+            robot.mouseRelease(MouseButton.PRIMARY, MouseButton.SECONDARY, MouseButton.MIDDLE,
+                    MouseButton.BACK, MouseButton.FORWARD);
```
kevinrushforth: I was doing some final testing, and discovered that unconditionally releasing all buttons is causing problems on Windows. It seems that releasing a button over a window or the desktop background is causing an action to occur, which causes random menus to appear and the tests to fail. This suggests that a cleanup method that releases all buttons, even after tests that don't press any buttons, isn't what we want to do.

So it looks like I led you astray when I suggested doing this (I hadn't tested it on Windows after you made this change at my request). For now, I recommend either going back to the approach of releasing the buttons in those tests that do a corresponding mousePress, or else setting a variable indicating which button(s) need to be released by the `@After` clean-up method.

modules/javafx.graphics/src/main/native-glass/win/Robot.cpp
```diff
@@ -176,11 +182,9 @@ JNIEXPORT void JNICALL Java_com_sun_glass_ui_win_WinRobot__1mousePress
 
     // Support for extra buttons
     if (buttons & (1 << 3)) {
-        mouseInput.mi.dwFlags |= MOUSEEVENTF_XDOWN;
         mouseInput.mi.mouseData = XBUTTON1;
     }
     if (buttons & (1 << 4)) {
-        mouseInput.mi.dwFlags |= MOUSEEVENTF_XDOWN;
         mouseInput.mi.mouseData = XBUTTON2;
```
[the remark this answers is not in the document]
brcolow: Ah you are absolutely correct. Made that change. Good catch!

brcolow:
Pushed a fixup commit that should address both issues you raised.

tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
```diff
@@ -406,6 +410,7 @@ private static void testMouseAction(MouseAction mouseAction, MouseButton mouseBu
             switch (mouseAction) {
                 case PRESSED:
                     robot.mousePress(mouseButton);
+                    pressedButtons.add(mouseButton);
```
kevinrushforth: As long as this is the only place you press without releasing a mouse button, then this looks good. I'll test it.

kevinrushforth:
The latest changes look good to me. I'll test it on all platforms (probably tomorrow).

tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
brcolow: Besides the exception tests (which should trigger exceptions and therefore the mouse button will never be pressed) I believe it is.

modules/javafx.graphics/src/main/native-glass/win/Robot.cpp
```diff
@@ -176,13 +178,14 @@ JNIEXPORT void JNICALL Java_com_sun_glass_ui_win_WinRobot__1mousePress
...
-        mouseInput.mi.mouseData = XBUTTON1;
+        dwFlags |= MOUSEEVENTF_XDOWN;
+        mouseFlags |= XBUTTON1;
     }
     if (buttons & (1 << 4)) {
-        mouseInput.mi.dwFlags |= MOUSEEVENTF_XDOWN;
-        mouseInput.mi.mouseData = XBUTTON2;
+        dwFlags |= MOUSEEVENTF_XDOWN;
+        mouseFlags |= XBUTTON2;
     }
```
kevinrushforth: The above two blocks need to be moved before the setting of the mouseInput structure (else the dwFlags setting will not have any effect).

brcolow: Fixed.

kevinrushforth:
I eyeballed your fix and it looks like the same thing I did locally. I'll apply the patch and confirm.

kevinrushforth:
Once @arapte finishes reviewing, I'll merge it in and push to HG.

modules/javafx.swt/src/main/java/javafx/embed/swt/FXCanvas.java
```diff
@@ -689,6 +689,8 @@ private void sendMouseEventToFX(MouseEvent me, int embedMouseType) {
         boolean primaryBtnDown = (me.stateMask & SWT.BUTTON1) != 0;
         boolean middleBtnDown = (me.stateMask & SWT.BUTTON2) != 0;
         boolean secondaryBtnDown = (me.stateMask & SWT.BUTTON3) != 0;
+        boolean backBtnDown = (me.stateMask & SWT.BUTTON4) != 0;
```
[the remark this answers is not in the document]
brcolow: Okay, I re-added the check.

kevinrushforth:
Changes since my last review look fine to me. As soon as Ambarish approves the PR I'll merge it in and push to HG.

PR #174, opened 2018-08-27
JDK-8209966: Update minimum boot JDK to 11

kevinrushforth:
Fix for JBS issue [JDK-8209966](https://bugs.openjdk.java.net/browse/JDK-8209966).

Bump the minimum JDK version needed to build JavaFX to jdk-11+28.

Note that this will not change the `javac` `--source` or --target flags (which are controlled in build.gradle by the `'sourceCompatibility'` option), which will remain at 10.

This will allow us to implement the following follow-on issues:

1. Compile with `--source 11 --target 11`
2. Stop building the old javafx.swing implementation which relies on internal API (using qualified exports on the command line)
3. Stop redistributing the Microsoft VS2017 and Windows SDK DLLs (since they are present in JDK 11).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3795,10 +3795,8 @@ task javadoc(type: Javadoc, dependsOn: createMSPfile) {
         ]);
 
     doLast {
-        // FIXME: remove JDK 10 logic once minimum JDK is bumped to 11
-        def isBootJdk10 = jdkVersion.startsWith("10")
         projectsToDocument.each { p ->
-            def destDir = isBootJdk10 ? "$buildDir/javadoc" : "$buildDir/javadoc/${p.ext.moduleName}"
+            def destDir = "$buildDir/javadoc/${p.ext.moduleName}"
             copy {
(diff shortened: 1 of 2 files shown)


kevinrushforth:
NOTE: this is not yet ready to merge, since we want to wait until JDK 11 has released its final (GA) build before requiring it for building FX.

kevinrushforth:
This is now ready for review.

Pending the review, this will only be merged after JDK 11 is released, which is scheduled for Tuesday, 2018-09-25.

kevinrushforth:
@tiainen this might require you to update your build machine with the latest JDK 11, since EA builds of JDK 11 will be below the minimum.

tiainen:
Thanks for the heads up. We'll update the JDK on our build machines to the released JDK 11.

kevinrushforth:
Note: I updated the date of the commit, and pushed it, to trigger a CI test build, but the content itself is unchanged from the previous version.

PR #185, opened 2018-08-30
JDK-8210219: GlassClipboard.cpp fails to compile with newer versions of VS2017

### LINKED ISSUE #183: Compilation Failure (_HASH_SEED) on MSVC 14.15.26726 (Appveyor)
Noticed this while working on my extended mouse PR. It seems that the MSVC version on Appveyor was upgraded from 14.14.26428 (Visual Studio 2017 version 15.7) to 14.15.26726 (Visual Studio 2017 version 15.8) and as a result broke the native Windows compilation:

```
C:\projects\openjdk-jfx\modules\javafx.graphics\src\main\native-glass\win\GlassClipboard.cpp(86): error C2065: '_HASH_SEED': undeclared identifier
```

It appears that `GlassClipboard.cpp` is the only file to reference `_HASH_SEED` (and it uses it once). Not sure why that constant would be un-defined in a newer MSVC version, besides the fact that `hash_map` and `hash_set` were deprecated (at least, the one's from `std` were, maybe the one's from `stdext`, which we use, were not).  Is there a maximum C++ version spec that we can use when developing the native Windows code? Apparently C++11 has `unordered_map` and `unordered_set` equivalents - so that would be a nice long-term solution. A short-term solution may be just defining it ourselves:

`#define _HASH_SEED (size_t)0xdeadbeef`

but of course knowing exactly what changed in the MSVC upgrade to make it so `_HASH_SEED` is not defined would be a prerequisite to a solution.

kevinrushforth:
fixes #183 

[JDK-8210219](https://bugs.openjdk.java.net/browse/JDK-8210219)

Use our own "GLASS_HASH_SEED" constant instead of relying on an internal constant that can change out from under us (and apparently has).

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
@@ -29,6 +29,12 @@
 #include <hash_map>
 #include <hash_set>
 
+// The following is derived from _HASH_SEED, which is an internal constant from
+// include/xhash; since this is an internal constant we define our own rather
+// that relying on something that Microsoft could remove (and apparately has in
+// newer version of Visual C++)
+#define GLASS_HASH_SEED (size_t)0xdeadbeef
+
 #include "GlassApplication.h"
 #include "GlassClipboard.h"
 #include "GlassDnD.h"
@@ -83,7 +89,7 @@ size_t hash_value(const FORMATETC &fr)
     _Val += size_t(fr.ptd);
     _Val >>= 13;
     _Val += size_t(fr.tymed);
-    return _Val ^ _HASH_SEED;
+    return _Val ^ GLASS_HASH_SEED;
 }
 
 //NB! There are two suffixes for mimes:
--- a/modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
@@ -31,8 +31,8 @@
 
 // The following is derived from _HASH_SEED, which is an internal constant from
 // include/xhash; since this is an internal constant we define our own rather
-// that relying on something that Microsoft could remove (and apparately has in
-// newer version of Visual C++)
+// than relying on something that Microsoft could remove (and apparently has in
+// newer versions of Visual C++)
 #define GLASS_HASH_SEED (size_t)0xdeadbeef
 
 #include "GlassApplication.h"


modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
```diff
@@ -29,6 +29,12 @@
 #include <hash_map>
 #include <hash_set>
 
+// The following is derived from _HASH_SEED, which is an internal constant from
+// include/xhash; since this is an internal constant we define our own rather
+// that relying on something that Microsoft could remove (and apparately has in
```
brcolow: nitpick: typo: apparately -> apparently

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
```diff
@@ -29,6 +29,12 @@
...
+// newer version of Visual C++)
```
brcolow: super nitpick: typo: version -> versions

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
```diff
@@ -29,6 +29,12 @@
...
+#define GLASS_HASH_SEED (size_t)0xdeadbeef
```
brcolow: Is there any logic behind this seemingly weird value? And why is it defined in hex as opposed to, say, decimal 3735928559?

brcolow: Is it some magical number that makes hash bucket collisions less frequent or something?

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
brcolow: Good on you, found one I missed! 👍

kevinrushforth: fixed that as well as one other on this line:

that --> than

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
kevinrushforth: fixed

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
kevinrushforth: I'm not sure why they chose this number, but rather than trying to figure out whether we rely on its value, it seems a safer change to leave it the same as the existing constant.

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
kevinrushforth: That's what I get for committing it quickly without proofreading it first. Three typos in one comment block. :)

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
brcolow: All I can think of is it happens to have a good, seemingly random, distribution of 1s and 0s: "11011110101011011011111011101111". Anyways, off-topic, just kind of got me thinking.

PR #189, opened 2018-08-31
Run checkWhiteSpace without -v option

### LINKED ISSUE #188: Running checkWhiteSpace with -v obscures the failing files
The [.github/pr-check-white-space](https://github.com/javafxports/openjdk-jfx/blob/4783d5e922d1f75af2c175c243c7f7511458aaf2/.github/pr-check-whitespace.sh) script runs the `checkWhiteSpace` script with the `-v` option. This lists all files being checked along with a status of `OK` or a failure status. Since the files are not sorted by status, nor are the failing files listed separately at the end, if there is a failure it can be difficult to see which files are failing.

We should instead first list of the files being checked, and then run `checkWhiteSpace` without `-v`. This will make it easy to see the failures at a glance.

kevinrushforth:
fixes #188

In addition to separating out the listing of the files from the checking of the files, with the latter only showing the failures, I added the `--no-pager` option to `git diff` to allow it to run in the background without trying to access the terminal.

### DIFF
--- a/.github/pr-check-whitespace.sh
+++ b/.github/pr-check-whitespace.sh
@@ -1,9 +1,5 @@
 #! /bin/bash
 
-echo ""
-echo "Checking for white space errors"
-echo ""
-
 if [ -z "$TRAVIS_BRANCH" ]; then
     echo "TRAVIS_BRANCH is undefined, skipping white-space check"
     exit 0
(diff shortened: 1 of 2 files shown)


kevinrushforth:
@brcolow can you take a look as well?

kevinrushforth:
Note that this PR only touches github-only files, so there is no need for a JBS issue.

PR #198, opened 2018-09-03
only publish Swing artifacts if we are going to build Swing

### LINKED ISSUE #148: build fails for platforms that don't include Swing
When Swing is not included in a project (e.g. for embedded), building fails as there is no task for publishing the Swing artifacts.

johanvos:
This fixed #148

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2437,7 +2437,9 @@ project(":swing") {
 
     compileJava.options.compilerArgs.addAll(qualExportsSwing)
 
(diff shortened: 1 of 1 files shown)


johanvos:
JBS bug at https://bugs.openjdk.java.net/browse/JDK-8210359

kevinrushforth:
Changing label to Bug, since this causes the build to break on platforms that don't build Swing.

Issue #175, opened 2018-08-27
[XWayland] libX11 SIGSEV


mwkroening:
Tinkering around with https://github.com/johanvos/javafx11samples I found that `./gradlew run` fails when run in the GNOME Wayland session on Arch Linux. I do not experience any problems when running the command in the X11 session or using the OracleJDK 10 with integrated JavaFX.
I tested it with the OpenJDK 10 and the OracleJDK 11 in combination with all openjfx versions from The Central Repository (11-ea+19, 11-ea+23 and 11-ea+24). All fail with the following message:
```
[...]
> Task :run FAILED

(java:3817): Gdk-CRITICAL **: 23:30:54.099: gdk_x11_display_set_window_scale: assertion 'GDK_IS_X11_DISPLAY (display)' failed
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007fdd058db7db, pid=3817, tid=3849
#
# JRE version: OpenJDK Runtime Environment (10.0.2+13) (build 10.0.2+13)
# Java VM: OpenJDK 64-Bit Server VM (10.0.2+13, mixed mode, tiered, compressed oops, g1 gc, linux-amd64)
# Problematic frame:
# C  [libX11.so.6+0x2b7db]
[...]
```
Here is the full error report: [hs_err_pid3817.log](https://github.com/javafxports/openjdk-jfx/files/2325721/hs_err_pid3817.log)

mwkroening:
I am able to reproduce this behavior on Ubuntu 18.04 (openjdk 10.0.2) and Fedora 28 (openjdk 11-ea+22) in virtual machines.
The step to reproduce is just running `./gradlew run` in https://github.com/johanvos/javafx11samples/tree/master/topics/javafx3d/hello3d .

brcolow:
[...]

PR #199, opened 2018-09-05
Set forcedGtkVersion to 2 in case XWayland is used.

johanvos:
See https://github.com/javafxports/openjdk-jfx/issues/175


johanvos:
Since there is a simple workaround (-Djdk.gtk.version=2) and the problem is wayland-related, it is a bad idea to change the default behavior in JavaFX.

PR #200, opened 2018-09-05
JDK-8210218: WebKit build fails with newer versions of VS 2017

ararunprasad:
There is a problem in the newer version(VC 14.15.26726) MSVC compiler which causes multiple symbol definition error when using lambda expression type in a function pointer argument.

Problem is already reported to Visual Studio team and it is been discussed over [here](https://developercommunity.visualstudio.com/content/problem/325062/msvs-151526726-throws-multiple-definition-error-wh.html).

The proposed workaround is to use constexpr.

### DIFF
--- a/modules/javafx.web/src/main/native/Source/WTF/wtf/Variant.h
+++ b/modules/javafx.web/src/main/native/Source/WTF/wtf/Variant.h
@@ -1868,13 +1868,13 @@ struct __visitor_table{
         return __visitor(get<_Type>(__v));
     }
 
-    static const __func_type __trampoline[sizeof...(_Types)];
+    static constexpr __func_type __trampoline[sizeof...(_Types)]={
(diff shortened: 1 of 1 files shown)


ararunprasad:
Though the problem happens only with MSVC, the proposed workaround uses `constexpr` for all platforms.

tiainen:
I can confirm that the build completes successfully now when compiling with version 14.15.26726 of the Visual Studio tools.

kevinrushforth:
Tracked in JBS as [JDK-8210218](https://bugs.openjdk.java.net/browse/JDK-8210218)

johanvos:
I'm testing on mac/linux to make sure the constexpr don't cause regression there.

kevinrushforth:
I also tested on Mac and Linux as well as on Windows with the existing 14.10.25017 compiler. I found no problems.

PR #201, opened 2018-09-13
JDK-8210691: Deliver release notes for JavaFX 11

kevinrushforth:
Here is the final draft of the JavaFX 11 release notes. The content is ready for review, but we still need to decide where to host it before it can be merged.

### DIFF
--- a/doc-files/release-notes-11.md
+++ b/doc-files/release-notes-11.md
@@ -102,7 +102,6 @@ Issue key|Summary|Subcomponent
 [JDK-8201285](https://bugs.openjdk.java.net/browse/JDK-8201285)|DateCell text color are not updated correctly when DateCell with disable = true is reused|controls
 [JDK-8208610](https://bugs.openjdk.java.net/browse/JDK-8208610)|Incorrect check for calling class in FXMLLoader::getDefaultClassLoader|fxml
 [JDK-8129582](https://bugs.openjdk.java.net/browse/JDK-8129582)|Controls slow considerably when displaying RTL-languages text on Linux|graphics
-[JDK-8170026](https://bugs.openjdk.java.net/browse/JDK-8170026)|Some tests in RegionBackgroundFillUITest fail with fractional window scale|graphics
 [JDK-8195801](https://bugs.openjdk.java.net/browse/JDK-8195801)|Replace jdk.internal.misc.Unsafe with sun.misc.Unsafe in MarlinFX|graphics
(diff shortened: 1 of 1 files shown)


kevinrushforth:
The easiest way to review it is to look at the new [release-notes-11.md](https://github.com/kevinrushforth/openjdk-jfx/blob/release-notes/doc-files/release-notes-11.md) fille.

kevinrushforth:
Question for reviewers: should the table of fixed bugs/enhancements be sorted by sub-component?

kevinrushforth:
Regardless of where we host this, we want to put it under version control in the openjfx/11-dev repo. I filed [JDK-8210691](https://bugs.openjdk.java.net/browse/JDK-8210691) to track this.

tiainen:
* I'm in favour of sorting the table by subcomponent.
* Regardless of where the release notes will be hosted, having them in the code repository is indeed a must. It allows anyone to link directly to the release notes.

nlisker:
Sorted table looks good.

Are the listed changes user-facing only? On one hand, I don't see [JDK-8196198 Remove obsolete apps/experiments](https://bugs.openjdk.java.net/browse/JDK-8196198https://bugs.openjdk.java.net/browse/JDK-8196198), on the other hand, I see [JDK-8196130 Eclipse configuration files need to be updated](https://bugs.openjdk.java.net/browse/JDK-8196130).

kevinrushforth:
> Are the listed changes user-facing only?

That was roughly the intent. I excluded the build and samples subcategories as well as bugs labeled with testbug, noreg-self, or noreg-demo.

Perhaps we could start using the "Task" issuetype for IDE-only changes, since they don't touch anything that could affect the build. In any case, I'll manually exclude that one.

kevinrushforth:
I changed [JDK-8196130](https://bugs.openjdk.java.net/browse/JDK-8196130) to a Task. I also found one more bug that should be excluded, since it is a test bug, but was missing the label (I added it).

[JDK-8170026](https://bugs.openjdk.java.net/browse/JDK-8170026) - Some tests in RegionBackgroundFillUITest fail with fractional window scale

The above two bugs no longer appear in my query. I will remove both of them from the table.

kevinrushforth:
Having nothing to do with this change... I see that the Travis CI build failed. This failure is due to a network error affecting jdk.java.net and openjdk.java.net. It seems the only reason the Appveyor build didn't fail is that an error in the download doesn't fail the build, and the build machine in question happened to have a jdk11 installed from a previous build.

In any case, "they" know about the network problem and are working on it.

PR #202, opened 2018-09-15
JDK-8210791: Update release notes for JavaFX 11

kevinrushforth:
[JDK-8210791](https://bugs.openjdk.java.net/browse/JDK-8210791)

A [new Swing interop bug](https://bugs.openjdk.java.net/browse/JDK-8210759) was filed after the JavaFX 11 release notes were created. This bug is a new issue in FX 11 with a convenient workaround, and should be listed in the "Known Issues" section.

I also updated the release note for the other two Swing interop Known Issues for consistency and to fix a couple wording problems.

We might take this opportunity to address any other small updates to the release notes that are found during this review.

### DIFF
--- a/doc-files/release-notes-11.md
+++ b/doc-files/release-notes-11.md
@@ -59,9 +59,9 @@ Workaround: explicitly force GTK 2 by passing the following system property on t
 
 See [JDK-8210411](https://bugs.openjdk.java.net/browse/JDK-8210411) for more information.
 
-### FX / Swing interop requires qualified exports when run with JDK 10
+### Swing interop requires qualified exports when run with JDK 10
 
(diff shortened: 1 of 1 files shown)


kevinrushforth:
If you want to review the entire document, you can click on the [release-notes-11.md](https://github.com/kevinrushforth/openjdk-jfx/blob/release-notes/doc-files/release-notes-11.md) file.

kevinrushforth:
I also agree with both points made by Phil and will send an update shortly.

@johanvos would you prefer the example of running a JavaFX application be inline or a pointer to an external page (or maybe both: a simple example inline with a pointer for more information)?

kevinrushforth:
I pushed a change that should address both of Phil's points. The example of running a JavaFX app was done with a pointer to the JavaFX 11 intro page (and no inline example).

johanvos:
+1

Issue #205, opened 2018-09-18
JDK-8210386: Clipping problems with complex affine transforms


bourgesl:
See JBS: https://bugs.openjdk.java.net/browse/JDK-8210386

bourgesl:
Fixed in jfx-dev

PR #206, opened 2018-09-18
Fix bug JDK-8210386 (merged with Marlin 0.9.3)

bourgesl:
Please review this PR for #205 that consist in merging latest changes from Marlin 0.9.3.

See also the OpenJDK 12 patch under review: 
http://mail.openjdk.java.net/pipermail/2d-dev/2018-September/009443.html

laurent

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/marlin/DRendererContext.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/marlin/DRendererContext.java
@@ -79,6 +79,8 @@ public final class DRendererContext extends ReentrantContext implements MarlinCo
     boolean closedPath = false;
     // clip rectangle (ymin, ymax, xmin, xmax):
     public final double[] clipRect = new double[4];
+    // clip inverse scale (mean) to adjust length checks
+    public double clipInvScale = 0.0d;
     // CurveBasicMonotonizer instance
     public final CurveBasicMonotonizer monotonizer;
     // CurveClipSplitter instance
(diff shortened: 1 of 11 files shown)


bourgesl:
FYI I adapted the ScaleClipTest class to use JavaFX API:
It fails on openjfx without the fix and passes with the patch.

kevinrushforth:
Don't forget to send a review request to openjfx-dev, in which case there will be no need to post a webrev. This PR can serve as the review to get it into openjfx/jfx-dev/rt

kevinrushforth:
Fix for [JDK-8210386](https://bugs.openjdk.java.net/browse/JDK-8210386)

bourgesl:
Phil, could you have a look or any other reviewer ?

johanvos:
I did a quick review and this patch makes sense -- but I didn't do before/after tests.

kevinrushforth:
@johanvos or @prrace can one of you approve so we have a second reviewer?

bourgesl:
Phil, you are right.
FX Color class is not equivalent to awt Color and it was disappointing that no method exist to translate Color <=> int (rgba), probably for good reasons (colorspace srgb vs rgb ?).

Kevin, could you create a RFE to enhance FX Color API ?

I will push asap this patch as is, in jfx-dev directly. Kevin, what is the process to ask for a jfx11u backport ?

kevinrushforth:
> Kevin, could you create a RFE to enhance FX Color API ?

You can file it if you want. I suspect that this isn't something we would work on in isolation, rather more likely as part of an "add image operations to JavaFX" RFE (which I thought we already had, but I can't find). If there is enough interest by someone to do this separately, in a way that makes sense, that could be another option.

> I will push asap this patch as is, in jfx-dev directly.

OK (in that case, no need to merge this PR into develop).

> Kevin, what is the process to ask for a jfx11u backport ?

Still pending, as we haven't nailed down how this will be done. Johan and I have been discussing the logistics of this, and will come up with something in the next week or so.

PR #208, opened 2018-09-19
JDK-8210093: Build FX class files with -source 11 and -target 11

kevinrushforth:
Fix for JBS issue [JDK-8210093](https://bugs.openjdk.java.net/browse/JDK-8210093).

This will build the classes in the javafx.* modules with `-source 11` and `-target 11`, meaning that we can start to use JDK 11 language features. It will also cause a fast-fail if a down-rev JDK is used.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1765,7 +1765,7 @@ allprojects {
     // All of our projects are java projects
 
     apply plugin: "java"
-    sourceCompatibility = 10
+    sourceCompatibility = 11
 
     // By default all of our projects require junit for testing so we can just
     // setup this dependency here.


kevinrushforth:
NOTE: this is not yet ready to merge. It is blocked by [JDK-8209966](https://bugs.openjdk.java.net/browse/JDK-8209966), PR #174, which must be integrated first.

kevinrushforth:
This is now ready for review.

Pending the review, this will be merged after JDK-8209966, PR #174, is merged.

kevinrushforth:
Note: I fixed the commit message (`-source` and `-target` rather than `--source` and `--target`), but the content itself is unchanged from the previous version.

PR #215, opened 2018-09-24
JDK-8207369 Add more robot tests.

brcolow:
(Partially? at least a step in the right direction...) resolves robot API follow-up issue [JDK-8207369](https://bugs.openjdk.java.net/browse/JDK-8207369).

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
@@ -28,9 +28,11 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
+import com.sun.javafx.PlatformUtil;
 import javafx.application.Application;
 import javafx.application.Platform;
 import javafx.beans.InvalidationListener;
+import javafx.event.EventHandler;
 import javafx.geometry.Insets;
 import javafx.geometry.Point2D;
 import javafx.scene.Scene;
@@ -39,6 +41,7 @@ import javafx.scene.control.TextField;
 import javafx.scene.image.WritableImage;
 import javafx.scene.input.KeyCode;
 import javafx.scene.input.MouseButton;
+import javafx.scene.input.MouseEvent;
 import javafx.scene.layout.Background;
 import javafx.scene.layout.BackgroundFill;
 import javafx.scene.layout.HBox;
@@ -52,9 +55,10 @@ import javafx.stage.StageStyle;
 import javafx.stage.WindowEvent;
 
 import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.BeforeClass;
 import org.junit.Test;
-import org.junit.Assert;
 import junit.framework.AssertionFailedError;
 import test.util.Util;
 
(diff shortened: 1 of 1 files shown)


tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
```diff
@@ -133,26 +258,135 @@ public void testMouseMove() {
...
+
+        @Override
+        public void handle(MouseEvent event) {
+            String expectedText = mouseAction + " " + mouseButton;
+            if (event.getButton() == mouseButton) {
+                button.setText(expectedText);
+                onClickLatch.countDown();
+            } else {
+                button.setText(String.format(mouseAction + " wrong button (expected \"%s\" but got \"%s\")",
+                        mouseButton, event.getButton()));
```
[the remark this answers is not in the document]
brcolow: Good catch. I will investigate the many failures.

brcolow:
I think the failures (besides the mouse scroll amount which seems to be hardcoded to a value of (-)40 on Linux) are happening because you can't release "in a vacuum" without a preceding press. I will revert those to testing mouse clicks (with press and release).

brcolow:
The mouse scroll wheel failure is kind of interesting. Basically the GtkRobot sends multiple fake mouse buttons presses (4 for scroll up, 5 for scroll down), one for each int of the `wheelAmt`. Therefore I think what I can do is add a sort of timed collection to accumulate all scroll events sent to the `Button` and then add them up. I may still need to mess with the multiplier though on a system-dependent basis which is going to be ugly. The reason for this is that the JavaFX event handler will get 5 scroll events each with a scroll amount of (+/-) 40 for a total of 200. Then one would need to divide by the multiplier to match the passed-in `wheelAmt`.

brcolow:
For some reason `event.getTotalDeltaY()` reports 0. But I was able to hack something together that passes on Windows and Linux. Still need to test macOS.

brcolow:
I finally was able to test and fix the new tests on macOS. I have now confirmed that the new tests pass on all 3 operating systems. I am not really happy with the approach I took in order to get a unified scroll amount between platforms but wasn't sure of an alternative approach.

brcolow:
The new drag tests pass locally for me on Windows but seem to fail in a Linux virtual machine but I can't tell if that's because it's really slow or because there is an actual problem. Do you have a local Linux install you could test them on?

kevinrushforth:
I can test them later this morning when I get into the office.

kevinrushforth:
The tests all PASS on my Linux system (a physical machine running Ubuntu 18.04). I'll run on Windows and Mac and then finish my review.

I don't have a good answer to your question about the platform differences regarding the scrolling multiplier (-1 on Mac versus 40 on Windows and Linux). The negative versus positive might be related to the "natural scrolling order" on Mac, which is intentionally backwards by default from the other platforms. However, a difference of 40x in magnitude points to something wrong in the implementation.

tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
```diff
@@ -167,23 +372,124 @@ public void testMouseClick() {
...
+    }
+
+    private static void testMouseWheel(int amount) {
         CountDownLatch onScrollLatch = new CountDownLatch(1);
         CountDownLatch setSceneLatch = new CountDownLatch(1);
         Button button = new Button("Scroll me");
         InvalidationListener invalidationListener = observable -> setSceneLatch.countDown();
+        int[] totalScroll = new int[]{0};
+        long[] firstScrollMillis = new long[]{0};
+        int scrollMultiplier = System.getProperty("os.name").toLowerCase().startsWith("mac") ? -1 : 40;
```
kevinrushforth: Can you add a comment here? This suggests that there might be an issue with scrolling in the Mac robot implementation. We might want to file a follow-up issue to investigate.

brcolow:
Added a commit which addresses both issues.

tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
```diff
@@ -474,10 +474,12 @@ private static void testMouseWheel(int amount) {
         InvalidationListener invalidationListener = observable -> setSceneLatch.countDown();
         int[] totalScroll = new int[]{0};
         long[] firstScrollMillis = new long[]{0};
+        // The scroll wheel amount is multiplied by 40 on Linux and Windows, but not on macOS. The
+        // directions are also reversed.
```
kevinrushforth: I filed JDK-8214580 to investigate this. Can you add to this comment that the difference in behavior is unexpected and might be a bug? Something like:

```
// This difference in behavior is unexplained and might be a bug. See JDK-8214580.
```

tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
```diff
@@ -474,10 +474,12 @@ private static void testMouseWheel(int amount) {
...
         int scrollMultiplier = System.getProperty("os.name").toLowerCase().startsWith("mac") ? -1 : 40;
         Util.runAndWait(() -> {
             button.setOnScroll(event -> {
-                totalScroll[0] += event.getDeltaY();
+                totalScroll[0] += event.getDeltaY() * Screen.getPrimary().getOutputScaleY();
```
kevinrushforth: Looks good. I'll test it on Monday.

tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
```diff
@@ -474,10 +474,13 @@ private static void testMouseWheel(int amount) {
...
         int[] totalScroll = new int[]{0};
         long[] firstScrollMillis = new long[]{0};
+        // The scroll wheel amount is multiplied by 40 on Linux and Windows, but not on macOS. The
+        // directions are also reversed. This difference in behavior is unexplained and might be a bug.
+        // See JDK-8214580.
         int scrollMultiplier = System.getProperty("os.name").toLowerCase().startsWith("mac") ? -1 : 40;
         Util.runAndWait(() -> {
             button.setOnScroll(event -> {
-                totalScroll[0] += event.getDeltaY();
+                totalScroll[0] += event.getDeltaY() * Screen.getPrimary().getOutputScaleY();
```
kevinrushforth: Multiplying by the output scale works correctly on Windows. It doesn't work on a Mac with a retina display, though. The test now fails for me. This suggests another platform difference between how Robot::mouseWheel works on Mac versus the other platforms.

brcolow: Do you think we should skip the test on Mac for now and un-skip it as part of JDK-8214580?

brcolow: Okay, I added a new commit and updated the bug.

kevinrushforth:
That looks good. One quick thought: for consistency, the following could also use `PlatformUtil.isMac()` now that the test imports it:

```
    int scrollMultiplier = System.getProperty("os.name").toLowerCase().startsWith("mac") ? -1 : 40;
```

I'll do a quick final test on Mac and Windows and then I think we can get this in.

kevinrushforth:
Btw, this would have been easier for me to review without the squash / force-push

brcolow:
Ah, sorry about that. Will keep that in mind going forward.

PR #216, opened 2018-09-24
Can access getRootHeader() and getColumnHeaderFor() outside the package

purnachand-jaddu:


kevinrushforth:
Submitting a pull request with no additional information or context is not the way to suggest an API change. Any such change needs to be discussed on the openjfx-dev mailing list. In this particular case, this overlaps with what is being worked on in issue #151 / PR #163. I suggest you add comments to #163 with an explanation as to why you think this particular API needs to be public.

kevinrushforth:
I also refer you to the guidelines in [CONTRIBUTING.md](https://github.com/javafxports/openjdk-jfx/blob/develop/.github/CONTRIBUTING.md).

PR #219, opened 2018-09-25
[WIP] test build to check whether we can download the JDK 11 GA build

kevinrushforth:
Test build...do not merge. I am testing that the logic to download the JDK 11 build from http://jdk.java.net/11/ still works after the GA release.


kevinrushforth:
Download still works, so closing this experimental PR.

PR #223, opened 2018-09-27
Allow to launch an application using a custom preloader

sbordes:
Useful since LauncherImpl is no more visible due to JPMS
and javafx modularization.

The JRebirth Application Framework was using the underlying LauncherImpl class but it's broken since Java 9


eugener:
This PR has to be retargeted to a `dev` branch

PR #224, opened 2018-09-27
Allow to launch an application using a custom preloader

sbordes:
Useful since LauncherImpl is no more visible due to JPMS
and javafx modularization.

The JRebirth Application Framework was using the underlying LauncherImpl class but it's broken since Java 9

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/application/Application.java
+++ b/modules/javafx.graphics/src/main/java/javafx/application/Application.java
@@ -184,6 +184,53 @@ public abstract class Application {
     public static final String STYLESHEET_MODENA = "MODENA";
 
     /**
+     * Launch a standalone application with its custom preloader.
+     * This method is typically called
+     * from the main method(). It must not be called more than once or an
+     * exception will be thrown.
+     *
+     * <p>
+     * The launch method does not return until the application has exited,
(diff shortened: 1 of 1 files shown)


kevinrushforth:
@sbordes welcome to openjfx. In order for us to evaluate your proposed enhancement, you will need to sign the Oracle Contributor Agreement. The [CONTRIBUTING.md](https://github.com/javafxports/openjdk-jfx/blob/develop/.github/CONTRIBUTING.md) doc has a pointer to this. Also, we will need an Enhancement request to be filed in JBS. The CONTRIBUTING doc also has instructions for that.

Since this is a new API and not a bug fix, you should send an email to openjfx-dev@openjdk.java.net indicating that you are proposing this enhancement with a pointer to the pull request.

A few other things will eventually be needed once there is general agreement that this is a good feature to take into OpenJFX.

1. API documentation for the new method(s), including `@since 12` tags if you propose to get it into OpenJFX 12. I see that you have a good start on this already.
2. Automated tests. You can look at the MainLauncher test underneath tests/system/
3. A [CSR](https://wiki.openjdk.java.net/display/csr/Main) will need to be filed. This usually happens after there is agreement on the API and documentation. Whoever ends up sponsoring your change can file the CSR on your behalf.

The final code review can proceed in parallel with the CSR review.

PR #233, opened 2018-10-02
Testing interop with Apache HTML/Java @JavaScriptBody annotation

dukescript:
Apache [HTML/Java API](https://dukescript.com/javadoc/html+java/1.5.1/) is using JavaFX `WebView` to its limits. It heavily depends on Java and JavaScript virtual machines interactions including proper garbage collecting behavior.

There have been regressions in this kind of behavior in the past. As such we propose to include the `@JavaScriptBody` TCK into the build of JavaFX itself. The [HTML/Java TCK](https://dukescript.com/javadoc/html+java/1.5.1/org/netbeans/html/json/tck/package-summary.html) is portable set of tests. Enough to instantiate a `WebView`, init it and execute appropriate test method. Which is exactly what the `WebViewInteropJavaScriptBodyTest` does in this pull request.

Please consider improving future compatibility of JavaFX WebView by including this PR in the JavaFX repository itself and thus executing the TCK on every future change.


kevinrushforth:
Please read [CONTRIBUTING.md](CONTRIBUTING.md) for the process for submitting pull requests to us.

In this case, your proposed contribution makes use of a new third-party library, which means that additional approval will be required (you can't just submit a patch that uses code that you don't own). In general, integrating an external test suite is not something we would do. This seems more suitable for a separate test project.

brcolow:
Creating a separate project that runs this test suite and uploads the results somewhere would be very useful for fixing WebView bugs. I am in support of that proposal and also think it would be the best way to integrate the intent behind this PR into the JavaFX build, specifically because building `webkit` as part of the automated CI builds is currently not possible, and so these tests would never be automatically run - they would have to be manually run and each developer would have to do this themselves. The results could be uploaded to a Github Pages repository and we could link to it in the documentation for this Github repository, for example.

dukescript:
Please accept my apology for providing non-compilable first version. The error was caused by late additional manual rename: cbad1cd fixes that.

Re. "3rd party library" - yes, I understand why accepting new 3rd party library dependency into production code would be problematic. However please note that this PR only adds `testCompile` dependencies. As a result users of `org.openjfx:javafx-web` artefact aren't going to be affected. It might be easier to accept new test only dependency.

> building webkit as part of the automated CI builds is currently not possible

Opps. The main reason for this PR is to include the check in CI. If automated CI builds skip `WebKit` tests, and developers have to run them manually, the value of the additional test decreases (but still they may catch regressions sooner than they get in).

Re. "where are the actual tests" - they are part of [Apache NetBeans HTML/Java](https://github.com/apache/incubator-netbeans-html4j/) repository. There is a `json-tck` module in the repository. All 63 currently executed tests come from [net.java.html.js.tests](https://github.com/apache/incubator-netbeans-html4j/tree/master/json-tck/src/main/java/net/java/html/js/tests) package.

Re. "separate module & JavaFX build infrastructure & uploaded to a Github Pages" - I can donate the HTML/Java TCK tests (unless I make a typo) and even add few more, but somebody more knowledgeable of your project infrastructure would have to setup the module, build and upload.

brcolow:
Yeah, ideally webkit would be tested on CI but so far our efforts in doing so have not borne fruit. I am not any kind of authority on these matters - only wanted to add my two cents and ideas. With feedback from the maintainers we can work together in finding the right solution. I am excited by any work in the area of JavaFX's tests/compliance/CI so this piqued my interest and I think it's the seed of a great idea.

kevinrushforth:
Closing this PR since it is not something we want to take into the main project, at least not at this time. It seems more suitable for a separate project.

Issue #229, opened 2018-09-30
Reduce color fringes in FreeType subpixel rendering


jgneff:
## Report Classification

**Issue Type:** Bug  
**Component:** JavaFX  
**Subcomponent:** *graphics:* JavaFX Graphics  
**Operating System:** Ubuntu  
**Java Release:** 11  
**Regression:** none  
**Frequency:** Always

### Synopsis

Reduce color fringes in FreeType subpixel rendering

### Description

The text rendered on Linux by JavaFX has severe color fringes because it fails to set the FreeType LCD filter.

The graphics module attempts to set the default LCD filter in `FTFactory.java`, shown below. Setting the filter is crucial. If the function returns an error, JavaFX disables subpixel rendering and reverts to grayscale anti-aliasing.

**javafx.graphics/src/main/java/com/sun/javafx/font/freetype/FTFactory.java**

```Java
/* This implementation only supports LCD if freetype has support. */
error = OSFreetype.FT_Library_SetLcdFilter(library, OSFreetype.FT_LCD_FILTER_DEFAULT);
LCD_SUPPORT = error == 0;
```

Yet its implementation in `freetype.c`, shown below, has a deceptive bug causing Linux users to see unfiltered subpixel rendering, while those working on the JavaFX FreeType support are unable to see the problem on their development workstations. On Linux systems with a default installation, the function fails silently and returns success (0) because the FreeType library is not found.

**javafx.graphics/src/main/native-font/freetype.c**

```C
#define LIB_FREETYPE "libfreetype.so"
JNIEXPORT jint JNICALL OS_NATIVE(FT_1Library_1SetLcdFilter)
    (JNIEnv *env, jclass that, jlong arg0, jint arg1)
{
//  return (jint)FT_Library_SetLcdFilter((FT_Library)arg0, (FT_LcdFilter)arg1);
    static void *fp = NULL;
    if (!fp) {
        void* handle = dlopen(LIB_FREETYPE, RTLD_LAZY);
        if (handle) fp = dlsym(handle, "FT_Library_SetLcdFilter");
    }
    jint rc = 0;
    if (fp) {
        rc = (jint)((jint (*)(jlong, jint))fp)(arg0, arg1);
    }
    return rc;
}
```

There are two problems, with two alternative solutions:

[...]

PR #235, opened 2018-10-02
JDK-8188810: Fonts are blurry on Ubuntu 16.04 and Debian 9

jgneff:
Reduce color fringes in FreeType subpixel rendering with a direct call
to the function FT_Library_SetLcdFilter, available since FreeType 2.3.0.
Note that the runtime reference to the shared library is the versioned
file name libfreetype.so.6.

Fixes javafxports/openjdk-jfx#229


kevinrushforth:
Great, thank you. Please also send a review request to openjfx-dev with a pointer to the JBS issue and this pull request.

PR #264, opened 2018-11-04
Adding back the flight recorder pulse logger

thegreystone:
1. Sorry about the formatting changes in the PrintLogger -  just habitually formatted the code after making the relevant changes.

2. These changes are not compatible with the attribute names in JDK 8. It is possible to make them compatible by renaming the fields, but in JDK 8 the attribute IDs and the attribute names differed. I think the current metadata is better going forward.

3. Because of 2, we would need to decide whether to drop backwards compatibility to Oracle JDK 8 in the JMC JavaFX plug-in or do some extra coding. Since the JavaFX events were dropped for some time (after 8?) and are just now re-introduced, I would suggest only supporting the new stuff in the next version of JMC.


thegreystone:
Trying a new pull request.

Issue #38, opened 2018-03-24
scroll events finish with wrong delta's on mac: JDK-8183399


johanvos:
see https://bugs.openjdk.java.net/browse/JDK-8183399

johanvos:
I could reproduce it, and found out why this happens.
The relevant events are sent to GlassViewDelegate.m sendJavaMouseEvent:(NSEvent *)theEvent

That method then uses the [theEvent deltaX] and [theEvent deltaY] and pass those to the Java layer unless they are 0.
I could observe these values, in the end, become 0 a number of times, and only after e.g. 3 0 values, the cumulative value was high enough to be passed.

I changed the [theEvent deltaX] to [theEvent scrollingDeltaX] and similar for Y, and that always return non-0 values, that gradually decrease as one should expect.

johanvos:
The following patch could fix this:
```
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
@@ -425,8 +425,9 @@ - (void)sendJavaMouseEvent:(NSEvent *)theEvent
     jdouble rotationY = 0.0;
     if (type == com_sun_glass_events_MouseEvent_WHEEL)
     {
-        rotationX = (jdouble)[theEvent deltaX];
-        rotationY = (jdouble)[theEvent deltaY];
+        rotationX = (jdouble)[theEvent scrollingDeltaX];
+        rotationY = (jdouble)[theEvent scrollingDeltaY];
+
 
         //XXX: check for equality for doubles???
         if (rotationX == 0.0 && rotationY == 0.0)

```

dlemmermann:
This only seems to fix it for the mouse wheel. How about gestures?

dlemmermann:
And also only on Mac. Aren‘t other operating systems affected?

johanvos:
It is not mouse wheel specific (I tested it with gestures). It's a mac-only issue indeed. It is in the native-glass layer, and the wrong values are passed to the java layer.

johanvos:
The values of scrollingDeltaX/Y are 10 times higher than the ones of deltaX/Y.
The native events are sent to the Java layer (MacGestureSupport.scrollGesturePerformed) and then forwarded to e.g. GestureSupport.handleScrollingPerformed where a multiplier (hard-coded to 10) is passed as well.
[...]

PR #274, opened 2018-11-09
Fix scrolling on mac [JDK-8183399]

johanvos:
This fixes https://bugs.openjdk.java.net/browse/JDK-8183399 (https://github.com/javafxports/openjdk-jfx/issues/38)

The test program that prints the scrollEvent.getDeltaY() has now different values that keep decreasing: Old values:
```
-31.00006103515625
-28.000030517578125
-26.00006103515625
-24.000091552734375
-22.0001220703125
-20.0
-18.000030517578125
-16.00006103515625
-15.0
-14.000091552734375
-13.000030517578125
-12.0001220703125
-11.00006103515625
-9.000091552734375
-9.000091552734375
-8.000030517578125
-8.000030517578125
-7.0001220703125
-6.00006103515625
-6.00006103515625
-5.0
-5.0
-15.0
-12.0001220703125

```
Note the jump from -5 to -15 at the end.

New values:
```
140.0
130.0
120.0
110.0
100.0
90.0
80.0
80.0
70.0
70.0
60.0
60.0
50.0
50.0
50.0
40.0
40.0
40.0
30.0
30.0
30.0
20.0
20.0
20.0
20.0
20.0
10.0
10.0
10.0
10.0
10.0
10.0
10.0
10.0
```

Note that these values are higher in absolute value. I doubt this is a problem, as the delta's should only be used to give a relative indication?

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
@@ -425,8 +425,8 @@ static jint getSwipeDirFromEvent(NSEvent *theEvent)
     jdouble rotationY = 0.0;
     if (type == com_sun_glass_events_MouseEvent_WHEEL)
     {
-        rotationX = (jdouble)[theEvent deltaX];
-        rotationY = (jdouble)[theEvent deltaY];
+        rotationX = (jdouble)[theEvent scrollingDeltaX];
+        rotationY = (jdouble)[theEvent scrollingDeltaY];
 
         //XXX: check for equality for doubles???
         if (rotationX == 0.0 && rotationY == 0.0)


dlemmermann:
I am not sure how the values were originally designed, but I definitely use them as "pixels". In FlexGanttFX for example I use it to calculate the time to scroll to (calculateTimeToScrollTo(double)):

```
    node.addEventHandler(ScrollEvent.SCROLL, evt -> {
            double deltaX = evt.getDeltaX();
            if (deltaX != 0) {
                TimelineModel<?> timelineModel = getTimeline().getModel();
                Instant time = timelineModel.calculateTimeForLocation(-deltaX);
                timelineModel.setStartTime(time);
            }
        });
```

So this fix would break that.

dlemmermann:
I could of course adjust the value in my code, but that would mean that I need to adjust the delta values depending on the JDK that is being used. And since FlexGanttFX is a framework my customers could use it on any Java version they like. I do not have control over that.

kevinrushforth:
I think it is very likely that these are being treated as pixels by other applications as well. In #38 you mentioned a similar AWT bug, [JDK-8173876](https://bugs.openjdk.java.net/browse/JDK-8173876). In looking at the their solution, they did a divide by 10 when using `scrollingDeltaX/Y`.

```
+    jdouble deltaX = [event deltaX];
+    jdouble deltaY = [event deltaY];
+    if ([AWTToolkit hasPreciseScrollingDeltas: event]) {
+        deltaX = [event scrollingDeltaX] * 0.1;
+        deltaY = [event scrollingDeltaY] * 0.1;
+    }
```

johanvos:
The divide by 10 would bring it in alignment with what was before, but it looks pretty arbitrary to me. I couldn't find the definition that justify the * .1
I think it would be best to do the *.1 but I would not recommend applications to assume the resulting delta values equal the assumed jump in pixels in the case of a long swipe.

PR #276, opened 2018-11-09
JDK-8213619 Initialized additional environment variables.

DeanWookey:
Added gradlew --stop so that the new environment variables get used by gradle. Added chocolatey install lines for the non-webkit dependencies mentioned in the windows build instructions.

Not all dlls were being added to the published artifacts if the machine's setup didn't match the defaults in win.gradle. For example vcruntime140.dll was not being included on my machine because MSVC_REDIST_VER was not set and WINDOWS_CRT_VER was not set, resulting in the win.gradle not finding vcruntime140.dll , and not including it. My WINDOWS_CRT_VER was 141 for example where the default was 150.

The only way I could find WINDOWS_CRT_VER was by looking at the folder names. There is probably a better way.

The last change was to change -PCONF=DebugNative to -PCONF=Debug. That's because with DebugNative, the prism dlls depend on the debug versions (vcruntime140d.dll for example), and those aren't packaged correctly in win.gradle in the DebugNative mode. I can remove this change, but I imagine this build script will be used as a starting point for developers, and it seems like it would be better to have a more general starting point that works reliably.

### DIFF
--- a/tools/scripts/build.ps1
+++ b/tools/scripts/build.ps1
@@ -1,23 +1,38 @@
+.\gradlew --stop
 choco install ant
 choco install vswhere
 choco install zip
+choco install visualstudio2017community
+choco install visualstudio2017-workload-nativedesktop
+choco install windows-sdk-7.1
+choco install cygwin
+
+$cygwinPath = (Get-ItemProperty -Path "HKLM:\SOFTWARE\Cygwin\setup").rootdir
+$env:Path += ";$($cygwinPath)\bin"
+
 $env:WINSDK_DIR = (Get-ItemProperty -Path "HKLM:\SOFTWARE\Wow6432Node\Microsoft\Windows Kits\Installed Roots").KitsRoot10
-$env:VCINSTALLDIR = "$(vswhere -legacy -latest -property installationPath)\VC\Auxiliary\Build"
+$vsRoot = "$(vswhere -legacy -latest -property installationPath)"
+$env:VCINSTALLDIR = "$($vsRoot)\VC\Auxiliary\Build"
 $msvcToolsVer = Get-Content "$env:VCINSTALLDIR\Microsoft.VCToolsVersion.default.txt"
+$msvcRedistVer = Get-Content "$env:VCINSTALLDIR\Microsoft.VCRedistVersion.default.txt"
 if ([string]::IsNullOrWhitespace($msvcToolsVer)) {
   # The MSVC tools version file can have txt *or* props extension.
   $msvcToolsVer = Get-Content "$env:VCINSTALLDIR\Microsoft.VCToolsVersion.default.props"
 }
 $env:MSVC_VER = $msvcToolsVer
+$env:MSVC_REDIST_VER = $msvcRedistVer
+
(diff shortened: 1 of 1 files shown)


DeanWookey:
AppVeyor timed out because the build script downloaded Visual Studio even though it was present on the instance. The script only downloads visual studio, cygwin, and windows-sdk-7.1 if not running on AppVeyor now. I also changed the debug mode if running on AppVeyor back to what it was before.

kevinrushforth:
@DeanWookey - thank you for your proposed contribution to OpenJFX. I can confirm your OCA status, so as soon as this PR is approved, we can merge it in and push it to the mainline HG openjfx/jfx-dev/rt repo.

kevinrushforth:
@brcolow I'd like you to review these changes, if you could.

kevinrushforth:
Pointer to JBS issue: [JDK-8213619](https://bugs.openjdk.java.net/browse/JDK-8213619).

tools/scripts/build.ps1
```diff
@@ -1,14 +1,32 @@
+.\gradlew --stop
```
kevinrushforth: Rather than do this, it would be better to not run the gradle daemon in the first place. The easiest way to accomplish this is to add `--no-daemon` to the gradlew run in non-Appveyor mode (it already is added in Appveyor mode).

tools/scripts/build.ps1
```diff
@@ -1,14 +1,32 @@
...
 choco install ant
 choco install vswhere
 choco install zip
+if ($env:APPVEYOR -ne "true") {
+  choco install visualstudio2017community
+  choco install visualstudio2017-workload-nativedesktop
+  choco install windows-sdk-7.1
+  choco install cygwin
```
kevinrushforth: I presume that if the user already has installed cygwin via some other manual means, then this will be a no-op?

brcolow:
In my experiments with the build script it wasn't actually necessary to install cygwin. Did you actually run into any errors with cygwin not installed when trying to build using the script? Only other thing I see is what's the reason in splitting:

` $env:VCINSTALLDIR = "$(vswhere -legacy -latest -property installationPath)\VC\Auxiliary\Build" `

into two lines?

Other than those two questions, LGTM. Thanks for opening this PR.

tools/scripts/build.ps1
DeanWookey: I thought it would be, but it doesn't seem like it. I think I can do a better job of detecting if they're installed before trying to install them.

tools/scripts/build.ps1
DeanWookey: I'll do that, thanks.

DeanWookey:
1. I had a problem where the gradle build was trying to use chmod, so I had to install cygwin for that.
2. I split it because vsRoot is used later to get the WINDOWS_CRT_VER. I thought it would be better to invoke vswhere once, but I could call vswhere twice and make those into one-liners again?

brcolow:
LGTM. This would be out of scope for this PR of course but ideally if we could replace the usage of chmod with the native gradle tasks that do the same thing and get rid of the cygwin dependency, that would be a pretty big win towards simplifying the Windows build. I think that maybe it passed on my system without cygwin because I have WSL binaries in my PATH, so it probably grabbed `chmod` from that.

PR #281, opened 2018-11-12
JDK-8213541: WebView does not handle HTTP response without ContentType

Maxoudela:


muralibilla:
looks fine..did you check the conditions for CF port ?  If CF port still has both conditions, then we can remove the line   "as e.g. the CF port  does"  from the comment.

muralibilla:
Looks like CF port also don't check for contentLength. So the existing comment in URLLoader.cpp looks fine..

 if (_CFURLRequestCopyProtocolPropertyForKey(m_handle->firstRequest().cfURLRequest(DoNotUpdateHTTPBody), CFSTR("ForceHTMLMIMEType")))

CFURLResponseSetMIMEType(cfResponse.get(), CFSTR("text/html"));

Maxoudela:
Hi @muralibilla ,
I literally have no idea what a "CF port" is so I can tell you for sure that I haven't checked that. But if you can give me a pointer to a documentation, I'll gladly dig into and investigate in order to refine the comment.

muralibilla:
@Maxoudela  You can refer https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/platform/network/cf/ResourceHandleCFURLConnectionDelegateWithOperationQueue.cpp

Maxoudela:
Thanks @muralibilla ,
It seems indeed that the check is not done on the content length so the comment is good.

kevinrushforth:
@muralibilla or @ararunprasad can I ask one of you two to sponsor this change? Make sure you squash & merge when merging the PR into `develop`. Then you can create an HG changeset from that commit and push it to jfx-dev repo.

PR #289, opened 2018-11-16
8207957: TableSkinUtils should not contain actual code implementation

Maxoudela:
Allright, this is a first temporary fix for [JDK-8207957](https://bugs.openjdk.java.net/browse/JDK-8207957).

Apparently the compilation failed around the TableColumnBase somehow but I cannot put my finger on it :
```
/home/travis/build/javafxports/openjdk-jfx/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java:574: error: incompatible types: TableColumnBase<CAP#1,CAP#2> cannot be converted to CAP#3
            getTableSkin().resizeColumnToFitContent(column, cellsToMeasure);
                                                    ^
  where CAP#1,CAP#2,CAP#3,CAP#4 are fresh type-variables:
    CAP#1 extends Object from capture of ?
    CAP#2 extends Object from capture of ?
    CAP#3 extends TableColumnBase<CAP#4,?> from capture of ?
    CAP#4 extends Object from capture of ?
```
I don't understand that issue...

@aghaisas I'll be on holidays for 2 weeks so I just wanted to provide some modifications in order to start the discussion on the direction we would go for the issue resolution.


kevinrushforth:
Added "WIP" label.

Note that if you do want to do this, it will need to go through API review (including a CSR) since you are adding new protected methods to TableViewSkin, TreeTableViewSkin, and TableViewSkinBase. More importantly, adding an abstract method to TableViewSkinBase is a potentially incompatible change, since if there were an existing 3rd-party control that subclassed TableViewSkinBase directly it would fail to compile after this change.

aghaisas:
@kevinrushforth, Thanks for pointing out. Yes. It is potentially incompatible change. One thing to note is - jfx-8u has method resizeColumnToFitContent in TableView skin classes. (the skin classes are not public in jfx-8u though). This method was moved to TableSkinUtils class in jfx-9. 

@Maxoudela, Have you thought about suggestion from Jeanette on JBS?

Maxoudela:
From my very own point of view (which is limited), the modifications made in JDK 9 were outstanding incompatible change and the point of all my modifications are to allow 3rd-party controls to subclass again TableView and others. But I do understand your point @kevinrushforth .

@aghaisas Thanks for the feedback. I've initially gone for the "JDK8" solution because it was the easier one. Now I do understand that it might not be the best one. 

If we do like Jeannette suggests, that could be better. The only thing that worries me is that `TableColumnHeader `is used in `TreeTableView `and `TableView`. When we put back the methods into the skins, it was easy because we had `TreeTableViewSkin `and `TableViewSkin`. So if we want to put in `TableColumnHeader`, we would then have to mix the two methods somehow and make the switch whereas the `TableColumnHeader `is attached to a `TreeTableView `or a `TableView`. 

It could be interesting if @kleopatra could give us some insight on how she actually made it on her side.

Maxoudela:
Allright, I have made a new version of it. 

I have simply put the methods in TableColumnHeader. The methods doing the resize are private. But the method that do the switch is protected. Right now it's static but I think it shouldn't. 

Therefore, anyone who wants to overrides the behavior can provide its custom code in the method. And we do not introduce an abstract method so this should be good for the CSR.

Maxoudela:
I just removed the `static `modifiers as it would not have been possible to overrides the method. I'll create the CSR.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
```diff
@@ -48,10 +50,7 @@
 import javafx.scene.AccessibleAttribute;
 import javafx.scene.AccessibleRole;
 import javafx.scene.Node;
-import javafx.scene.control.ContextMenu;
-import javafx.scene.control.Label;
-import javafx.scene.control.TableColumn;
-import javafx.scene.control.TableColumnBase;
+import javafx.scene.control.*;
```
aghaisas: Wildcard imports should be avoided. Please add specific imports that you need.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
```diff
@@ -586,8 +587,177 @@ private void doColumnAutoSize(TableColumnBase<?,?> column, int cellsToMeasure) {
...
         if (prefWidth == DEFAULT_COLUMN_WIDTH) {
-            TableSkinUtils.resizeColumnToFitContent(getTableSkin(), column, cellsToMeasure);
-//            getTableViewSkin().resizeColumnToFitContent(column, cellsToMeasure);
+//            TableSkinUtils.resizeColumnToFitContent(getTableSkin(), column, cellsToMeasure);
+            resizeColumnToFitContent(getTableSkin(), column, cellsToMeasure);
+        }
+    }
+
+    /**
+     * Resizes the given column based on the pref width of all items contained in this column. This can be potentially
```
aghaisas: I suggest two changes to documentation :
1. pref --> preferred
2. "in this column" --> "in it"

"Resizes the given column based on the preferred width of all items contained in it"

aghaisas (commented):
Can you please provide a unit test?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
```diff
@@ -586,8 +587,177 @@ private void doColumnAutoSize(TableColumnBase<?,?> column, int cellsToMeasure) {
...
+        double headerTextWidth = Utils.computeTextWidth(header.label.getFont(), tc.getText(), -1);
+        Node graphic = header.label.getGraphic();
+        double headerGraphicWidth = graphic == null ? 0 : graphic.prefWidth(-1) + header.label.getGraphicTextGap();
+        double headerWidth = headerTextWidth + headerGraphicWidth + 10 + header.snappedLeftInset() + header.snappedRightInset();
+        maxWidth = Math.max(maxWidth, headerWidth);
+
+        // RT-23486
+        maxWidth += padding;
+        if (tv.getColumnResizePolicy() == TableView.CONSTRAINED_RESIZE_POLICY && tv.getWidth() > 0) {
+
```
muralibilla: you can remove the empty line here..

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkin.java
```diff
@@ -29,6 +29,7 @@
 import java.util.List;
 
 import com.sun.javafx.scene.control.Properties;
+import com.sun.javafx.scene.control.TableColumnBaseHelper;
```
muralibilla: You can re-order this import as per alphabetic order.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
Maxoudela: Thanks I have not seen that, my IDE was misconfigured

Maxoudela:
Thanks for the feedback. I have added a unit test. I spent the whole morning trying to add the unit test. I'm not proud of it but I don't quite know how to make it better. 

Do you have something to mock the class? Basically, I had extends the TableColumnHeader to allow the resizeCOlumnToFitContent, then override everything. And cherry on top, the TableColumnHeader is not accessible through lookup so I have to put it in a variable..

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
```diff
@@ -586,8 +590,175 @@ private void doColumnAutoSize(TableColumnBase<?,?> column, int cellsToMeasure) {
...
+     * Resizes the given column based on the preferred width of all items contained in it. This can be potentially very
+     * expensive if the number of rows is large.
+     *
+     * @param tableSkin the {@code TableViewSkinBase} used to determine if we are dealing with a {@link TableView} or a
+     *                  {@link TreeTableView}
+     * @param tc        the column to resize
+     * @param maxRows   the number of rows considered when resizing. If -1 is given, all rows are considered.
+     * @since 12
+     */
+    protected void resizeColumnToFitContent(TableViewSkinBase<?,?,?,?,?> tableSkin, TableColumnBase<?, ?> tc, int maxRows) {
```
kevinrushforth: Can you say why `TableViewSkinBase tableSkin` needs to be passed as an argument to this method? Unless I'm missing something, this method is always called with `this.getTableSkin()`, which suggests that the `tableSkin` parameter can be removed.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
```diff
@@ -586,8 +590,175 @@ private void doColumnAutoSize(TableColumnBase<?,?> column, int cellsToMeasure) {
...
+        if (!tc.isResizable()) return;
+
+        Object control = tableSkin.getSkinnable();
+        if (control instanceof TableView) {
+            resizeColumnToFitContent((TableView)control, (TableColumn)tc, tableSkin, maxRows);
+        } else if (control instanceof TreeTableView) {
+            resizeColumnToFitContent((TreeTableView)control, (TreeTableColumn)tc, tableSkin, maxRows);
+        }
```
kevinrushforth: This assumes that the only two specializations of a table control that uses `TableViewSkinBase` are `TableView` and `TreeTableView`. This is likely fine, since I doubt anyone could actually implement a new type of table whose Skin subclassed `TableViewSkinBase` without subclassing one of `TableView` or `TreeTableView` (and thus also subclassing TableViewSkin or TreeTableViewSkin).

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkin.java
```diff
@@ -136,7 +137,6 @@ public TableViewSkin(final TableView<T> control) {
     }
 
 
-
```
kevinrushforth: There are no changes in this file other than white space and unused imports, so this file can be reverted.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
```diff
@@ -346,9 +346,6 @@ public TableViewSkinBase(final C control) {
      **************************************************************************/
 
 
-
-
-
```
kevinrushforth: There are no changes in this file other than white space, so this file can be reverted.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableViewSkin.java
```diff
@@ -275,8 +276,6 @@ protected void executeAccessibleAction(AccessibleAction action, Object... parame
         }
     }
 
-
-
```
kevinrushforth: There are no changes in this file other than white space and unused imports, so this file can be reverted.

kevinrushforth (commented):
1. Since you are proposing new API, and there are a couple alternatives under discussion, can you send an email to openjfx-dev with a pointer to the JBS issue and this pull request so others who might not be watching the PR can comment on the proposed API addition?
2. This is still listed as WIP. If you are ready for a formal review of the proposed changes, then please remove the WIP label and the "[WIP]" text from the title.
3. I'd like a little more info on how a subclass would take advantage of the new API. Is it being exposed primarily to allow subclasses of `TableColumnHeader` to call the new `resizeColumnToFitContent` method? Or is it also expected that those subclasses might override the method? Either way, this should be documented in the javadoc comments.
4. Related to point 3 above: Based on [this comment](https://github.com/javafxports/openjdk-jfx/pull/163#issuecomment-438297212), have you checked whether it is sufficient to expose resizeColumnToFitContent given that resizeColumn is not accessible (it still lives in TableSkinUtils)? Does this proposed change allow ControlsFX TableView2 to customize the column resizing?

As a final note: Given the RDP1 deadline, the Christmas holidays, and the fact that this is still work in progress, it is very likely too late to get this into openjfx12 prior to the RDP1, so targeting it for openjfx13 seems best.

Maxoudela:
I have made some changes. I have indeed removed the `TableSkin `parameter as it was not helpful now that the method in in `TableColumnHeader `class.

Subclass of `TableColumnHeader `can either call the method which was not possible before. They also can override the method to either closely control when a resize is happening (they may prevent it by not calling `super()`). Also overriding this method without the `super()` call is possible to completely override the behavior and provide a custom resize algorithm. The `resizeColumn()` of `TableSkinUtils `is indeed unreachable but it is possible to call `resizeColumn()` of the Control itself. Anyone overriding the `resizeColumnToFitContent()` will call directly the `resizeColumn()` on the `TableView `or `TreeTableView`.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
```diff
@@ -586,8 +590,174 @@ private void doColumnAutoSize(TableColumnBase<?,?> column, int cellsToMeasure) {
...
         if (prefWidth == DEFAULT_COLUMN_WIDTH) {
-            TableSkinUtils.resizeColumnToFitContent(getTableSkin(), column, cellsToMeasure);
-//            getTableViewSkin().resizeColumnToFitContent(column, cellsToMeasure);
+            resizeColumnToFitContent(column, cellsToMeasure);
+        }
+    }
+
+    /**
+     * Resizes the given column based on the preferred width of all items contained in it. This can be potentially very
+     * expensive if the number of rows is large. Subclass can either call this method or override it (no need to call
```
kleopatra: Hmm... I think the doc is a) not entirely correct and b) overspecific on this particular implementation

a) Resizes the column based on content of header _and_ content of cells
b) This implementation measures the preferred width of the header, the preferred width of the first maxRow cells configured with their corresponding content and sizes the column to the max width of all measured values. 

Custom implementations can ... [whatever - exclude headers, use the last n, exclude null content, use min ... ]

Maybe the different aspects should be separated into distinct methods (that's what we did way back in SwingX, those were the days ;) , names pending ...

protected void resizeColumnToFitContent(...) {
     ...
    // measure
    double headerWidth = measureHeaderWidth(...);
    double contentWidth = measureContentWidth(...);
    // do the sizing 
   resizeColumnToFit(headerWidth, contentWidth);

The default implementations can be the same as now, custom descendents can override each aspect separately. In my particular use-case the header was very different from the default, such that the default measure was way off.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
```diff
@@ -586,8 +590,174 @@ private void doColumnAutoSize(TableColumnBase<?,?> column, int cellsToMeasure) {
...
+     * {@code super()}) to provide their custom algorithm.
+     *
+     * @param tc      the column to resize
+     * @param maxRows the number of rows considered when resizing. If -1 is given, all rows are considered.
+     * @since 12
+     */
+    protected void resizeColumnToFitContent(TableColumnBase<?, ?> tc, int maxRows) {
+        if (!tc.isResizable()) return;
```
kleopatra: why do we need the column as parameter? Now that the header is the responsible party, it should be concerned about its own column only - or not?

Maxoudela: In the `rectMousePressed` event of `NestedTableColumnHeader`, the column is retrieved like that :
```
Rectangle rect = (Rectangle) me.getSource();
        TableColumnBase column = (TableColumnBase) rect.getProperties().get(TABLE_COLUMN_KEY);

```
I do not know if it's guaranteed that the column retrieved is the TableColumn associated with the current NestedTableColumnHeader. If yes, we can indeed remove the TableColumn parameter.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
kleopatra: that's still too specific, IMO: the general contract should be to auto-size _somehow_ based on content of header and cells. This specific implementation uses max of _prefWidth_ of all contributing nodes. There might be others: min/max or header-only or cell-only or first cells in unsorted backing list or not-empty cells ... the contract shouldn't block those :)

As to the additional methods, after sleeping over it: if we have a single override hook now we can add them at a later time, otherwise the scope of this issue might explode.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
kleopatra: good point, I'll check later

kleopatra: with the current impl we need the tableColumn as parameter: the header that's stored in the dragRect properties is the parent of the tableColumnHeader that represents it. 

But then, we are back to letting unrelated parties do the sizing: the parent does the sizing on behalf of the real header: pre-9 it was the skin, since then it was the utils, now it's the parent. Ideally, the sizing would be handled by the "real" header, IMO. That's where the current sizing code resides (in TableColumnHeader) but it is invoked from upper with an arbitrary column (which is a child of the column the nested is responsible for) ... hmm ...

kleopatra: To clarify my base a bit: the setup of auto-sizing in core is a unusual (read: never was/is what I expected ;)- it's managed by nestedColumnHeader when it should be handled by columnHeader and then only on its own tableColumn (not others) and no other party should interfere with that sizing. The mouse interaction is installed in Nested because it manages the dragRects and the handlers are intalled on those. That's fine, iff the handler delegates the actual sizing to the header that's representing the column. 

In the pressed handler, something like:

````java

TableColumnBase column = (TableColumnBase) rect.getProperties().get(TABLE_COLUMN_KEY);
 NestedTableColumnHeader parentHeader = (NestedTableColumnHeader) rect.getProperties().get(TABLE_COLUMN_HEADER_KEY);
   ...

if (me.getClickCount() == 2 && me.isPrimaryButtonDown()) {
     // the user wants to resize the column such that its
     // width is equal to the widest element in the column
     TableHeaderRow tableHeader = parentHeader.getTableHeaderRow(); 
     TableColumnHeader columnHeader = tableHeader.getColumnHeaderFor(column);
     // note the missing column param ... 
     columnHeader.resizeColumnToFitContent(-1);
 } else {
  ....
}
````

Just a snippet, untested and not fully thought through :) In particular, didn't check other locations that message the auto-size method.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
Maxoudela: Do you have a Javadoc proposition for this point?

Maxoudela:
Hi,

What is the status on this PR? @kleopatra ? 

What I see is a documentation issue left and potentially the fact that the code should be a the "real" header.

Maxoudela:
@kevinrushforth  Could we remove the WIP label as no more work is needed on my side.

kevinrushforth:
> Could we remove the WIP label as no more work is needed on my side.

Certainly. I didn't realize you were ready for formal review. Given that you sent out your preliminary RFR (back when it was still a WIP) a while ago, can you please resend it? It will prompt others who might not be watching this PR to take a look.

kevinrushforth:

This [sandbox repository](https://github.com/javafxports/openjdk-jfx) is being retired on 1-Oct-2019. You will need to migrate your pull request to the openjdk/jfx repo in order for this review to continue. Here are [instructions for migrating your pull request](https://mail.openjdk.java.net/pipermail/openjfx-dev/2019-September/023560.html). The updated [CONTRIBUTING.md](https://github.com/openjdk/jfx/blob/master/CONTRIBUTING.md) doc has additional information on how to proceed.

kevinrushforth:
Moved to: [openjdk/jfx/pull/6](https://github.com/openjdk/jfx/pull/6)

PR #300, opened 2018-11-26
Improvements in javafx cache for native files (JDK-8214397)

johanvos:
* first check for a new property "javafx.cacheDir"
* if that exists, use it
* it not, use user.home directory
* if a cache directory can't be created under user.home, use java.io.tmpdir

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
@@ -240,7 +240,8 @@ public class NativeLibLoader {
             }
         }
         if (!cacheDirOk) {
-            String tmpCache = System.getProperty("java.io.tmpdir") + "/.openjfx/cache/" + jfxVersion;
+            String username = System.getProperty("user.name", "anonymous");
+            String tmpCache = System.getProperty("java.io.tmpdir") + "/.openjfx/"+username+"/cache/" + jfxVersion;
             cacheDir = new File(tmpCache);
             if (cacheDir.exists()) {
                 if (!cacheDir.isDirectory()) {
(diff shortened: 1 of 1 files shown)


modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -222,15 +222,34 @@ private static boolean installLibraryFromResource(String libraryName, List<Strin
...
             }
         } else {
             if (!cacheDir.mkdirs()) {
-                throw new IOException ("Can not create cache at "+cacheDir);
+                System.err.println("Can not create cache at "+cacheDir);
+                cacheDirOk = false;
+            }
+        }
+        if (!cacheDirOk) {
+            String tmpCache = System.getProperty("java.io.tmpdir") + "/.openjfx/cache/" + jfxVersion;
```
kevinrushforth: Do you think appending the username might be useful? Otherwise it may not work for multiple users accessing the same directory.

johanvos: I added the username.

johanvos:
I tested this on mac, by chowning the existing cache directory to root, and removing the existing cache files.
* supply -Djavafx.cachedir=/tmp/foo creates the /tmp/foo directory 
* not doing anything creates a directory inside the system tmpdir, adding the version to the path.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
manorrock: May I suggest auto generating a sub directory within the java.io.tmpdir instead?

manorrock: Using createTempFile?

kevinrushforth:
Pointer to JBS issue: [JDK-8214397](https://bugs.openjdk.java.net/browse/JDK-8214397).

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
johanvos: I think although in the case the user.home is for some reason not applicable and we use the tmp directory, the directory we create should not really be considered a temporarily directory. 
When the process exists, there is no need to remove the directory, at the contrary (so we don't need the hooks for deleting on exit etc)

PR #308, opened 2018-11-29
Fix toolbar overflow [JDK-8152395]

Ciruman:
* This is a proposed fix for JDK-8152395
* Correct toolbar overflow if toolbar items size changes. It can be done by evaluating the items size change during layout process.


Ciruman:
I did not read the corresponding contribution files yet. I will reopen it as soon as I read it. I sent the OCA a few years ago(and I sent contributions to Jonathan already) but I have to re-send it.

kevinrushforth:
@Ciruman go ahead and reopen this when you are ready. Dalibor can update the OCA signatories list in parallel with the review (since we have email confirmation that your OCA was accepted).

kevinrushforth:
Also, please add a unit test for this fix. The test should fail without your fix and pass with your fix.

Issue #273, opened 2018-11-07
Embedded: Error undefined pixcoord starting javafx app


dellgreen:
[pixcoord_error.txt](https://github.com/javafxports/openjdk-jfx/files/2557179/pixcoord_error.txt)

Currently using Vivante GC2000 OpenGL ES 3.0 V6.2.4.p1.150331
If i run exact same tests on an older version of our product it works fine, it uses
Currently using Vivante GC2000 OpenGL ES 3.0 V5.0.11.p4.25762

dellgreen:
[vivante.txt](https://github.com/javafxports/openjdk-jfx/files/2557189/vivante.txt)

dellgreen:
I can see this was discussed back in 2016, but unclear whether a fix actually happened or whether its a regression in javafx?

http://mail.openjdk.java.net/pipermail/openjfx-dev/2016-February/018753.html

dellgreen:
In the meantime, i'll try to downgrade vivante driver to find out when the breakage was introduced

dellgreen:
so i have failed miserably in trying to downgrade from v6 to v5 in yocto. Any ideas on what to fix in javafx to make it usable with v6?

johanvos:
The fix you refer to in your early comments never made it to the OpenJFX repository (we only ported it to mobile). It's probably a good time to do it now.

dellgreen:
That would be awseome, and save my bacon, as i have rebased on latest yocto distro and cannot get our existing javafx frontend running on the device.  :)

dellgreen:
Please let me know if you need help testing changes on our board, or any other help I can provide to to make this happen :)

dellgreen:
Do you think its worth me trying the patches here
https://bitbucket.org/javafxports/8u60-rt/commits/595633bbaae36f98d85d47d276294442ea43488c
to see if it fixes my issue?

dlemmermann:
Yes, this looks great. Thanks!

And I can freely share this now? Asking because we are recording the keynote session.

Dirk

kevinrushforth:
Link to JBS bug: [JDK-8167068](https://bugs.openjdk.java.net/browse/JDK-8167068)

kevinrushforth:
This is fixed, so closing this bug.

PR #312, opened 2018-12-03
8167068: Fixed undefined pixcoord error starting javafx

dellgreen:
This pull request is based upon patches created by @johanvos and fixes the problem on embedded javafx and allows the application to start up as opposed to crashing on startup.

Seems to be duplicate of JBS: 8167068

### DIFF
--- a/modules/javafx.graphics/src/jslc/java/com/sun/scenario/effect/compiler/backend/hw/SLBackend.java
+++ b/modules/javafx.graphics/src/jslc/java/com/sun/scenario/effect/compiler/backend/hw/SLBackend.java
@@ -81,8 +81,20 @@ public abstract class SLBackend extends TreeScanner {
         sb.append(s);
     }
 
+
+    static String PIXCOORD = "vec2 pixcoord = vec2(\n"+
+        "    gl_FragCoord.x-jsl_pixCoordOffset.x,\n" +
+        "    ((jsl_pixCoordOffset.z-gl_FragCoord.y)*jsl_pixCoordOffset.w)-jsl_pixCoordOffset.y);\n";
+
+    static String MAIN = "void main() {\n";
+
+
     public final String getShader() {
-        return getHeader() + sb.toString();
+        String answer = getHeader() + sb.toString();
+        if (isPixcoordReferenced) {
+            answer = answer.replace(MAIN, MAIN + PIXCOORD);
+        }
+        return answer;
     }
 
     protected final JSLParser getParser() {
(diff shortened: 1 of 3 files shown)


dellgreen:
Fixes  #273

modules/javafx.graphics/src/jslc/java/com/sun/scenario/effect/compiler/backend/hw/SLBackend.java
```diff
@@ -81,8 +81,20 @@ protected final void output(String s) {
         sb.append(s);
     }
 
+
+    static String PIXCOORD = "vec2 pixcoord = vec2(\n"+
+        "    gl_FragCoord.x-jsl_pixCoordOffset.x,\n" +
+        "    ((jsl_pixCoordOffset.z-gl_FragCoord.y)*jsl_pixCoordOffset.w)-jsl_pixCoordOffset.y);\n";
```
kevinrushforth: This GLSL shader code doesn't belong here. `SLBackend` is the super-class for both of the `HLSLBackend` (DirectX) and `GLSLBackend` (OpenGL). If there is no way to currently do what is needed entirely in `GLSLBackend`, then one possible approach might be to make `SLBackend::getShader` non-final, and override it in `GLSLBackend`

dellgreen: Sounds like the latter would be the preferred choice as the bugs are reported for opengl on armhf and x86_64 specifically, and Its unclear if this could be an issue for directx or not.

dellgreen: I'll refactor in the morning, and update unless @johanvos has any further thoughts on this?

kevinrushforth:
@dellgreen I got confirmation this morning that your OCA has been processed, so I'll put this back on my queue to review.

kevinrushforth:
Link to JBS bug: [JDK-8167068](https://bugs.openjdk.java.net/browse/JDK-8167068)

modules/javafx.graphics/src/jslc/java/com/sun/scenario/effect/compiler/backend/hw/SLBackend.java
johanvos: iirc, the bug is specific to some drivers that are used on linux-based systems, so I don't think it will show in platforms using DirectX. It's a pity we need separate code for the HLSLBackend and the GLSLBackend but it's probably the cleanest way to address a driver issue like this.

johanvos:
Yes, I will do more testing first though.

dellgreen:
cool, many thanks :+1:

PR #313, opened 2018-12-04
 Remove final from HashSet instance (JDK-8215686)

BlueGoliath:
Removes final from HashSet instance in buildSrc/src/main/groovy/com/sun/javafx/NativeCompileTask.groovy at about line 107 which causes build failure. There is no negative impact by doing this as code after the instance is created requires that the HashSet instance be mutable to begin with.

### DIFF
--- a/buildSrc/src/main/groovy/com/sun/javafx/gradle/NativeCompileTask.groovy
+++ b/buildSrc/src/main/groovy/com/sun/javafx/gradle/NativeCompileTask.groovy
@@ -102,7 +102,7 @@ class NativeCompileTask extends DefaultTask {
         updateFiles();
         def source = project.files(allFiles);
         boolean forceCompile = false;
-        final Set<File> files = new HashSet<File>();
+        Set<File> files = new HashSet<File>();
         source.each { File file ->
             final Map fileData = dependencies.get(file.toString());
             final boolean isModified = fileData == null ||


buildSrc/src/main/groovy/com/sun/javafx/gradle/NativeCompileTask.groovy
```diff
@@ -102,7 +102,7 @@ class NativeCompileTask extends DefaultTask {
         updateFiles();
         def source = project.files(allFiles);
         boolean forceCompile = false;
-        final Set<File> files = new HashSet<File>();
+        Set<File> files = new HashSet<File>();
```
kevinrushforth: I'm not sure that this code is in error. The `files` variable is a set and is assigned only once in the above declaration. Elements are added to the `files` set, but that can still be done even if `files` is final.

BlueGoliath:
Version 5.0-1 from Arch Linux's repo.

After looking at the code again I realized that you're technically right. The code that is supposed to add the files to the HashSet instance is what is wrong. All 3 lines(118, 122, and  128) try to add the files using the += operator and not by HashSet.add(<File>) which cause the error that I was getting because it's trying to reassign the HashSet variable.

I don't know Groovy so I guess I kind of just assumed it was magically somehow valid and that the error that I was getting was the actual problem but none of the Groovy language += usages docs that I can find says that += operator is valid in those 3 lines(how could it be?).

BlueGoliath:
That's fine, go ahead and push it. In hindsight I feel like using add() to add the file object to the HashSet would be a way better "fix" as using += is both wrong in that it supposedly reassigns instead of simply adding and is both syntactically misleading and confusing which is probably why they are getting more strict about it.

kevinrushforth:
Tracked in JBS as [JDK-8215686](https://bugs.openjdk.java.net/browse/JDK-8215686).

kevinrushforth:
I'll go with the simpler solution of removing `final`, since it really doesn't add any value where it is used.

kevinrushforth:
A quick test on Linux with the removal of "final" gets past this initial error but then dies a little further on. I'm going to close this PR, and open a new one once all of the gradle 5 issues are fixes.

DJViking:
Did you go with the fix of removing final?
I tried using add/addAll instead and it worked fine with Gradle 5.2

kevinrushforth:
I put this on hold during the stabilization for openjfx12 and am just this week getting back to it in order to upgrade to gradle 5.3, which is tracked by [JDK-8218172](https://bugs.openjdk.java.net/browse/JDK-8218172).

I plan to remove `final` from the definition of files as the least-intrusive change. You mention that it work on gradle 5.2 for you with the fix to `NativeCompileTask.groovy`, but I still see two follow-on errors that need to be addressed:

1. A complaint about the maven publishing task since gradle 5 no longer does deferred evaluation of the publishing block by default. This is already tracked by [JDK-8210509](https://bugs.openjdk.java.net/browse/JDK-8210509).

2. An error in the javafxSwt task caused by a misuse of the archiveName property (that property should just be the base filename and the path to the directory should be set in the `directoryName` property).

Without fixing the above two, running `gradle` to build the default `sdk` task fails.

I already have a fix for issue 2 and will be looking at issue 1.

PR #318, opened 2018-12-07
Fix toolbar overflow (JDK-8152395)

Ciruman:
test (JDK8152395) toolbar overflow control visibility
    
    - if children size changed and they are bigger than toolbar shall be shown
    - if children size changed and they are smaller than toolbar shall not be shown

fix (JDK8152395) toolbar overflow control visibility
    
    - if children size is bigger than toolbar shall be shown
    - if children size is smaller than toolbar shall not be shown


kevinrushforth:
JBS bug: https://bugs.openjdk.java.net/browse/JDK-8152395

tests/system/src/test/java/test/robot/javafx/scene/ToolBarTest.java
```diff
@@ -0,0 +1,159 @@
+package test.robot.javafx.scene;
```
kevinrushforth: This will need a standard Copyright header

tests/system/src/test/java/test/robot/javafx/scene/ToolBarTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+    private static final double TOOLBAR_SIZE = 300.0;
+    private static final double CHILDREN_SIZE_TOO_BIG = 250.0;
+    private static final double ORIGINAL_CHILDREN_SIZE = 100.0;
+
+    private static final String ORIGINAL_STYLE = "-fx-border-width: 1; -fx-border-color: deepskyblue";
+    private static final String MODIFIED_STYLE = "-fx-border-width: 1; -fx-border-color: red";
+
+    static CountDownLatch startupLatch = new CountDownLatch(1);
+    static Robot robot;
```
aghaisas: robot is not at all used. 
The question is why should this test be under test.robot.javafx.scene package?

kevinrushforth: Assuming that it even needs to be a system test, then `test/javafx/scene/control` seems an OK place.

My question is whether this needs to be a system test at all? If you can make it a javafx.controls unit test then that would be best (almost all controls tests can be done that way).

aghaisas: If you decide to write a unit test - it can be added as a test case in existing test file rt/modules/javafx.controls/src/test/java/test/javafx/scene/control/ToolbarTest.java

else, system test under location suggested by Kevin looks fine.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
```diff
@@ -588,7 +609,11 @@ private void addNodesToToolBar() {
                 break;
             }
         }
+        return hasOverflow;
+    }
 
+    private void organizeOverflow(double length, boolean hasOverflow) {
+        double x;
```
aghaisas: This declaration can be moved to it's first use.

tests/system/src/test/java/test/robot/javafx/scene/ToolBarTest.java
Ciruman: I think the system test is nice to have but it could make sense to have a unit test.
- I will try to implement one.
- I will move the test to the test/javafx/scene/control package.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
Ciruman: I reorganized the methods close to its first use.

tests/system/src/test/java/test/robot/javafx/scene/ToolBarTest.java
Ciruman: Tests move to proposed test class:
rt/modules/javafx.controls/src/test/java/test/javafx/scene/control/ToolbarTest.java

kevinrushforth: To answer your question, by "javafx.controls unit test", I meant a test under `modules/javafx.controls/src/test/java/...` (as opposed to `tests/system/...`), which is what your latest change does. Did you verify that the test fails without your fix and passes with your fix?

Ciruman:
Is correct now? Should I do something else?

kevinrushforth:
You still need to send out a review request email to trigger the formal review.

Also, can you answer the following question:

>Did you verify that the test fails without your fix and passes with your fix?

Ciruman:
> You still need to send out a review request email to trigger the formal review.
> 
> Also, can you answer the following question:
> 
> > Did you verify that the test fails without your fix and passes with your fix?

It does.

With the fix there is no error but without the fix it fails with the following error:
test.javafx.scene.control.ToolbarTest > overflowShownInHorizontalAfterChildrenReziseTest FAILED
    java.lang.AssertionError: 
        at org.junit.Assert.fail(Assert.java:91)
        at org.junit.Assert.assertTrue(Assert.java:43)
        at org.junit.Assert.assertTrue(Assert.java:54)
        at test.javafx.scene.control.ToolbarTest.assertOverflowShown(ToolbarTest.java:309)
        at test.javafx.scene.control.ToolbarTest.testOverflowVisibility(ToolbarTest.java:275)
        at test.javafx.scene.control.ToolbarTest.overflowShownInHorizontalAfterChildrenReziseTest(ToolbarTest.java:249)

test.javafx.scene.control.ToolbarTest > overflowShownInVerticalAfterChildrenReziseTest FAILED
    java.lang.AssertionError: 
        at org.junit.Assert.fail(Assert.java:91)
        at org.junit.Assert.assertTrue(Assert.java:43)
        at org.junit.Assert.assertTrue(Assert.java:54)
        at test.javafx.scene.control.ToolbarTest.assertOverflowShown(ToolbarTest.java:309)
        at test.javafx.scene.control.ToolbarTest.testOverflowVisibility(ToolbarTest.java:275)
        at test.javafx.scene.control.ToolbarTest.overflowShownInVerticalAfterChildrenReziseTest(ToolbarTest.java:257)

16 tests completed, 2 failed

Is it enough like this or do we need a better error message? I think the test method name is clear: overflowShownIn{Vertical|Horizontal}AfterChildrenReziseTest. But If it could be better, please, let me know.

I will send the email tomorrow I do not have access to the Zeiss emails.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ToolbarTest.java
```diff
@@ -58,16 +63,30 @@
...
+
     @Before public void setup() {
         tk = (StubToolkit)Toolkit.getToolkit();//This step is not needed (Just to make sure StubToolkit is loaded into VM)
         toolBar = new ToolBar();
         node1 = new Rectangle();
         node2 = new Rectangle(2.0,4.0);
         toolBarWithItems = new ToolBar(node1,node2);
+
+        root = new StackPane();
+        scene = new Scene(root);
```
aghaisas: scene & stage creation + stage.setScene() can be moved to private method initializeToolBar(), Also we need not have them as class members.

kevinrushforth:
I'll review it when I get a chance. To answer your question:

>16 tests completed, 2 failed
>
>Is it enough like this or do we need a better error message?

This is sufficient.

Ciruman:
I was considering squashing all changes in a single commit because in the documentation recommends this. But if I do this now, I do not know what will happens with the comments.

kevinrushforth:
> I was considering squashing all changes in a single commit because in the documentation recommends this. But if I do this now, I do not know what will happens with the comments.

This is only recommended for the initial creation of the PR. Once you have a PR, the recommendation is to not squash the commits (so that reviewers who have already started reviewing can look at the delta changes).

modules/javafx.controls/src/test/java/test/javafx/scene/control/ToolbarTest.java
aghaisas: Thanks for making the change.
You have not addressed the second half my earlier comment - scene and stage need not be class members.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
aghaisas: Thanks.
My comment was for variable 'x' declaration. It is declared on line 578, but initialized on line 589.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ToolbarTest.java
```diff
@@ -221,4 +234,78 @@
...
+
+        assertOverflowShown();
+
+        setFixSize(node1, ORIGINAL_CHILDREN_SIZE);
+
+        assertOverflowNotShown();
+    }
+
+    private void setFixSize(Node node, double size) {
+        if(node instanceof Region) {
```
aghaisas: There should be a space between if and (

modules/javafx.controls/src/test/java/test/javafx/scene/control/ToolbarTest.java
```diff
@@ -221,4 +234,78 @@
...
+            ((Region) node).setMinSize(Pane.USE_PREF_SIZE, Pane.USE_PREF_SIZE);
+            ((Region) node).setMaxSize(Pane.USE_PREF_SIZE, Pane.USE_PREF_SIZE);
+            ((Region) node).setPrefSize(size, size);
+        } else if(node instanceof Rectangle) {
```
aghaisas: There should be a space between if and (

modules/javafx.controls/src/test/java/test/javafx/scene/control/ToolbarTest.java
```diff
@@ -221,4 +234,78 @@
...
+        }
+
+        root.applyCss();
+        root.autosize();
+        root.layout();
+    }
+
+    private void assertOverflowNotShown() {
+        Pane pane = (Pane) toolBar.queryAccessibleAttribute(AccessibleAttribute.OVERFLOW_BUTTON);
+        if(pane!=null) {
```
aghaisas: There should be a space between if and (.
Also around comparison operator

Correct version would be : if (pane != null)

aghaisas: I think, you can use assertNotNull(pane); instead of if condition. This is already done in assertOverflowShown() method.

Ciruman: You are right. overflowMenu is always initialized, and never should be null. I will leave it in two asserts and if it fails will be easier to find why.

Ciruman:
BTW, modules/javafx.controls/src/test/java/test/javafx/scene/control/ToolbarTest.java should be modules/javafx.controls/src/test/java/test/javafx/scene/control/ToolBarTest.java should I rename it? The control name is ToolBar.

kevinrushforth:
> BTW, .../ToolbarTest.java should be .../ToolBarTest.java should I rename it?

No. This would be an unrelated change that should not be done as part of fixing this bug.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
Ciruman: Done.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ToolbarTest.java
aghaisas: Have you missed fixing this?
I am OK with the changes and will approve once you fix this comment.

aghaisas:
> Have you missed fixing this?
> I am OK with the changes and will approve once you fix this comment.

@Ciruman, I guess, you did not see my last comment made 2 days ago?

aghaisas:
This is ready to be committed now.

Have you signed Oracle Contributor Agreement?
This is pre-requisit for the submission as listed at - https://github.com/johanvos/openjfxorg/blob/master/CONTRIBUTING.md

Ciruman:
> This is ready to be committed now.
> 
> Have you signed Oracle Contributor Agreement?
> This is pre-requisit for the submission as listed at - https://github.com/johanvos/openjfxorg/blob/master/CONTRIBUTING.md

Yes, I have.

kevinrushforth:
@aghaisas Thank you for checking, since this is a requirement. I can confirm that he has signed the OCA and is listed on the OCA signatory list.

PR #328, opened 2018-12-18
Allow for skipping Cygwin installation when using build.ps1 (JDK-8215629)

### LINKED ISSUE #297: Is Cygwin actually necessary to build OpenJFX on Windows?
I am setting up a new Windows PC with a fresh environment (specifically it does not have Cygwin nor WSL added to the Windows `%PATH%` environment variable).

In #276 I asked the question whether or not Cygwin was really necessary to build OpenJFX on Windows and @DeanWookey wisely noted that the only usage of it seems to be `chmod`. Looking at the source it seems this bit:

```gradle
// Need to modify file permissions Windows to make sure that the
// execute bit is set, and that the files are world readable
def chmodArtifactsSdkTask = task("chmodArtifactsSdk$t.capital", dependsOn: copyArtifactsSdkTask) {
	if (IS_WINDOWS) {
		doLast {
			exec {
				workingDir(sdkArtifactsDir)
				commandLine("chmod", "-R", "755", ".")
			}
		}
	}
}
```

is the only *hard* requirement for a Cygwin-like environment (specifically, a `chmod` binary in the `%PATH%`). A couple things to note:

1.) Simply commenting out the `exec` bit that calls `chmod` results in a working build. Is this change of permissions actually necessary if one is not running in Cygwin?
2.) I know there is quite a lot of stuff in the build files related to Cygwin (for example the `cygpath` function which is called on nearly every path) but from my testing it is *not* necessary to have Cygwin installed nor be in a Cygwin shell to build OpenJFX.
3.) Running the `tools\scripts\build.ps1` build file on a fresh Windows 10 install with only `choco` installed, and the part of the `build.ps1` script that installs Cygwin and adds it to the %PATH% commented out, the build (and all tests) succeed.

So it seems that Cygwin really is not necessary to build OpenJFX on Windows. Cygwin is a great tool, and I have used it extensively in the past, but it is a burden for new developer's setting up a Windows development environment to work on OpenJFX that we require it. Am I missing something critical/important about why Cygwin is actually required? If we can come to a consensus about it not actually being required we could change `build.ps1` to not install it, remove the call to `chmod` on Windows, change the Wiki build instructions, and maybe even remove the `cygpath` method down the line.

Thoughts?

brcolow:
Currently this only skips a call to chmod and is only useful for
local development builds of JavaFX. To skip Cygwin installation
one can use the "-nocygwin" flag to build.ps1 thusly:

.\tools\scripts\build.ps1 -nocygwin

The USE_CYGWIN gradle property will default to true.

Fixes #297

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -475,6 +475,11 @@ ext.IS_INCLUDE_ES2 = Boolean.parseBoolean(INCLUDE_ES2)
 defineProperty("JCOV", "false")
 ext.DO_JCOV = Boolean.parseBoolean(JCOV)
 
+// Specifies whether to use Cygwin when building OpenJFX. This should only ever
+// be set to false for development builds (that skip building media and webkit).
+defineProperty("USE_CYGWIN", "true")
+ext.IS_USE_CYGWIN = Boolean.parseBoolean(USE_CYGWIN)
+
 // Define the number of threads to use when compiling (specifically for native compilation)
 // On Mac we limit it to 1 by default due to problems running gcc in parallel
 if (IS_MAC) {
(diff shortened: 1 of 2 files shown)


brcolow:
@kevinrushforth Will do in the next couple days.

brcolow:
JBS enhancement request: https://bugs.openjdk.java.net/browse/JDK-8215629

Sent code review request to openjfx-dev.

PR #336, opened 2019-01-02
JDK-8215799: [Windows] Complex texts are not rendered properly

ararunprasad:
WebKit uses GlyphPage table to quickly decide whether any glyph is available for given character. For our JavaFX port, we get char to glyph mapping using CharToGlyphMapper. 

On macOS and Linux it is working because the LogicalFont had valid fallback for complex texts also, but it is missing for Windows. Refer PrismFontFactory.getLinkedFonts. 

606.1 WebKit used GlyphTable only for simple texts, however 607.1 uses the same for complex texts also. 

I could think of the following workarounds., 

1) Add "Arial Unicode MS Regular" as a fallback for Windows, it has almost all glyphs for the unicode range. 
2) Call "TextLayout.getRuns" before invoking "CharToGlyphMapper", "TextLayout.getRuns" finds & adds the native fallback font.

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/font/PrismFontFactory.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/font/PrismFontFactory.java
@@ -1024,6 +1024,10 @@ public abstract class PrismFontFactory implements FontFactory {
         fontRegInfo[0].add(jreFontDir + jreDefaultFontFile);
         fontRegInfo[1].add(jreDefaultFont);
 
+        // Add Arial Unicode to Windows fallback list
+        fontRegInfo[0].add(getPathNameWindows("arialuni.ttf"));
+        fontRegInfo[1].add("Arial Unicode MS");
+
         if (PlatformUtil.isWinVistaOrLater()) {
             // CJK Ext B Supplementary character fallbacks.
             fontRegInfo[0].add(getPathNameWindows("mingliub.ttc"));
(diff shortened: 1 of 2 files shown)


ararunprasad:
This PR is based on approach 1.

modules/javafx.graphics/src/main/java/com/sun/javafx/font/PrismFontFactory.java
```diff
@@ -1024,6 +1024,10 @@ private static void getPlatformFontDirs() {
         fontRegInfo[0].add(jreFontDir + jreDefaultFontFile);
         fontRegInfo[1].add(jreDefaultFont);
```
ararunprasad: Above 2 statements are invalid because there is no JRE present now.

kevinrushforth:
Link to JBS bug: https://bugs.openjdk.java.net/browse/JDK-8215799

modules/javafx.graphics/src/main/java/com/sun/javafx/font/PrismFontFactory.java
```diff
@@ -1024,6 +1024,10 @@ private static void getPlatformFontDirs() {
...
 
+        // Add Arial Unicode to Windows fallback list
+        fontRegInfo[0].add(getPathNameWindows("arialuni.ttf"));
+        fontRegInfo[1].add("Arial Unicode MS");
```
kevinrushforth: This seems fragile to me.

modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/java/ComplexTextControllerJava.cpp
```diff
@@ -92,6 +92,10 @@ unsigned jGetEnd(jobject jRun)
 
 unsigned jGetCharOffset(jobject jRun, unsigned glyphIndex)
 {
+    if (!jGetGlyphCount(jRun)) {
+        return { };
```
kevinrushforth: Shouldn't this return an explicit (perhaps an out-of-band) value?

PR #337, opened 2019-01-02
JDK-8215799: Complex text is not rendered by webkit on Windows

ararunprasad:
WebKit uses GlyphPage table to quickly decide whether any glyph is available for given character. For our JavaFX port, we get char to glyph mapping using CharToGlyphMapper.

On macOS and Linux it is working because the LogicalFont had valid fallback for complex texts also, but it is missing for Windows. Refer PrismFontFactory.getLinkedFonts.

606.1 WebKit used GlyphTable only for simple texts, however 607.1 uses the same for complex texts also.

I could think of the following workarounds.,

1. Add "Arial Unicode MS Regular" as a fallback for Windows, it has almost all glyphs for the unicode range.
2. Call "TextLayout.getRuns" before invoking "CharToGlyphMapper", "TextLayout.getRuns" finds & adds the native fallback font.


ararunprasad:
This PR is based on approach 2.

kevinrushforth:
Link to JBS bug: https://bugs.openjdk.java.net/browse/JDK-8215799

modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/java/ComplexTextControllerJava.cpp
```diff
@@ -92,6 +92,10 @@ unsigned jGetEnd(jobject jRun)
 
 unsigned jGetCharOffset(jobject jRun, unsigned glyphIndex)
 {
+    if (!jGetGlyphCount(jRun)) {
+        return { };
```
kevinrushforth: Shouldn't this return an explicit (perhaps an out-of-band) value?

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/WCFontImpl.java
```diff
@@ -126,6 +126,9 @@ private FontStrike getFontStrike()
     }
 
     @Override public int[] getGlyphCodes(char[] chars) {
+        // Calling TextLayout.getRuns will lead to loading correct fallback
+        // font from native font engine.
+        TextUtilities.createLayout(new String(chars), getPlatformFont()).getRuns();
```
kevinrushforth: Will there be a performance impact (e.g., will this cause all text strings to be laid out twice?)

modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/java/ComplexTextControllerJava.cpp
ararunprasad: Sure, I think it is better to return the same value as `TextRun.getCharOffset`. i.e it just returns the `glyphIndex`.

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/WCFontImpl.java
ararunprasad: There won't be any performance impact because the above method only be called by the WebCore::GlyphPage for a given character sequence

> Will there be a performance impact (e.g., will this cause all text strings to be laid out twice?)

There won't be any performance impact because the above method only be called by the WebCore::GlyphPage for a given character sequence. It is mainly used to check whether the font supports the given unicode char point, and based on that it will decide how to break the text. Also this method will be called only once during the entire life of WebKit for the char sequence + Font + size + weight combination.

muralibilla:
looks fine..

ararunprasad:
@kevinrushforth , should I wait for Phil's review comments?

modules/javafx.graphics/src/main/java/com/sun/javafx/font/CompositeGlyphMapper.java
```diff
@@ -100,7 +100,6 @@ private final int convertToGlyph(int unicode) {
                 return glyphCode;
             }
         }
-        glyphMap.put(unicode, missingGlyph);
```
kevinrushforth: Does removing this entry from the map have any side effects?

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/WCFontImpl.java
```diff
@@ -125,10 +125,24 @@ private FontStrike getFontStrike()
         return getFontStrike().getMetrics().getXHeight();
     }
 
+    private static boolean needsTextLayout(final int glyphs[]) {
+        for (int g : glyphs) {
+            if (g != 0) {
```
prrace: So if *any* glyph is present, you don't use layout.
I was expecting the other way round, if any glyph is missing you want to try layout ..

modules/javafx.graphics/src/main/java/com/sun/javafx/font/CompositeGlyphMapper.java
ararunprasad: AFAIK, it shouldn't have side effect. @prrace , what is your opinion?

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/WCFontImpl.java
kevinrushforth: I presume it was intentional, given the comment at the usage site, but it seems better to do as Phil suggests.

ararunprasad: Sure. I will change according to @prrace's suggestion.

PR #341, opened 2019-01-08
JDK-8211454: Update SQLite to version 3.26.0

ararunprasad:
We currently use SQLite version 3.23.0. We should update to the latest stable release version 3.26.0 
https://www.sqlite.org/index.html

This PR is targeted for https://bugs.openjdk.java.net/browse/JDK-8211454.


muralibilla:
looks fine

PR #342, opened 2019-01-08
JDK-8216377: Fix memoryleak for initial nodes of Window

FlorianKirmaier:
BugID: [JDK-8216377](https://bugs.openjdk.java.net/browse/JDK-8216377)

While working on an update for [JPro](https://www.jpro.one/), I've found out, that a unit test for a memory leak doesn't work on JavaFX11. It's a regression and doesn't happen on JavaFX8

The initial nodes of a Scene are not collected by the GC when they are removed from the Scene.
They get collected when the initial Window is closed, but this is too late and not the expected behavior.

Affected JavaFX versions: 9 10 11 develop

This happens because the listener windowShowingChangedListener is not removed from the old Window, when the Scene is Changed.
This commit makes sure, that this listener is properly removed.

This bug would create a memory leak for every node inside a window if there wouldn't be another bug.
The listener windowShowingChangedListener is only added to the Scene when the window is changed.
I'm pretty sure, this is not intentionally and also fixed with this commit.
This might fix an unknown rendering related bug.

This commit also contains an unit-test to make sure no regression will happen.

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2019, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -1080,7 +1080,7 @@ public abstract class Node implements EventTarget, Styleable {
             oldScene.windowProperty().removeListener(sceneWindowChangedListener);
 
             Window window = oldScene.windowProperty().get();
-            if(window != null) {
+            if (window != null) {
                 window.showingProperty().removeListener(windowShowingChangedListener);
             }
         }
@@ -1088,7 +1088,7 @@ public abstract class Node implements EventTarget, Styleable {
             newScene.windowProperty().addListener(sceneWindowChangedListener);
 
             Window window = newScene.windowProperty().get();
-            if(window != null) {
+            if (window != null) {
                 window.showingProperty().addListener(windowShowingChangedListener);
             }
 
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -1078,9 +1078,20 @@ public abstract class Node implements EventTarget, Styleable {
         // isTreeShowing needs to take into account of Window's showing
         if (oldScene != null) {
             oldScene.windowProperty().removeListener(sceneWindowChangedListener);
+
+            Window window = oldScene.windowProperty().get();
+            if(window != null) {
+                window.showingProperty().removeListener(windowShowingChangedListener);
+            }
         }
         if (newScene != null) {
             newScene.windowProperty().addListener(sceneWindowChangedListener);
+
+            Window window = newScene.windowProperty().get();
+            if(window != null) {
+                window.showingProperty().addListener(windowShowingChangedListener);
+            }
+
         }
         updateTreeShowing();
 
--- a/tests/system/src/test/java/test/javafx/scene/InitialNodesMemoryLeakTest.java
+++ b/tests/system/src/test/java/test/javafx/scene/InitialNodesMemoryLeakTest.java
@@ -31,30 +31,33 @@ import javafx.scene.Group;
 import javafx.scene.Scene;
 import javafx.stage.Stage;
 import javafx.stage.WindowEvent;
+
 import junit.framework.Assert;
+import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import test.util.Util;
+import static org.junit.Assert.fail;
 
 import java.lang.ref.WeakReference;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
-import static org.junit.Assert.fail;
 
 public class InitialNodesMemoryLeakTest {
 
     static CountDownLatch startupLatch;
-    static WeakReference<Group> toCleanUp;
+    static WeakReference<Group> groupWRef;
+    static Stage stage;
 
     public static class TestApp extends Application {
 
         @Override
-        public void start(Stage stage) throws Exception {
-            toCleanUp = new WeakReference<>(new Group());
-
-            Group root = new Group(toCleanUp.get());
-
+        public void start(Stage primaryStage) throws Exception {
+            stage = primaryStage;
+            Group group = new Group();
+            groupWRef = new WeakReference<>(group);
+            Group root = new Group(group);
             stage.setScene(new Scene(root));
 
             stage.addEventHandler(WindowEvent.WINDOW_SHOWN, e -> {
@@ -63,9 +66,7 @@ public class InitialNodesMemoryLeakTest {
                     startupLatch.countDown();
                 });
             });
-
             stage.show();
-
         }
     }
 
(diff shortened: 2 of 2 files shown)


kevinrushforth:
@FlorianKirmaier - Welcome to OpenJFX, and thank you for reporting this bug, and for proposing a solution. Before we can review it, we need a signed [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html) (OCA) on file from you. This can take several days, so if you have already sent one in, there is nothing else to do but wait until you hear back about your OCA.

FlorianKirmaier:
I've just signed the OCA. Now it probably just takes some time.

kevinrushforth:
> I've just signed the OCA. Now it probably just takes some time.

In the mean time, I'll provide some quick feedback.

FlorianKirmaier:
I've added a commit according to your feedback.

I did only run the unit-tests provided in the openjfx project. 
I couldn't find any regression.

tests/system/src/test/java/test/javafx/scene/InitialNodesMemoryLeakTest.java
```diff
@@ -0,0 +1,102 @@
...
+    static CountDownLatch startupLatch;
+    static WeakReference<Group> toCleanUp;
+
+    public static class TestApp extends Application {
+
+        @Override
+        public void start(Stage stage) throws Exception {
+            toCleanUp = new WeakReference<>(new Group());
+
+            Group root = new Group(toCleanUp.get());
```
kevinrushforth: There is a very small (but non-zero) window where the newly created child Group could be garbage collected between the time it goes out of scope (right after adding it as the referent of `toCleanUp`) and the call to `toCleanUp().get`. I recommend holding the new Group a local variable temporarily. Something like this:

```
    Group group = new Group();
    toCleanUp = new WeakReference<>(group);
    Group root = new Group(group);
```

You could also set `group = null` after creating the` root`, if you want, but it isn't necessary, since `group` will go out of scope when `start()` returns, leaving the children list of `root` as the only hard reference to the object (until you clear it).

tests/system/src/test/java/test/javafx/scene/InitialNodesMemoryLeakTest.java
```diff
@@ -0,0 +1,102 @@
...
+                fail("Timeout waiting for FX runtime to start");
+            }
+        } catch (InterruptedException ex) {
+            fail("Unexpected exception: " + ex);
+        }
+    }
+
+    @Test
+    public void testCleanup() throws Exception {
+        System.out.println("got: " + toCleanUp.get());
```
kevinrushforth: This will show up in the log and in the console if you run with `gradle --info`. Since it is a debugging aid, I recommend to remove it.

tests/system/src/test/java/test/javafx/scene/InitialNodesMemoryLeakTest.java
```diff
@@ -0,0 +1,102 @@
...
+            if (!startupLatch.await(15, TimeUnit.SECONDS)) {
+                fail("Timeout waiting for FX runtime to start");
+            }
+        } catch (InterruptedException ex) {
+            fail("Unexpected exception: " + ex);
+        }
+    }
+
+    @Test
+    public void testCleanup() throws Exception {
```
arapte: I would recommend to use a name like testRootNodeMemoryLeak() and 
similarly the file name can be changed as, RootNodeMemoryLeakTest.java

tests/system/src/test/java/test/javafx/scene/InitialNodesMemoryLeakTest.java
```diff
@@ -0,0 +1,102 @@
...
+            System.runFinalization();
+
+            if (toCleanUp.get() == null) {
+                break;
+            }
+
+            Util.sleep(500);
+        }
+
+        Assert.assertTrue("Couldn't collect Node", toCleanUp.get() == null);
```
arapte: Here assertNull() could be used instead of assertTrue()
Assert.assertNull("toCleanUp should get GCed.", toCleanUp.get());

tests/system/src/test/java/test/javafx/scene/InitialNodesMemoryLeakTest.java
```diff
@@ -0,0 +1,102 @@
...
+                    startupLatch.countDown();
+                });
+            });
+
+            stage.show();
+
+        }
+    }
+
+    @BeforeClass
```
arapte: Please include a @ AfterClass method for cleanly exiting the test,
Something like,
@ AfterClass
public static void teardownOnce() {
    stage.hide(); //optional
    Platform.exit();
}

tests/system/src/test/java/test/javafx/scene/InitialNodesMemoryLeakTest.java
```diff
@@ -0,0 +1,102 @@
...
+        System.out.println("got: " + toCleanUp.get());
+
+        for (int j = 0; j < 10; j++) {
+            System.gc();
+            System.runFinalization();
+
+            if (toCleanUp.get() == null) {
+                break;
+            }
+
```
arapte: Some of the empty lines can be removed. like line# 55, 57, 66, 68

tests/system/src/test/java/test/javafx/scene/InitialNodesMemoryLeakTest.java
FlorianKirmaier: It's about being the initial node, instead of being the root node. So the name should be ok.

tests/system/src/test/java/test/javafx/scene/InitialNodesMemoryLeakTest.java
```diff
@@ -0,0 +1,102 @@
...
+import java.lang.ref.WeakReference;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.fail;
+
+public class InitialNodesMemoryLeakTest {
+
+    static CountDownLatch startupLatch;
+    static WeakReference<Group> toCleanUp;
```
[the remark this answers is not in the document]
FlorianKirmaier: I take the name groupWRef for now, because rootWRef isn't correct - it's not a root.

tests/system/src/test/java/test/javafx/scene/InitialNodesMemoryLeakTest.java
FlorianKirmaier: That's a great Idea! I've changed the code accordingly.

FlorianKirmaier:
Thank you for the feedback! 
I've changed the code according to your feedback.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -1078,9 +1078,20 @@ private void invalidatedScenes(Scene oldScene, SubScene oldSubScene) {
...
+            }
         }
         if (newScene != null) {
             newScene.windowProperty().addListener(sceneWindowChangedListener);
+
+            Window window = newScene.windowProperty().get();
+            if (window != null) {
+                window.showingProperty().addListener(windowShowingChangedListener);
+            }
+
```
arapte: I missed this in earlier feedback, Please remove this empty line # 1094 along with any further update.

kevinrushforth:
@FlorianKirmaier your OCA has been recorded. I will finish my review and merge the PR when done testing.

Thank you for your contribution.

kevinrushforth:
Replying to a comment in the initial PR:

>  This might fix an unknown rendering related bug.

Indeed it does. I was able to confirm that your fix also fixes the following rending issue:

[JDK-8207837](https://bugs.openjdk.java.net/browse/JDK-8207837): Indeterminate ProgressBar does not animate if content is added after scene is set on window

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
kevinrushforth: Since there was no further update, and I wanted to get this in for openjfx12, I merged it without waiting for the blank line removal.

FlorianKirmaier:
@kevinrushforth
Thank you for merging. Is it going be backported to jfx-11? Is there anything I can do about it?

kevinrushforth:
> Is it going be backported to jfx-11?

Not by default, no. In general we only backport critical bugs. It requires approval from a Project Lead (either @johanvos or myself) to get this backported to FX 11. In this instance, I would defer to Johan, so depending on what he says, you can send a formal request to openjfx-dev mailing list asking for approval to backport it.

PR #346, opened 2019-01-15
JDK-8217333: Pass additional arguments to gradle from build.ps1

brcolow:
If there are any extra parameters passed in to .\tools\scripts\build.ps1
that are not used by the script itself, then pass any additional
parameters through to gradle.

This allows for the following:

.\tools\scripts\build.ps1 -nocygwin --tests *SomeTest

Where the "--tests *SomeTest" bit is actually an argument for gradle,
not the powershell build script.

### DIFF
--- a/tools/scripts/build.ps1
+++ b/tools/scripts/build.ps1
@@ -1,4 +1,7 @@
-param ([switch]$nocygwin = $false)
+param (
+[switch]$nocygwin = $false,
+[parameter(ValueFromRemainingArguments)][String[]]$args
+)
 
 choco install ant
 choco install vswhere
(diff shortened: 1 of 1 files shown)


brcolow:
I'll file an RFE soon-ish (next couple days max).

brcolow:
RFE filed: https://bugs.openjdk.java.net/browse/JDK-8217333 and review request send to openjfx-dev mailing list.

kevinrushforth:
This seems like a good fix for openjfx13 (we'll fork the repo soon and then I can take a look).

brcolow:
Thanks Kevin!

PR #347, opened 2019-01-16
JDK-8215702: SVG gradients are not rendered

ararunprasad:
Gradient points are not mapped to the current transformation matrix. The fix was there before 606.1 upgrade, however during the merge it got missed out while incorporating the Gradient class related refactoring changes.


muralibilla:
looks fine..

PR #353, opened 2019-01-21
JDK-8205092: Added dirty check for viewOrderChildren

DeanWookey:
If the scene graph is modified while handling a mouse event, viewOrderChildren can contain nodes which are no longer in the scene graph, causing an NPE in PickResultChooser.processOffer. This is fairly rare, and only affects parent nodes with children that have a view order set. testPickingChildNodeWithDirtyViewOrder replicates these conditions and results in the same NPE as reported in the bug report.

This solution doesn't unset the dirty bits because I'm unsure of the consequences. In the worst case the viewOrderChildren will be recalculated one extra time because of this.

An alternative solution is to keep viewOrderChildren in sync with the children list, however that would mean sorting it whenever the view order property of a child is changed (the child calls markViewOrderChildrenDirty of its parent when this happens). That could happen many times. Updating the list as it's needed seems like a better solution to me.

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
@@ -194,7 +194,8 @@ public abstract class Parent extends Node {
         }
 
         if (isDirty(DirtyBits.PARENT_CHILDREN_VIEW_ORDER)) {
-            computeViewOrderChildrenAndUpdatePeer();
+            computeViewOrderChildren();
+            peer.setViewOrderChildren(viewOrderChildren);
         }
 
         if (Utils.assertionEnabled()) validatePG();
@@ -265,7 +266,7 @@ public abstract class Parent extends Node {
         NodeHelper.markDirty(this, DirtyBits.PARENT_CHILDREN_VIEW_ORDER);
     }
 
-    private void computeViewOrderChildrenAndUpdatePeer() {
+    private void computeViewOrderChildren() {
         boolean viewOrderSet = false;
         for (Node child : children) {
             double vo = child.getViewOrder();
@@ -284,9 +285,6 @@ public abstract class Parent extends Node {
                     -> a.getViewOrder() < b.getViewOrder() ? 1
                             : a.getViewOrder() == b.getViewOrder() ? 0 : -1);
         }
-
-        final NGGroup peer = getPeer();
-        peer.setViewOrderChildren(viewOrderChildren);
     }
 
     // Call this method if children view order is needed for picking.
@@ -294,7 +292,7 @@ public abstract class Parent extends Node {
     private List<Node> getOrderedChildren() {
         if (isDirty(DirtyBits.PARENT_CHILDREN_VIEW_ORDER)) {
             //Fix for JDK-8205092
-            computeViewOrderChildrenAndUpdatePeer();
+            computeViewOrderChildren();
         }
         if (!viewOrderChildren.isEmpty()) {
             return viewOrderChildren;
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2019, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -292,6 +292,10 @@ public abstract class Parent extends Node {
     // Call this method if children view order is needed for picking.
     // The returned list should be treated as read only.
     private List<Node> getOrderedChildren() {
+        if (isDirty(DirtyBits.PARENT_CHILDREN_VIEW_ORDER)) {
+            //Fix for JDK-8205092
+            computeViewOrderChidrenAndUpdatePeer();
+        }
         if (!viewOrderChildren.isEmpty()) {
             return viewOrderChildren;
         }
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
@@ -194,7 +194,7 @@ public abstract class Parent extends Node {
         }
 
         if (isDirty(DirtyBits.PARENT_CHILDREN_VIEW_ORDER)) {
-            computeViewOrderChidrenAndUpdatePeer();
+            computeViewOrderChildrenAndUpdatePeer();
         }
 
         if (Utils.assertionEnabled()) validatePG();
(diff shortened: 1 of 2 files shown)


kevinrushforth:
Link to JBS bug: https://bugs.openjdk.java.net/browse/JDK-8205092

kevinrushforth:
I'll take a look.

modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
```diff
@@ -741,6 +743,101 @@ public void sceneListenerCanAddChild() {
         assertSame(scene, ParentShim.getChildren(child).get(3).getScene());
     }
 
+    @Test
+    public void testPickingChildNode(){
```
kevinrushforth: Minor: add a space before `{`

modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
```diff
@@ -741,6 +743,101 @@ public void sceneListenerCanAddChild() {
...
+        rect2.setY(10);
+        rect2.setWidth(100);
+        rect2.setHeight(100);
+        Group g = new Group();
+
+        // needed since picking doesn't work unless rooted in a scene and visible
+        Scene scene = new Scene(g);
+        stage.setScene(scene);
+        stage.show();
+        ParentShim.getChildren(g).addAll(rect1,rect2);
```
kevinrushforth: Minor: add space after the comma.

modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
```diff
@@ -741,6 +743,101 @@ public void sceneListenerCanAddChild() {
...
+        PickResultChooser res = new PickResultChooser();
+        NodeHelper.pickNode(g, new PickRay(50, 50, 1, Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY), res);
+        assertSame(rect2, res.getIntersectedNode());
+        res = new PickResultChooser();
+        NodeHelper.pickNode(g, new PickRay(0, 0, 1, Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY), res);
+        assertNull(res.getIntersectedNode());
+    }
+
+    @Test
+    public void testPickingChildNodeWithViewOrderSet(){
```
kevinrushforth: Minor: add space before `{`

modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
```diff
@@ -741,6 +743,101 @@ public void sceneListenerCanAddChild() {
...
+        rect2.setY(10);
+        rect2.setWidth(100);
+        rect2.setHeight(100);
+        Group g = new Group();
+
+        // needed since picking doesn't work unless rooted in a scene and visible
+        Scene scene = new Scene(g);
+        stage.setScene(scene);
+        stage.show();
+        ParentShim.getChildren(g).addAll(rect1,rect2);
```
kevinrushforth: Minor: add space after comma

modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
```diff
@@ -741,6 +743,101 @@ public void sceneListenerCanAddChild() {
...
+        PickResultChooser res = new PickResultChooser();
+        NodeHelper.pickNode(g, new PickRay(50, 50, 1, Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY), res);
+        assertSame(rect1, res.getIntersectedNode());
+        res = new PickResultChooser();
+        NodeHelper.pickNode(g, new PickRay(0, 0, 1, Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY), res);
+        assertNull(res.getIntersectedNode());
+    }
+
+    @Test
+    public void testPickingChildNodeWithDirtyViewOrder(){
```
kevinrushforth: Minor: add space before `{`

modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
```diff
@@ -741,6 +743,101 @@ public void sceneListenerCanAddChild() {
...
+        rect2.setY(10);
+        rect2.setWidth(100);
+        rect2.setHeight(100);
+        Group g = new Group();
+
+        // needed since picking doesn't work unless rooted in a scene and visible
+        Scene scene = new Scene(g);
+        stage.setScene(scene);
+        stage.show();
+        ParentShim.getChildren(g).addAll(rect1,rect2);
```
kevinrushforth: Minor: add space after comma.

kevinrushforth:
I tagged @aghaisas as a second reviewer.

DeanWookey:
I made the requested changes and fixed a spelling error in computeViewOrderChildrenAndUpdatePeer as it seemed like a good opportunity.

kevinrushforth:
I missed this when initially reviewing it, and only saw it when I was double-checking something related to the updated list of children in the peer.

The problem with the fix as currently proposed is that the `Parent::computeViewOrderChidrenAndUpdatePeer` method combines combines the "compute ordered children" operation with the "update the ordered children in the peer" operation. The latter must only be called from the `Scene.ScenePulseListener::pulse` while holding the render lock (all of the `doUpdatePeer`methods in the nodes are only ever called with the lock held, so the existing usage is fine). Performing the "update peer" operation at any other time is not thread-safe.

We will need an updated fix that splits  "compute ordered children" and "update the ordered children in the peer" into two parts, with only "compute ordered children" being called from `getOrderedChildren`. One way to do this would be to rename the existing method to `computeViewOrderChildren` and add a `boolean updatePeer` argument which would be called with true from `doUpdatePeer` and `false` from `getOrderedChildren`. Or you could just split it into two methods and have `doUpdatePeer` call both, while `getOrderedChildred` would only call the compute method.

modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
```diff
@@ -292,6 +292,10 @@ private void computeViewOrderChidrenAndUpdatePeer() {
     // Call this method if children view order is needed for picking.
     // The returned list should be treated as read only.
     private List<Node> getOrderedChildren() {
+        if (isDirty(DirtyBits.PARENT_CHILDREN_VIEW_ORDER)) {
+            //Fix for JDK-8205092
+            computeViewOrderChildrenAndUpdatePeer();
```
kevinrushforth: As mentioned in my earlier comment, this needs to be changed as it is not thread-safe.

DeanWookey: Thanks for the detailed explanation. I went with the second option, although I didn't make a second method to update the peer's view order since it was a one-liner.

kevinrushforth:
I'll take a closer look and run some tests, but at a quick glance, the latest changes look fine.

DeanWookey:
To avoid double (or perhaps more in extreme cases) computing the viewOrderChildren when this error would occur, we could store an additional flag in the markViewOrderChildrenDirty() method to store a separate dirty flag for the node's viewOrderChildren list only. computeViewOrderChildren() could check and set that variable. 

Like I mentioned before, I'm not sure that optimisation is worth it, but it could be done relatively easily if you think it's worth it.

PR #355, opened 2019-01-22
JDK-8213806: WebView - JVM crashes for given HTML

ararunprasad:
### Root cause
"baseURL" is set to empty(null) when loading the html string using `WebEngine.loadContent` API. The URL has been used as one of the key while caching the inline CSS style content for Shadow Root(Shadow DOM). While calculating the hash value for the key, `URLHash::hash` method calls the `StringImpl::hash` method without checking for null.

It is not observed on other WebKit ports because they expose loadContent API which also takes baseURL. See [WKWebView::loadHTMLString](https://developer.apple.com/documentation/webkit/wkwebview/1415004-loadhtmlstring?language=objc).

We also should consider to provide another variant of `WebEngine.loadContent` which takes `baseURL` as a parameter.

### Solution
Use to `String::hash` instead of `StringImpl::hash` from `URLHash::hash`. `String::hash` still calls `StringImpl::hash` after testing for null.

### DIFF
--- a/modules/javafx.web/src/main/native/Source/WebCore/platform/URLHash.h
+++ b/modules/javafx.web/src/main/native/Source/WebCore/platform/URLHash.h
@@ -35,7 +35,7 @@ namespace WebCore {
     struct URLHash {
         static unsigned hash(const URL& key)
         {
-            return key.string().impl()->hash();
+            return key.string().hash();
         }
 
         static bool equal(const URL& a, const URL& b)
--- a/modules/javafx.web/src/test/java/test/javafx/scene/web/MiscellaneousTest.java
+++ b/modules/javafx.web/src/test/java/test/javafx/scene/web/MiscellaneousTest.java
@@ -497,4 +497,31 @@ public class MiscellaneousTest extends TestBase {
             assertTrue("Color should be opaque yellow:" + pixelAt399x145, isColorsSimilar(Color.YELLOW, pixelAt399x145, 1));
         });
     }
+
+    @Test public void testShadowDOMWithLoadContent() {
+        loadContent("<html>\n" +
+                    "  <body>\n" +
+                    "    <template id='element-details-template'>\n" +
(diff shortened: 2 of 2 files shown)


ararunprasad:
I tried to reproduce the same using macOS SDK's WebKit API (from WebKit framework), it is not reproducible because macOS port of WebKit generates unique `applewebdata://` when `baseURL` is set to `nil`. Below is the Xcode playground code,
```
  import WebKit
  import PlaygroundSupport
  
  let webView = WebView(frame: CGRect(x: 0, y: 0, width: 1200.0, height: 1200.0))
  let mainFrame = webView.mainFrame
  mainFrame?.loadHTMLString(
    "<html>\n" +
        "  <body>\n" +
    "    <template id='element-details-template'>\n" +
    "      <style>\n" +
    "        p { font-weight: bold; }\n" +
    "      </style>\n" +
    "    </template>\n" +
    "\n" +
    "    <element-details>\n" +
    "    </element-details>\n" +
    "\n" +
    "    <script>\n" +
    "    customElements.define('element-details',\n" +
    "      class extends HTMLElement {\n" +
    "        constructor() {\n" +
    "          super();\n" +
    "          const template = document\n" +
    "            .getElementById('element-details-template')\n" +
    "            .content;\n" +
    "          const shadowRoot = this.attachShadow({mode: 'open'})\n" +
    "            .appendChild(template.cloneNode(true));\n" +
    "        }\n" +
    "      })\n" +
        "    document.body.innerText=(document.location);" +
    "    </script>\n" +
    "    hello\n" +
    "  </body>\n" +
    "</html>\n"
, baseURL: nil)

  PlaygroundPage.current.liveView = webView
```

muralibilla:
seems like safe fix as  String calls StringImpl in case of not null...

unsigned hash() const { return isNull() ? 0 : impl()->hash(); }

PR #363, opened 2019-01-26
JDK-8201539: WICCreateImagingFactory: defer CoUninitialize() call to thread exit (native approach)

pyokagan:
GitHub issue: #66 
JBS Bug Link: https://bugs.openjdk.java.net/browse/JDK-8201539

The crash was occurring because the call to `CoUninitialize()` was done too early and resulted in `WindowsCodecs.dll` being prematurely unloaded even though `IWICImagingFactory` needed it. This caused a crash when `IWICImagingFactory` was accessed afterwards.

This is the "native approach" that uses a fibre local storage slot to call a destructor on thread exit.

### DIFF
--- a/modules/javafx.graphics/src/main/native-font/directwrite.cpp
+++ b/modules/javafx.graphics/src/main/native-font/directwrite.cpp
@@ -845,6 +845,13 @@ jobject newD2D1_MATRIX_3X2_F(JNIEnv *env, D2D1_MATRIX_3X2_F *lpStruct)
 /*                                                                        */
 /**************************************************************************/
 
+static void WICImagingFactoryFlsCallback(void *data)
+{
+    IWICImagingFactory *iwicImagingFactory = (IWICImagingFactory *)data;
+    iwicImagingFactory->Release();
+    CoUninitialize();
+}
+
 JNIEXPORT jlong JNICALL OS_NATIVE(_1WICCreateImagingFactory)
     (JNIEnv *env, jclass that)
 {
@@ -866,10 +873,20 @@ JNIEXPORT jlong JNICALL OS_NATIVE(_1WICCreateImagingFactory)
             CLSCTX_INPROC_SERVER,
             IID_PPV_ARGS(&result)
             );
+    if (!SUCCEEDED(hr)) goto fail;
 
-    /* Unload COM as no other COM objects will be create directly */
+    /* Setup a fibre local storage slot to call WICImagingFactoryFlsCallback()
+     * with the created IWICImagingFactory on thread exit. */
+    DWORD flsIndex = FlsAlloc(WICImagingFactoryFlsCallback);
+    if (flsIndex == FLS_OUT_OF_INDEXES) goto fail;
+    FlsSetValue(flsIndex, result);
+
+    return (jlong)result;
+
+fail:
+    if (result) result->Release();
     CoUninitialize();
-    return SUCCEEDED(hr) ? (jlong)result : NULL;
+    return NULL;
 }
 
 JNIEXPORT jlong JNICALL OS_NATIVE(_1D2D1CreateFactory)


brcolow:
For reviewers: 

The fiber local storage API is only being used in it's thread-local storage capacity because no fibers are explicitly created (no `ConvertThreadToFiber` and `CreateFiber` calls). Because the thread-local storage API is older and harder to work with (not so easy to set a callback on thread exit), and because on newer versions of Windows TLS functions are in-lined to FLS functions, I think it's a wise choice and not too exotic of an API.

modules/javafx.graphics/src/main/native-font/directwrite.cpp
```diff
@@ -845,6 +845,13 @@ jobject newD2D1_MATRIX_3X2_F(JNIEnv *env, D2D1_MATRIX_3X2_F *lpStruct)
 /*                                                                        */
 /**************************************************************************/
 
+static void WICImagingFactoryFlsCallback(void *data)
+{
+    IWICImagingFactory *iwicImagingFactory = (IWICImagingFactory *)data;
+    iwicImagingFactory->Release();
```
kevinrushforth: I don't think this is correct. According to the documentation for `CoCreateInstance`, it already does the Release so it should not be done again. Can you comment on this?

modules/javafx.graphics/src/main/native-font/directwrite.cpp
```diff
@@ -866,10 +873,20 @@ JNIEXPORT jlong JNICALL OS_NATIVE(_1WICCreateImagingFactory)
...
             IID_PPV_ARGS(&result)
             );
+    if (!SUCCEEDED(hr)) goto fail;
 
-    /* Unload COM as no other COM objects will be create directly */
+    /* Setup a fibre local storage slot to call WICImagingFactoryFlsCallback()
+     * with the created IWICImagingFactory on thread exit. */
+    DWORD flsIndex = FlsAlloc(WICImagingFactoryFlsCallback);
+    if (flsIndex == FLS_OUT_OF_INDEXES) goto fail;
+    FlsSetValue(flsIndex, result);
```
kevinrushforth: I think you can skip this call, given that there is no need to use it to call Release in the callback method.

modules/javafx.graphics/src/main/native-font/directwrite.cpp
```diff
@@ -866,10 +873,20 @@ JNIEXPORT jlong JNICALL OS_NATIVE(_1WICCreateImagingFactory)
...
+
+    return (jlong)result;
+
+fail:
+    if (result) result->Release();
```
kevinrushforth: See my earlier comment. I don't think calling Release is the right thing to do.

modules/javafx.graphics/src/main/native-font/directwrite.cpp
pyokagan: I guess you are referring to this paragraph in the documentation?

> The CoCreateInstance function provides a convenient shortcut by connecting to the class object associated with the specified CLSID, creating an uninitialized instance, and releasing the class object. As such, it encapsulates the following functionality:
> ```
> CoGetClassObject(rclsid, dwClsContext, NULL, IID_IClassFactory, &pCF); 
> hresult = pCF->CreateInstance(pUnkOuter, riid, ppvObj) 
> pCF->Release(); 
> ```

From my understanding, what it does is that it releases the *class factory* (`pCF`). The instance created with `pCF->CreateInstance` still has a non-zero reference count and so `Release()` still needs to be called on it.

The code examples in https://docs.microsoft.com/en-us/windows/desktop/learnwin32/managing-the-lifetime-of-an-object show `Release()` being called on objects created with `CoCreateInstance`, so it should be correct.

kevinrushforth: Yes, I think you are correct.

kevinrushforth:
Let's proceed with the dispose hook approach as implemented in PR #367

Issue #339, opened 2019-01-07
JDK-8217605: Add support for e-paper displays


jgneff:
## Report Classification

**Issue Type:** Request for Enhancement  
**Component:** JavaFX  
**Subcomponent:** *graphics:* JavaFX Graphics  
**Operating System:** Ubuntu  
**Java Release:** 12  

### Synopsis

Add support for e-paper displays.

#### Summary

This is a request for JavaFX to support devices with electrophoretic displays, commonly known as electronic paper displays, with a new Monocle platform called “EPD.”

#### Goals

The ultimate goal of this request is to be able to run a JavaFX application on a device with an e-paper display by simply installing the Ubuntu OpenJFX package for the *armhf* architecture. That goal includes:

* making it easy to develop application software for e-paper that can also run on desktop platforms without having to recompile or modify the source code in any way;

* having a method to enable the black-and-white animation capabilities of e-paper, with frame rates up to eight frames per second (120 ms per frame).

#### Non-Goals

The following are specifically out of scope for this request:

* It is not a goal to allow any JavaFX application to work well on e-paper displays. These displays have inherent limits in grayscale levels and frame rates, so some applications written for LCD screens may never be appropriate for them.

* It is not a goal to provide a JavaFX programming interface to the lower-level features of e-paper. These features can be selected based on the application by setting specific system properties when launching it. Allowing for runtime changes to the system properties could be added later.

* It is not a goal to support all classes of devices with e-paper displays, but rather to start with just one. That first implementation can form the basis for others to be added later.

#### Success Metrics

One measure of success would be running a JavaFX animation program on an e-paper display at eight frames per second and delivering the correct touch screen input events to the program.

#### Motivation

The motivation for this request comes from a work of art by Joseph Farbrook exhibited at the SIGGRAPH 2014 Art Gallery and part of a series called [*Lineographs*](http://s2014.siggraph.org/attendees/art-gallery/events/lineographs.html "Lineographs at SIGGRAPH 2014"). See the artist’s website for a brief description of the work and a [one-minute video](https://www.farbrook.net/lineographs/ "Dancer (with Micaela Gardner), by Joseph Farbrook") demonstration. The video is also available directly on YouTube as [*Dancer*](https://youtu.be/nCKPZuK8XyI "Dancer (with Micaela Gardner), by Joseph Farbrook").

[...]

PR #369, opened 2019-02-04
JDK-8217605: Add support for e-paper displays

jgneff:
Closes javafxports/openjdk-jfx#339


jgneff:
My new unit tests succeeded, but the following unrelated test failed:
```
test.javafx.concurrent.ServiceLifecycleTest > cancelCalledFromOnSucceeded FAILED
    java.lang.AssertionError: expected:<SUCCEEDED> but was:<CANCELLED>
    at test.javafx.concurrent.ServiceLifecycleTest
        .cancelCalledFromOnSucceeded(ServiceLifecycleTest.java:1176)
```

The test case is:
```Java
@Test public void cancelCalledFromOnSucceeded() {
    final AtomicInteger cancelNotificationCount = new AtomicInteger();
    service.addEventFilter(WorkerStateEvent.WORKER_STATE_SUCCEEDED, workerStateEvent -> {
        service.cancel();
    });
    service.addEventFilter(WorkerStateEvent.WORKER_STATE_CANCELLED, event -> {
        cancelNotificationCount.incrementAndGet();
    });

    service.start();
    executor.executeScheduled();
    task.complete();
    assertEquals(Worker.State.SUCCEEDED, service.getState());
    assertEquals(0, cancelNotificationCount.get());
}
```

Any ideas? I think it worked when it ran for the previous commit, and they all run successfully on my build machine.

kevinrushforth:
JBS bug ID: https://bugs.openjdk.java.net/browse/JDK-8217605

johanvos:
This PR builds fine when using the armv6hf build, and it doesn't break the existing functionality. I don't have an e-paper device so I can't test it, but the code takes a similar approach as the other monocle implementations.
A drawback is that the EPD classes will show up in the javafx.graphics.jar for other implementations, but this is a more general issue we probably need to address with monocle.

modules/javafx.graphics/src/test/java/test/com/sun/glass/ui/monocle/FramebufferY8Test.java
```diff
@@ -0,0 +1,167 @@
...
+     * test cases. It loads the image into the buffer as 32-bit pixels in ARGB32
+     * format.
+     *
+     * @throws IOException if an error occurs loading the image
+     */
+    @BeforeClass
+    public static void onlyOnce() throws IOException {
+        InputStream stream = FramebufferY8Test.class.getResourceAsStream(ALL_24BIT_RGB);
+        BufferedImage image = ImageIO.read(stream);
+        width = image.getWidth();
```
johanvos: running the test gives me a Nullpointerexception here -- the resource is not on the classpath?

jgneff: Thanks, Johan. When I run the graphics tests from scratch on my build machine (after _cleanAll_ and _cleanTest_), the image file is copied to the classpath as shown below. Is that the correct location for image test files?

```ShellSession
$ find . -name 16777216colors_diffpatt.png
./modules/javafx.graphics/src/test/resources/test/com/sun/glass/ui/monocle/16777216colors_diffpatt.png

$ gradle :graphics:test
    ...
> Task :graphics:compileShimsJava
    ...
> Task :graphics:compileTestJava
    ...
BUILD SUCCESSFUL in 2m 46s
138 actionable tasks: 138 executed

$ find . -name 16777216colors_diffpatt.png
./modules/javafx.graphics/build/resources/test/test/com/sun/glass/ui/monocle/16777216colors_diffpatt.png
./modules/javafx.graphics/src/test/resources/test/com/sun/glass/ui/monocle/16777216colors_diffpatt.png
```

jgneff:
> I don't have an e-paper device so I can't test it, but ...

I plan to address this problem in two ways. First, I'm creating videos of the new features using a test application that I will publish in GitHub. I also plan to request comments from developers who do have these devices and are familiar with the low-level programming interface, but who might be less familiar with Java. I wanted to get comments from the JavaFX experts first, though. Thanks again.

modules/javafx.graphics/src/main/native-glass/monocle/epd/mxcfb.h
```diff
@@ -0,0 +1,173 @@
+/*
+ * Copyright 2004-2011 Freescale Semiconductor, Inc. All Rights Reserved.
+ */
+
+/*
+ * The code contained herein is licensed under the GNU Lesser General
+ * Public License.  You may obtain a copy of the GNU Lesser General
+ * Public License Version 2.1 or later at the following locations:
+ *
```
kevinrushforth: This is third-party code, so we cannot accept this file.

modules/javafx.graphics/src/test/java/test/com/sun/glass/ui/monocle/FramebufferY8Test.java
kevinrushforth: This is similar to what we do for other tests.

modules/javafx.graphics/src/main/native-glass/monocle/epd/mxcfb.h
jgneff: Thank you, Kevin. I considered the license compatibility but forgot about the Oracle Contributor Agreement. If I understand the history correctly, David Hill ran into the same problem back in 2013. Look for `mxcfb.h` in the bug reports [JDK-8103599](https://bugs.openjdk.java.net/browse/JDK-8103599 "Add support for imx6") and [JDK-8103668](https://bugs.openjdk.java.net/browse/JDK-8103668 "Add support for imx6 - modify gradle files to fetch armv6-vfp-04.tgz"). There are two solutions to the problem still in the code today.

The first solution was to copy the header file manually to the directory for the ARM Linux system headers. The script that sets up the cross-compile toolchain still reminds you to do so:

```ShellSession
$ ../buildSrc/crosslibs/crosslibs-armv6hf.sh
...
WARNING: remember to hand patch in usr/include/linux/mxcfb.h for i.mx cursor support

Done.
```

The header file is never used, though, because a better solution hard-coded the field offsets in their C structures. For example, see the nested classes `MXCFBColorKey`, `MXCFBGblAlpha`, and `MXCFBPos` in the file [MX6Cursor.java](https://github.com/javafxports/openjdk-jfx/blob/develop/modules/javafx.graphics/src/main/java/com/sun/glass/ui/monocle/MX6Cursor.java). I should be able to use the same technique in my code.

There is another third-party file in my pull request. The file is [16777216colors diffpatt.png](https://commons.wikimedia.org/wiki/File:16777216colors_diffpatt.png), published under the Creative Commons "CC BY-SA 4.0" license and loaded by the unit test case [FramebufferY8Test.java](https://github.com/javafxports/openjdk-jfx/blob/b3963fddc47c75b9179604562bc4c97199bacd72/modules/javafx.graphics/src/test/java/test/com/sun/glass/ui/monocle/FramebufferY8Test.java). If this file cannot be included, I could generate the image within the test case itself. Generating the image will also prevent the problem that Johan encountered.

Unless I hear otherwise, I will test both changes over the next week:

* hard-code the field offsets and delete the header file, and
* generate the image in the test case and delete the image file.

kevinrushforth: This seems like a good solution.

modules/javafx.graphics/src/test/java/test/com/sun/glass/ui/monocle/FramebufferY8Test.java
jgneff: The image is removed, thereby removing the possibility of the `NullPointerException`. It turns out that all 24-bit RGB values can be generated by just one line of code!

```Java
IntStream.range(0, WIDTH * HEIGHT).forEachOrdered(pixels::put);
```

I made other changes to FramebufferY8Test unit tests, including some quick performance *sanity tests* of my replacement RGB565 pixel conversion.

modules/javafx.graphics/src/main/native-glass/monocle/epd/mxcfb.h
jgneff: This turned out to be a good change. Removing the header file allowed me to:

* remove 208 lines of native C source code plus the 120 lines of the C header file,
* move the initialization of the IOCTL request codes into the EPDSystem constructor and make their fields final,
* compute the size of the C structures at compile time instead of at runtime through a native call, and
* eliminate all native calls when getting and setting fields in the C structures.

The removal of the header file should also make it easier to accomodate changes to the structures in later versions of the EPD Controller API.

jgneff:
> JBS bug ID: https://bugs.openjdk.java.net/browse/JDK-8217605

Some kind of text-encoding tragedy occurred when my Request for Enhancement was added to the JDK Bug System. 😕 The problems are most visible on the [bugs.java.com](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8217605) page, which is covered in black diamond question marks (������), but the same errors exist on the direct [bugs.openjdk.java.net](https://bugs.openjdk.java.net/browse/JDK-8217605) page as well.

It doesn't seem right to leave the official report in such a state. Can you replace the description with the ASCII-encoded version attached below (`JDK-8217605.ASCII.txt`)? That file is identical to the existing text except for the following changes:

* Replaced mangled encodings for single quotes, double quotes, bullets, em dashes, and multiplication signs with their ASCII equivalents.
* Removed double quotes around all words italicized in the original GitHub issue.
* Fixed the trailing bracket on five links that were mistakenly surrounded by `<&gt;` instead of `<>`.

You should be able to do a direct copy and replace of the entire description. Thank you.

Attached file: [JDK-8217605.ASCII.txt](https://github.com/javafxports/openjdk-jfx/files/2972560/JDK-8217605.ASCII.txt)

johanvos:
This looks ok to me now.

modules/javafx.graphics/src/test/java/test/com/sun/glass/ui/monocle/EPDSettingsTest.java
```diff
@@ -0,0 +1,257 @@
...
+     */
+    @Test
+    public void testRotate() {
+        System.setProperty(ROTATE, "0");
+        settings = EPDSettingsShim.newInstance();
+        Assert.assertEquals(settings.rotate, 0);
+
+        System.setProperty(ROTATE, "1");
+        settings = EPDSettingsShim.newInstance();
+        Assert.assertEquals(settings.rotate, 1);
```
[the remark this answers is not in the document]
jgneff: Thanks. They should all be in their correct order now.

modules/javafx.graphics/src/test/java/test/com/sun/glass/ui/monocle/FramebufferY8Test.java
```diff
@@ -0,0 +1,372 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package test.com.sun.glass.ui.monocle;
+
+import com.sun.glass.ui.monocle.FramebufferY8Shim;
+import com.sun.glass.ui.monocle.FramebufferY8SuperShim;
+import java.awt.image.BufferedImage;
```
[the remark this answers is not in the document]
jgneff: The new tests are now under the system test directory with all the other Monocle tests.

modules/javafx.graphics/src/test/java/test/com/sun/glass/ui/monocle/FramebufferY8Test.java
```diff
@@ -0,0 +1,372 @@
...
+            printTime(source, "write", end - begin);
+        }
+    }
+
+    /**
+     * Copies the test image to an 8-bit buffer in Y8 grayscale format and saves
+     * the result as a PNG file when {@link #IMAGE_DEBUG} is {@code true};
+     * otherwise, does nothing.
+     */
+    @Test
```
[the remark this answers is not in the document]
jgneff: Thanks for the reminder about the `@Ignore` annotation. I dropped the boolean switch and added `@Ignore` on the two tests that save the pixel buffers as PNG image files. It's nice to have the tests available to verify the buffers visually, but they're not necessary for the automated tests.

jgneff:
Thank you for the comments, Kevin.

> Also, the FramebufferY8Test takes 35 seconds to run on my somewhat old Linux box, which is a bit high, and might be another reason to move them to the system tests;

The two calls to `Assert.assertArrayEquals` in `FramebufferY8Test` will get very, very much faster when we switch to JUnit version 4.12 from our current version 4.8.2. For details, see the thread with the subject, [*New unit tests 12 times slower under Gradle*](https://mail.openjdk.java.net/pipermail/openjfx-dev/2019-February/023113.html). On my somewhat new Linux box, JUnit 4.12 runs those particular tests 60 times faster, reducing their time to just 0.5 seconds instead of the current 30 seconds.

> My reading of the test is that this is expected. If so, can you add a print statement at the beginning of the tests that produce the log errors indicating that they are expected?

I print the notices to *standard output* just to keep them separate from the error log messages printed to *standard error.*

tests/system/src/test/java/test/com/sun/glass/ui/monocle/EPDSettingsTest.java
```diff
@@ -70,19 +72,20 @@ public void initialize() {
...
         settings = EPDSettingsShim.newInstance();
-        Assert.assertEquals(settings.bitsPerPixel, 16);
+        Assert.assertEquals(16, settings.bitsPerPixel);
 
         System.setProperty(BITS_PER_PIXEL, "32");
         settings = EPDSettingsShim.newInstance();
-        Assert.assertEquals(settings.bitsPerPixel, 32);
+        Assert.assertEquals(32, settings.bitsPerPixel);
 
+        System.out.println(String.format(VERIFY_ERROR, BITS_PER_PIXEL, 64));
```
kevinrushforth: Should print to `System.err`

tests/system/src/test/java/test/com/sun/glass/ui/monocle/EPDSettingsTest.java
```diff
@@ -159,27 +163,28 @@ public void testNoWait() {
...
         settings = EPDSettingsShim.newInstance();
-        Assert.assertEquals(settings.waveformMode, 3);
+        Assert.assertEquals(3, settings.waveformMode);
 
         System.setProperty(WAVEFORM_MODE, "4");
         settings = EPDSettingsShim.newInstance();
-        Assert.assertEquals(settings.waveformMode, 4);
+        Assert.assertEquals(4, settings.waveformMode);
 
+        System.out.println(String.format(VERIFY_ERROR, WAVEFORM_MODE, 5));
```
kevinrushforth: Should print to `System.err`

tests/system/src/test/java/test/com/sun/glass/ui/monocle/EPDSettingsTest.java
```diff
@@ -92,23 +95,24 @@ public void testBitsPerPixel() {
...
         settings = EPDSettingsShim.newInstance();
-        Assert.assertEquals(settings.rotate, 2);
+        Assert.assertEquals(2, settings.rotate);
 
         System.setProperty(ROTATE, "3");
         settings = EPDSettingsShim.newInstance();
-        Assert.assertEquals(settings.rotate, 3);
+        Assert.assertEquals(3, settings.rotate);
 
+        System.out.println(String.format(VERIFY_ERROR, ROTATE, 4));
```
kevinrushforth: Should print to `System.err`

jgneff:
> The messages about the log errors being expected need to go to `System.err` in order to be effective.

Thanks, Kevin. I went back and forth on that, but I was looking only through the Web interface. I'll change them so that the two new tests print the following:

```ShellSession
$ gradle --info -PFULL_TEST=true :systemTests:test --tests test.com.sun.glass.ui.monocle.*
...
test.com.sun.glass.ui.monocle.EPDSettingsTest > testBitsPerPixel STANDARD_ERROR
    Verify the error log message for monocle.epd.bitsPerPixel=64.
    Mar 22, 2019 8:17:24 AM com.sun.glass.ui.monocle.EPDSettings getInteger
    SEVERE: Value of monocle.epd.bitsPerPixel=64 not in [8, 16, 32]; using default (32)

test.com.sun.glass.ui.monocle.EPDSettingsTest > testRotate STANDARD_ERROR
    Verify the error log message for monocle.epd.rotate=4.
    Mar 22, 2019 8:17:24 AM com.sun.glass.ui.monocle.EPDSettings getInteger
    SEVERE: Value of monocle.epd.rotate=4 not in [0, 1, 2, 3]; using default (0)

test.com.sun.glass.ui.monocle.EPDSettingsTest > testWaveformMode STANDARD_ERROR
    Verify the error log message for monocle.epd.waveformMode=5.
    Mar 22, 2019 8:17:24 AM com.sun.glass.ui.monocle.EPDSettings getInteger
    SEVERE: Value of monocle.epd.waveformMode=5 not in [1, 2, 3, 4, 257]; using default (257)
...
test.com.sun.glass.ui.monocle.FramebufferY8Test > timeCopyTo16 STANDARD_OUT
    Converted 10 frames of 4,096 × 4,096 px to RGB565 in 594 ms (59 ms/frame): Framebuffer.copyToBuffer
    Converted 10 frames of 4,096 × 4,096 px to RGB565 in 471 ms (47 ms/frame): FramebufferY8.copyToBuffer

test.com.sun.glass.ui.monocle.FramebufferY8Test > timeWriteTo16 STANDARD_OUT
    Converted 10 frames of 4,096 × 4,096 px to RGB565 in 601 ms (60 ms/frame): Framebuffer.write
    Converted 10 frames of 4,096 × 4,096 px to RGB565 in 474 ms (47 ms/frame): FramebufferY8.write

test.com.sun.glass.ui.monocle.FramebufferY8Test > saveImage SKIPPED

test.com.sun.glass.ui.monocle.FramebufferY8Test > saveImageY8 SKIPPED
...
```

By the way, the sole purpose of `FramebufferY8Test` is to demonstrate (for a follow-up pull request) that the methods in `FramebufferY8` can replace the ones they override in `Framebuffer`. After doing that, we will be able to delete all of `FramebufferY8`, `FramebufferY8Test`, `FramebufferY8Shim`, and `FramebufferY8SuperShim`.

jgneff:
Thank you, Kevin. What's next? Should I request review comments now from outside the JavaFX community (specifically, from [these experts](https://www.mobileread.com/forums/forumdisplay.php?f=247 "Kobo Developer's Corner") on Kobo programming), or would you prefer that I gather those comments in a separate issue? In either case, should I merge the more recent changes in the *develop* branch into this *monocle-epd* pull request branch now?

If this pull request gets merged into OpenJFX for version 13, I plan to create two more pull requests before the July deadline, just to cover all of the Kobo e-reader devices now available:

* Add support for e-paper displays on the i.MX6SL processor
* Add support for e-paper displays on the i.MX6SLL processor

If all of them get into OpenJFX 13, they have a chance of getting into Ubuntu Base 20.04 LTS, too, making it trivial to install the JavaFX support in a *chroot* environment on e-paper devices.

kevinrushforth:
>  What's next?

I think this is ready to merge. The JBS issue is assigned to Johan, although either one of us can sponsor this change.

> Should I request review comments now from outside the JavaFX community (specifically, from these experts on Kobo programming), or would you prefer that I gather those comments in a separate issue?

Probably in a separate issue, unless you feel more feedback is needed prior to merging this initial change.

> In either case, should I merge the more recent changes in the develop branch into this monocle-epd pull request branch now?

In general it is a good idea to trigger a build if there have been any changes in develop that might impact this fix. Probably no need in this specific case, since I have no reason to believe that a recent change in the develop branch will cause problems with this fix.

jgneff:
> Probably in a separate issue, unless you feel more feedback is needed prior to merging this initial change.

No, I'm quite happy with the changes you requested. I probably should write more documentation on setting up the device before asking for outside reviews, anyway.

> Probably no need in this specific case, since I have no reason to believe that a recent change in the develop branch will cause problems with this fix.

I modified only two existing files, `buildSrc/armv6hf.gradle` for the build and `modules/javafx.graphics/src/test/addExports` for the tests, so this pull request should be very safe to merge.

**Wait!** Listing those two files made me realize that the change to `addExports` is no longer required now that I moved the new unit tests into the system tests. The corresponding file for the system tests, `tests/system/src/test/addExports`, already contains an entry for the required package:

```
--add-exports javafx.graphics/com.sun.glass.ui.monocle=ALL-UNNAMED
```

I'll remove my change to the unit test file right now, and **then** we'll be good to go (with only one existing file modified).

kevinrushforth:
> I'll remove my change to the unit test file right now, and then we'll be good to go (with only one existing file modified).

OK. That will also trigger a build, which is good.

kevinrushforth:
Build with reverted change in the graphics test addExports file passed. The change looks good.

Issue #366, opened 2019-01-31
Scrolling has become slow on Mac with OpenJFX 12


mipastgt:
When trying out the EA-release of OpenJFX 12 I observed that scrolling of controls with the mouse-wheel has become about 10 times slower than it was before. I first saw this in a TextArea but other scrollable controls are also affected. I observed this on a Mac with 
java.runtime.version: 11.0.2+9
javafx.version: 12-ea
This worked correctly until and including OpenJFX 11.0.2.

johanvos:
Is this with an external mouse?

mipastgt:
Yes, an USB mouse attached to a MacBook Pro.

johanvos:
If you talk about the factor 10, this might be related to https://github.com/javafxports/openjdk-jfx/pull/274

mipastgt:
Well, the figure 10 is just an (educated) guess because I knew about this change.

johanvos:
The change in http://hg.openjdk.java.net/openjfx/jfx-dev/rt/rev/6859462e6792 also includes a division by 10 in the value of the mouse delta's.
We should probably make a difference between real mousewheel scrolls and the long scrolling via the trackpad.

johanvos:
JBS link: https://bugs.openjdk.java.net/browse/JDK-8218424

mipastgt:
Hi Johan,
from my side the 12+12 release looks good. I tested it with the full application and scrolling seems to work as expected now.
Michael

johanvos:
Fixed in jfx-12, will be synced to master

PR #371, opened 2019-02-05
JDK-8218424 - Use precise scrolling only when supported (#366)

jperedadnr:
Use `hasPreciseScrollingDeltas` to find out if the scroll event comes from a trackpad (or some mice), that provides much more precise delta, or from a generic scroll wheel, that issues rather coarse scroll deltas.


johanvos:
closing PR as it is in hg

PR #373, opened 2019-02-08
JDK-8218611: [DRT] fast/xslt tests fails with Unsupported encoding windows-1251

ararunprasad:
Linked JBS: https://bugs.openjdk.java.net/browse/JDK-8218611


ararunprasad:
ICU is enabled for Windows and Linux build of libxml2, but the same is missing for macOS. It is a regression of PR #340 .

johanvos:
this looks like something we really need in 12.

ararunprasad:
> this looks like something we really need in 12.

@johanvos , Absolutely. :)

PR #378, opened 2019-02-14
JDK-8219008: Update OpenGL Headers to version 4.6

kevinrushforth:
This fix is for JBS bug [JDK-8219008](https://bugs.openjdk.java.net/browse/JDK-8219008), and updates the version of the Khronos OpenGL Header Files to 4.6.

We currently use the older glext files from GL 4.3; the latest is GL 4.6.

The following was done:

1. Replaced existing `glext.h`, `glxext.h`, and `wglext.h` files with the latest versions from Khronos.
2. Copied the latest `khrplatform.h` file which all of the above now depend on.
3. Added one needed typedef to our version of `glx.h` to account for the fact that newer versions of glxext.h expect it to already be defined by a `glx.h` file that claims GLX 1.4 support.

Notes to reviewers:

* No changes were needed to the 4 copied files; they are identical to the latest files from Khronos
* The diffs are very large because the files were significantly rearranged between version 4.3 and 4.6
* We don't actually use anything from OpenGL 4.x, so this change is intended to be behavior-neutral
* I have done a full build and test on Mac and Linux, and a sanity build on Windows with INCLUDE_ES2=true


kevinrushforth:
The newly-added opengl_fx.md file is almost identical to the equivalent opengl.md file for Java2D. The only differences are the version number and copyright years.

Issue #391, opened 2019-02-28
JDK-8220012: Accordion control holds reference to child pane after it is removed


fthevenet:
The Accordion control keeps a reference to its first TitledPane even after all panes have been removed, which prevents it to be collected, alongside everything attached to that pane.

The reason for that is the skin for the control explicitly  keeps a reference to it named firstTitledPane, which is kept in sync with the list of panes in the control through the use of a ListChangeListener, but it doesn't appear to ever sever the reference after the last TitlePane has been removed:

https://github.com/javafxports/openjdk-jfx/blob/3179dbd96653c22f329d75e7c7074612aef337b3/modules/javafx.controls/src/main/java/javafx/scene/control/skin/AccordionSkin.java#L101-L118

Unless I'm missing something, simply setting firstTitledPane to null if control.getPanes().isEmpty() is true should do the trick.

fthevenet:
I have filed a bug in the JBS for this issue: [JDK-8220012](https://bugs.openjdk.java.net/browse/JDK-8220012)

kevinrushforth:
Thank you for filing the JBS bug.

If you are interested in contributing a fix for this issue, please read [CONTRIBUTING.md](https://github.com/javafxports/openjdk-jfx/blob/develop/.github/CONTRIBUTING.md).

kevinrushforth:
Also, we will need a test case that reproduces the bug.

fthevenet:
I've submitted the following to the JBS bug:

> Steps to reproduce the issue:
> - Create a scene with an Accordion control
> - Start monitoring the heap usage.
> - Add a TitledPane to the Accordion, and to that pane any number of child controls susceptible to have noticeable impact on the heap usage (in the application where the issue was spotted, it consists of several XYCharts with several thousands of samples each, for instance)
> - Take a note of the increase in heap usage.
> - Remove the TitledPane from the Accordion
> - Force a GC
> - Notice that the old gen usage on the heap does not significantly decrease
> - Add a new TitledPane, this time with no children,  to the accordion
> - Force a GC
> - Notice that now the heap usage has decrease significantly, as the reference to the "big" titled pane was replace by one the the "small" one, and the former could therefore be collected.
> 
> Alternatively, using a Java profiler capable of tracing paths of strongly referenced objects to GC roots can also be used to illustrate the issue more accurately (I used Yourkit to identify the bug in the first place, for instance.)
>

kevinrushforth:
[...]

PR #395, opened 2019-03-02
[JDK-8220012] Sever reference to titled pane if accordion is empty.

fthevenet:
The Accordion control keeps a reference to its first TitledPane even after all panes have been removed, which prevents it to be collected, alongside everything attached to that pane.

The reason for that is the skin for the control explicitly keeps a reference to it named firstTitledPane, which is kept in sync with the list of panes in the control through the use of a ListChangeListener, but it doesn't sever the reference after the last TitlePane has been removed.

Ref #391

### DIFF
--- a/tests/system/src/test/java/test/javafx/scene/AccordionTitlePaneLeakTest.java
+++ b/tests/system/src/test/java/test/javafx/scene/AccordionTitlePaneLeakTest.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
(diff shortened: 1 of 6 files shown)


kevinrushforth:
Thank you for your proposed contribution. Before we can evaluate this, we will need a signed [OCA](https://www.oracle.com/technetwork/community/oca-486395.html) on file. If you have already sent one in, great. Please be aware that it usually takes 5-7 business days.

Also, please provide an automated unit test that fails without your fix and passes with your fix.

kevinrushforth:
@fthevenet I can confirm that your OCA has been recorded. We will now be able to review it. Thank you for your patience.

kevinrushforth:
Pointer to JBS issue: https://bugs.openjdk.java.net/browse/JDK-8220012

tests/system/src/test/java/test/javafx/scene/AccordionTitlePaneLeakTest.java
```diff
@@ -0,0 +1,109 @@
...
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+
+package test.javafx.scene;
```
kevinrushforth: Minor: I recommend creating a "control" subdirectory for this test. Also, can you remove the extra blank line before the package statement (one is sufficient).

fthevenet: Done in b327742

aghaisas:
The fix looks OK.

I ran the test without code change. The test failed as expected.
The test passed with the code change.

aghaisas:
Do we need changes made to files under .idea directory?

kevinrushforth:
> Do we need changes made to files under .idea directory?

Those changes need to be reverted.

kevinrushforth:
Other than the extra changes to the .idea files, this version looks good. Once those files are reverted (so that they don't appear in the aggregate patch), I'll approve this PR.

fthevenet:
I'm so used to gitignore the IDE files I didin't pay attention, sorry about that.

What would you say is best: to reset and force commit without the unwanted files, or to submit a new commit that revert the unwanted changes? I'd normally be inclined to do the former, but CONTRIBUTING.md states force commit on PR should be avoided.

kevinrushforth:
The latter, please. push a commit that reverts the changes to their previous state. It's much easier for me to review it that way. Thanks.

Issue #370, opened 2019-02-05
Accelerator + System MenuBar


abhinayagarwal:
### Background

If the accelerator set on a MenuItem, placed on a MenuBar, is similar to the key events pre-defined on a control, the accelerator shouldn't fire the action of the MenuItem. This can be observed in edit actions of a `TextInputControl`. When an edit action (like copy, paste etc.) are defined as actions on a MenuItem using KeyEvent(s), these are only fired when the focus is not on a `TextInputControl`.

### Bug

The above mentioned contract is invalid when `useSystemMenuBar` property of the `MenuBar` is set to `true` on a mac. In case of a system menu bar, actions like paste ( CMD+V ) triggers the event for both `TextInputControl` and `MenuItem`.

 The following example can be used to reproduce the issue on a mac:

```
import javafx.application.Application;
import javafx.scene.Scene;
import javafx.scene.control.Menu;
import javafx.scene.control.MenuBar;
import javafx.scene.control.MenuItem;
import javafx.scene.control.TextField;
import javafx.scene.input.KeyCode;
import javafx.scene.input.KeyCodeCombination;
import javafx.scene.input.KeyCombination;
import javafx.scene.layout.BorderPane;
import javafx.stage.Stage;

public class MenuBarAcceleratorBug extends Application {

    @Override
    public void start(Stage stage) {
        
        TextField textField = new TextField();
        
        MenuItem paste = new MenuItem("Paste");
        paste.setOnAction(e -> textField.paste());
        paste.setAccelerator(new KeyCodeCombination(KeyCode.V, KeyCombination.META_DOWN));
        
        Menu menu = new Menu("Edit");
        menu.getItems().add(paste);
        MenuBar menuBar = new MenuBar(menu);
        menuBar.setUseSystemMenuBar(true);
        
        BorderPane pane = new BorderPane(textField);
        pane.setTop(menuBar);
        
        stage.setScene(new Scene(pane, 300, 200));
        stage.show();
    }
}
```

TimurJD:
Here is the original issue found https://github.com/gluonhq/scenebuilder/issues/221

So, yes, the bug can be easily reproduced

johanvos:
The problem is in ` - (BOOL)performKeyEquivalent:(NSEvent *)theEvent` in `native-glass/mac/GlassView3D.m`
That method always return `NO`, allowing the regular key processing to do its work. In case the action is processed by the application, we should return `YES` here.

[...]

PR #397, opened 2019-03-04
8089986: Removed beeping when alt+key is pressed on Windows

DeanWookey:
On Windows, a beep sound always plays when using accelerators/mnemonics with the alt key, even if those accelerators exist. It's only supposed to beep when no such key combination is registered (alt+L vs alt+F for example)

I've found a solution which never beeps which I believe is preferable. https://github.com/DeanWookey/openjdk-jfx/commit/3f8520e22f5090a35097005254133f2eac8f97cc, based on this stack overflow post: https://stackoverflow.com/questions/3662192/disable-messagebeep-on-invalid-syskeypress

Javafx on windows doesn't support the windows system menu, as seen in WinApplication which doesn't override the _supportsSystemMenu() method and returns false. The f10 and alt+f10 events are also prevented from being processed by Windows. 

`                if (wParam == VK_MENU || (wParam == VK_F10 && !GetModifiers())) {
                    // Disable activation of the window's system menu
                    return 0;
                }
`

Hence we should be safe to respond to the WM_MENUCHAR message by closing the system menu which doesn't beep ever. 

We could try make it beep when alt+key is not consumed. Applications like Chrome don't do this, but I think it could be done.

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2019, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


kevinrushforth:
Pointer to JBS issue: https://bugs.openjdk.java.net/browse/JDK-8089986

kevinrushforth:
As mentioned on [this thread](https://mail.openjdk.java.net/pipermail/openjfx-dev/2019-March/023146.html) on the openjfx-dev mailing list, a fix for #370 might allow for a better fix (rather than the all or nothing approach we must choose between now).

kevinrushforth:
@pankaj-bansal will also review this.

kevinrushforth:
@DeanWookey did you have any more thoughts as to whether to wait for a fix for #370 or proceed with the fix you have proposed?

DeanWookey:
@kevinrushforth I would say proceed and that this could be enhanced when #370 gets resolved. I'm not 100% sure what the complete solution would look like. Last time I looked at it, it looked like there were 2 separate events, one which fired the event in javafx and this event which needs to be responded to appropriately. In the worst case we might have to store some state as to whether the event was dealt with in javafx so that we can respond to this event appropriately.

In my opinion, as it stands now, it's much better not to beep then beep every single time. I will look at it again when #370 gets resolved. If we'd rather wait and re-evaluate when it's done, that's fine too.

kevinrushforth:
@TurekBot the 8u code line in openjfx is no longer being maintained -- see [this message]([1] https://mail.openjdk.java.net/pipermail/openjfx-dev/2019-January/023039.html) -- so there is nowhere for you to backport it to.

Issue #4, opened 2018-02-28
Remove IDE specific files from the source code repository (JDK-8198795)


eugener:
Support for Gradle is pretty advanced in most modern IDEs, which means they can import the modules based just on Gradle project definition. So I propose to remove IDE specific files from the repo. This is the best practice anyway.

https://bugs.openjdk.java.net/browse/JDK-8198795

eugener:
This is relatively "low hanging fruit" so the process of reviewing and pushing fixes back to OpenJDK repo can be proved here.

Ciruman:
My college @streifi and I wanted to remove all IDE specific code and we found out that the IDEs(Netbeans, Eclipse and IntelliJ) couldn´t import the project from gradle. We had a look at the problem and we have found the problem in the gradle.build has a few errors(the dependencies are not correctly specified). We fixed it and now Netbeans and IntelliJ can import openjfx without a problem. Eclipse cannot import gradle/jigsaw projects yet.

We would like to push first this build fix and then remove the IDE files(and ignore them). 

Pull request: https://github.com/javafxports/openjdk-jfx/pull/393

DJViking:
> We fixed it and now Netbeans and IntelliJ can import openjfx without a problem. Eclipse cannot import gradle/jigsaw projects yet.

Eclipse can import Gradle projects just fine with Eclipse Gradle Buildship. I have no problem with using Eclipse with Gradle on Eclipse 2018-12, Gradle 5.x and Buildship 3.0.
Though I have an eclipse.gradle with additional eclipse configuration to get it working with jigsaw.

    eclipse {
        project {
            natures 'org.eclipse.buildship.core.gradleprojectnature'
        }

        classpath {
            file {
                whenMerged {
                    entries.findAll { isModule(it) }.each { //(1)
                        it.entryAttributes['module'] = 'true'
                    }

                    entries.findAll { isSource(it) && isTestScope(it) }.each {
                        it.entryAttributes['test'] = 'true'
                    }

                    entries.findAll { isLibrary(it) && isTestScope(it) }.each {
                        it.entryAttributes['test'] = 'true'
                    }
                }
            }

            downloadSources = true
            downloadJavadoc = true
        }
    }

    boolean isLibrary(entry) { return entry.properties.kind.equals('lib') }
    boolean isTestScope(entry) { return entry.entryAttributes.get('gradle_used_by_scope').equals('test'); }
[...]

PR #401, opened 2019-03-08
JDK-8220222: specify clearly gradle project dependencies

christianheilmann:
In order to import the openjdk-jfx project correctly in the different IDEs, the dependencies should be clearly defined. At the moment the project is working in IDEA and Netbeans correctly.

Currently, Netbeans has an issue open with gradle and jigsaw (https://issues.apache.org/jira/browse/NETBEANS-2004 and kelemen/netbeans-gradle-project#417).

So the only IDE working 100% correctly is IDEA for importing the openjdk-jfx project.

https://bugs.openjdk.java.net/browse/JDK-8220222


build.gradle
```diff
@@ -1847,12 +1847,24 @@ project(":base") {
 
     sourceSets {
         main
-        shims
-        test
+        shims{
```
nlisker: Missing space before `{`. Same for all subsequent places.

build.gradle
```diff
@@ -2396,13 +2431,14 @@ project(":swing") {
         //shims // no test shims needed
         test
     }
-
+
```
kevinrushforth: Please revert this unwanted white-space change.

build.gradle
```diff
@@ -2396,13 +2431,14 @@ project(":swing") {
...
     }
-
+            
     project.ext.moduleSourcePath = defaultModuleSourcePath
     project.ext.moduleSourcePathShim = defaultModuleSourcePathShim
 
     commonModuleSetup(project, [ 'base', 'graphics', 'swing' ])
 
     dependencies {
+        compile project(":media")
```
kevinrushforth: `javafx.swing` does not depend on `javafx.media`. I think this should instead be:

```
                compile project(":base")
                compile project(":graphics")
```

build.gradle
```diff
@@ -3119,6 +3179,9 @@ project(":web") {
     commonModuleSetup(project, [ 'base', 'graphics', 'controls', 'media', 'web' ])
 
     dependencies {
+        compile project(":base")
+        compile project(":graphics")
+        compile project(":media")
```
kevinrushforth: Should a dependency on `controls` be added here as well?

christianheilmann:
@kevinrushforth thanks for your review. You are right in all points and I integrated all your reviews and merged it with the develop branch of upstream.

This checkin enables full working on OpenJFX in IntelliJ IDEA Version 20.18.3 or higher.
Simply open the build.gradle in IntelliJ and you are ready to code.

The checked in IntelliJ project files should be removed, because they do not work and are misleading.
The remove of the IntelliJ project files should be done with https://github.com/javafxports/openjdk-jfx/issues/4

Issue #133, opened 2018-07-17
JDK-8208173: isComplexCharCode() returns false for U+11FF


nakajima-akira:
modules/javafx.graphics/src/main/java/com/sun/javafx/text/ScriptMapper.java

        else if (code < 0x11ff) { // U+1100 - U+11FF Old Hangul
            return true;
        }

---->  Fix

        else if (code <= 0x11ff) { // U+1100 - U+11FF Old Hangul
            return true;
        }

nakajima-akira:
Do I need to report this bug to [http://bugreport.java.com/](http://bugreport.java.com/)?
This bug code is that **_=_** is missing only.

kevinrushforth:
> Do I need to report this bug to http://bugreport.java.com/?

Yes.

nakajima-akira:
BUG-ID was published.
[https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8208173](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8208173)

kevinrushforth:
Do you plan to submit a PR for this bug? We are almost out of time for these sort of bugs for openjfx11.

nakajima-akira:
If no one does, I will do it.
openjfx11 is too late to patch from now.
My all patches (including already posted. PR 125,126,137) is for openjfx12.

PR #407, opened 2019-03-15
JDK-8208173: isComplexCharCode() returns false for U+11FF

nakajima-akira:
Issue is
https://github.com/javafxports/openjdk-jfx/issues/133

Simply equal is missing.

Old Hangul is defined U+1100 to U+11FF.
http://www.unicode.org/charts/PDF/U1100.pdf


kevinrushforth:
Pointer to JBS bug: https://bugs.openjdk.java.net/browse/JDK-8208173

PR #410, opened 2019-03-19
Update boot JDK to 12

kevinrushforth:
Note that this is in support of [JDK-8221097](https://bugs.openjdk.java.net/browse/JDK-8221097), but this PR touches exclusively the GitHub-only files for the CI build. As such, once it is ready, I will need to also commit and push it to the `master` branch manually.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,12 +11,12 @@ shallow_clone: true
 build_script:
   - ps: |
       $client = New-Object net.webclient
-      $client.DownloadFile('http://jdk.java.net/12/', 'C:\Users\appveyor\openjdk12.html')
-      $openJdk12 = cat C:\Users\appveyor\openjdk12.html | where { $_ -match "href.*https://download.java.net.*jdk12.*windows-x64.*zip\`"" } | %{ $_ -replace "^.*https:", "https:" } | %{ $_ -replace ".zip\`".*$", ".zip" }
-      echo "Download boot JDK from: $openJdk12"
-      $client.DownloadFile($openJdk12, 'C:\Users\appveyor\openjdk12.zip')
-      Expand-Archive -Path 'C:\Users\appveyor\openjdk12.zip' -DestinationPath 'C:\Users\appveyor\openjdk12'
-      Copy-Item -Path 'C:\Users\appveyor\openjdk12\*\' -Destination 'C:\jdk12' -Recurse -Force
+      $client.DownloadFile('http://jdk.java.net/12/', 'C:\Users\appveyor\openjdk.html')
+      $openJdk = cat C:\Users\appveyor\openjdk.html | where { $_ -match "href.*https://download.java.net.*jdk[1-9][0-9].*windows-x64.*zip\`"" } | %{ $_ -replace "^.*https:", "https:" } | %{ $_ -replace ".zip\`".*$", ".zip" }
+      echo "Download boot JDK from: $openJdk"
+      $client.DownloadFile($openJdk, 'C:\Users\appveyor\openjdk.zip')
+      Expand-Archive -Path 'C:\Users\appveyor\openjdk.zip' -DestinationPath 'C:\Users\appveyor\openjdk'
+      Copy-Item -Path 'C:\Users\appveyor\openjdk\*\' -Destination 'C:\jdk12' -Recurse -Force
       . .\tools\scripts\build.ps1 2>&1
 
 on_finish:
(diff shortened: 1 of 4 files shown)


kevinrushforth:
Notes to reviewers:

1. The Windows build script is a straightforward substitution of jdk11 --> jdk12

2. The Mac and Linux build script now uses the same logic as Windows, in that it downloads the html file from jdk.java.net/12 and parses it to get the location of the Mac or Linux JDK bundle, which it then downloads and unpacks (the "openjdk12" artifact isn't available directly).

kevinrushforth:
@brcolow you might want to review as well?

kevinrushforth:
I ran into a problem with our internal build of javadoc failing with JDK 12, so I am going to investigate further before proceeding on this.

brcolow:
Is the reason we are not using `jdk: openjdk12` because Travis CI doesn't support it yet or because we want to not be reliant upon it? Other than that question, LGTM.

kevinrushforth:
> Is the reason we are not using `jdk: openjdk12` because Travis CI doesn't support it yet

Yes. It failed when I used `jdk: openjdk12`.

kevinrushforth:
> I ran into a problem with our internal build of javadoc failing with JDK 12

This is due to a long-standing issue in our local build environment (we are still pointing to a JDK tree when building behind the firewall). I'll fix this locally and then can proceed with this.

mwkroening:
Travis is using [bach/install-jdk.sh](https://github.com/sormuras/bach/blob/master/install-jdk.sh) for downloading the JDK ([Source](https://sormuras.github.io/blog/2018-05-31-jdk-matrix.html)), which has been updated today for 12. I'm not sure when this will be in Travis, but using this script directly instead might be an alternative to unpacking it manually.

kevinrushforth:
I'll probably wait until next week then and see if it makes it into Travis by then. If so, then that seems an easier and better approach.

If not, then using the install-jdk.sh script directly might be another option (although one advantage of my proposed approach is that it will work with early access builds if we have a need to switch to one at some point).

kevinrushforth:
> Is it possible to name variables generically?

Good suggestion. I'll do that.

kevinrushforth:
It looks like openjdk12 is now available in Travis, so I reverted the changes to those files and instead use Travis to download openjdk12 (so a simple replacement).

I also implemented the suggestions that @ararunprasad made, which will minimize the diffs the next time we upgrade.

kevinrushforth:
> Much cleaner!

Yeah, if I had just waited a couple of days until now to start on this I wouldn't have had to do all that extra throw-away work. Oh well...

appveyor.yml
```diff
@@ -3,20 +3,20 @@ version: "{branch} {build}"
...
-      echo "Download boot JDK from: $openJdk11"
-      $client.DownloadFile($openJdk11, 'C:\Users\appveyor\openjdk11.zip')
-      Expand-Archive -Path 'C:\Users\appveyor\openjdk11.zip' -DestinationPath 'C:\Users\appveyor\openjdk11'
-      Copy-Item -Path 'C:\Users\appveyor\openjdk11\*\' -Destination 'C:\jdk11' -Recurse -Force
+      $client.DownloadFile('http://jdk.java.net/12/', 'C:\Users\appveyor\openjdk.html')
+      $openJdk = cat C:\Users\appveyor\openjdk.html | where { $_ -match "href.*https://download.java.net.*jdk[1-9][0-9].*windows-x64.*zip\`"" } | %{ $_ -replace "^.*https:", "https:" } | %{ $_ -replace ".zip\`".*$", ".zip" }
+      echo "Download boot JDK from: $openJdk"
+      $client.DownloadFile($openJdk, 'C:\Users\appveyor\openjdk.zip')
+      Expand-Archive -Path 'C:\Users\appveyor\openjdk.zip' -DestinationPath 'C:\Users\appveyor\openjdk'
+      Copy-Item -Path 'C:\Users\appveyor\openjdk\*\' -Destination 'C:\jdk12' -Recurse -Force
```
ararunprasad: nit: Is it possible to replace the string `C:\jdk12` with `$JAVA_HOME`.

kevinrushforth: It should be possible as long as the `environment` block is executed before the `build_script` block (it's easy enough to check). If so, I'll make that change and then merge it (else I'll merge it as-is).

kevinrushforth:
Using `$env:JAVA_HOME` works, so I'll included that change and merge this on Monday.

PR #414, opened 2019-03-21
JDK-8215686: FX build fails using gradle 5

kevinrushforth:
This is a preliminary review for [JDK-8215686](https://bugs.openjdk.java.net/browse/JDK-8215686) to allow the JavaFX to work with gradle 5.x, in support of upgrading to gradle 5.3 -- [JDK-8218172](https://bugs.openjdk.java.net/browse/JDK-8218172). It includes a fix for [JDK-8210509](https://bugs.openjdk.java.net/browse/JDK-8210509), since a gradle 5 build will fail without that fix.

I've run a build and test on Linux and compared the artifacts from the following builds:

gradle 4.8 without patch
gradle 4.8 with patch
gradle 5.3 with patch

All three sets of generated artifacts are identical.

The next steps are to test it on the other two platforms (Windows and Mac) and to verify the maven publishing step, which I didn't do. The (minor) changes to the maven publishing task were needed because gradle 5 no longer defers the configuration of publishing definition. I added some verbose print statements, which I will remove before the final review, to verify that the tasks are being configureed correctly.

@tiainen can you test this by publishing a snapshot build using `gradle publish`? Once you give the OK, I'll remove the verbose print statements and send out a formal review.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1563,9 +1563,11 @@ void addMavenPublication(Project project, List<String> projectDependencies) {
                 maven(MavenPublication) {
                     artifactId = "javafx-${project.name}"
 
-                    artifact project.tasks."moduleEmptyPublicationJar$t.capital"
-                    artifact project.tasks."modularPublicationJar$t.capital" {
-                        classifier "$t.name"
+                    afterEvaluate {
+                        artifact project.tasks."moduleEmptyPublicationJar$t.capital"
+                        artifact project.tasks."modularPublicationJar$t.capital" {
+                            classifier "$t.name"
+                        }
                     }
 
                     pom.withXml {
@@ -3868,7 +3870,8 @@ compileTargets { t ->
         enabled = COMPILE_SWT
         group = "Basic"
         description = "Creates the javafx-swt.jar for the $t.name target"
-        archiveName = "${project(":swt").buildDir}/libs/javafx-swt.jar";
+        destinationDir = file("${project(":swt").buildDir}/libs")
+        archiveName = "javafx-swt.jar"
         includeEmptyDirs = false
         from("${project(":swt").buildDir}/classes/java/main");
         include("**/javafx/embed/swt/**")
@@ -3888,7 +3891,7 @@ compileTargets { t ->
         dependsOn(javafxSwtTask)
 
         doLast() {
-            ant.jar (update: true, index: true, destfile: javafxSwtTask.archiveName)
+            ant.jar (update: true, index: true, destfile: "${javafxSwtTask.destinationDir}/${javafxSwtTask.archiveName}")
         }
     }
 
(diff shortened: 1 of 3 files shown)


kevinrushforth:
The following three problems were fixed:

1. Gradle 5.x includes Groovy 2.5.x, which is more strict that earlier versions of Groovy. It doesn't like one of the final declarations in NativeCompileTask.groovy. See the discussion in PR #313. The fix is to remove the `final` keyword.

2. The javafxSwtJar task uses `archiveName` to point to a fully qualified path. Instead that property should contain only the base filename and the `destinationDir` property should be set to the path of the directory.

3. The rules for the Maven publishing block changed between gradle 4.x and gradle 5.x. This is described [here](https://docs.gradle.org/4.8/userguide/publishing_maven.html#publishing_maven:deferred_configuration) as well as in the [Updating from gradle 4 to gradle 5](https://docs.gradle.org/current/userguide/upgrading_version_4.html) doc, and was originally filed as [JDK-8210509](https://bugs.openjdk.java.net/browse/JDK-8210509). The fix is to add `enableFeaturePreview('STABLE_PUBLISHING')` to `settings.gradle` (so that 4.8 builds use the new behavior) and surround the commands that set the artifacts in the publishing task in an `afterEvaluate` block.

tiainen:
OK, I'll do a build on all platforms and verify if the produced artifacts are identical in all cases.

tiainen:
I've published the artifacts to the following staging repository: https://oss.sonatype.org/content/repositories/orgopenjfx-1041. The artifacts have version `13-ea+302`. I will drop it once it is verified that everything looks in order.

Additionally, for convenience, I've also provided downloads for the jmods and sdks for each platform:
https://download2.gluonhq.com/openjfx/13/openjfx-13-ea+302_linux-x64_bin-jmods.zip
https://download2.gluonhq.com/openjfx/13/openjfx-13-ea+302_linux-x64_bin-sdk.zip
http://download2.gluonhq.com/openjfx/13/openjfx-13-ea+302_osx-x64_bin-jmods.zip
http://download2.gluonhq.com/openjfx/13/openjfx-13-ea+302_osx-x64_bin-sdk.zip
http://download2.gluonhq.com/openjfx/13/openjfx-13-ea+302_windows-x64_bin-jmods.zip
http://download2.gluonhq.com/openjfx/13/openjfx-13-ea+302_windows-x64_bin-sdk.zip

kevinrushforth:
> I've published the artifacts to the following staging repository

Thanks for testing. I did a quick sanity check on Linux and the artifacts look fine. The fact that you were able to publish to maven checked the only thing I couldn't test ahead of time, so that's good.

kevinrushforth:
The Mac build timed out while installing the build tools...it looks like a problem with the build environment.

kevinrushforth:
This is now ready for a formal review. I also send an [RFR](https://mail.openjdk.java.net/pipermail/openjfx-dev/2019-March/023155.html) to the mailing list.

PR #421, opened 2019-03-27
JDK-8218172: Upgrade to gradle 5.3

kevinrushforth:
https://bugs.openjdk.java.net/browse/JDK-8218172

This updates the gradle wrapper script to gradle 5.3. Production builds should start using 5.3 after this is merged.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1235,7 +1235,7 @@ if (gradle.gradleVersion != jfxGradleVersion) {
 
     logger.warn("*****************************************************************");
     logger.warn("Unsupported gradle version $gradle.gradleVersion in use.");
-    logger.warn("Only version 5.3 is supported. Use this version at your own risk");
+    logger.warn("Only version $jfxGradleVersion is supported. Use this version at your own risk");
     if ( err != "") logger.warn(err);
     logger.warn("*****************************************************************");
 }
--- a/build.gradle
+++ b/build.gradle
@@ -1216,14 +1216,17 @@ if (Integer.parseInt(NUM_COMPILE_THREADS.toString()) < 1) {
     NUM_COMPILE_THREADS = 1
 }
 
-// Check for Gradle 5.3, error if < 4.8.
-if (gradle.gradleVersion != "5.3") {
+// Check gradle version
+if (gradle.gradleVersion != jfxGradleVersion) {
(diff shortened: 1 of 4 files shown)


kevinrushforth:
Per the evaluation of PR #414 I have run tests on all platforms and verified that gradle 5.3 produces the same artifacts as gradle 4.8.

kevinrushforth:
Ok, I’ll do that.

tiainen:
Just for clarity, the minimum required gradle version remains at 4.8?

kevinrushforth:
This will not change the minimum, which remains at 4.8.

I also noticed that I forgot to update the gradle.md legal file, so I'll do that in the next rev

nlisker:
I updated the wiki page: https://wiki.openjdk.java.net/display/OpenJFX/Building+OpenJFX#BuildingOpenJFX-Gradle

mwkroening:
[v5.3.1](https://github.com/gradle/gradle/releases/tag/v5.3.1) has just been released.

"We recommend that you use Gradle 5.3.1 over 5.3."

aghaisas:
I did a build and sample unit test run on Mac.

kevinrushforth:
> v5.3.1 has just been released.

Thanks for letting me know. I see that it's new enough to not yet be on the main gradle download page.

Given that we have completed our testing and are not hitting the bugs fixed in 5.3.1, I think it makes sense to go ahead with the 5.3 upgrade and file a new issue to consider switching to 5.3.1. We will need to redo all our testing (in particular, the Maven publishing testing) before upgrading to 5.3.1 anyway, as well as getting approval (which is trivial for a minor bug fix release, but still needs to be done).

@tiainen @erwin1 - what is your preference?

kevinrushforth:
> I see that it's new enough to not yet be on the main gradle download page.

The 5.3.1 release is there now (it wasn't earlier when I checked).

kevinrushforth:
I'll merge this later today absent any objections.

nlisker:
I tested 5.3.1 on Win10 and the build was successful.

kevinrushforth:
> I tested 5.3.1 on Win10 and the build was successful.

Thanks for testing. We'll go with 5.3 at this time (to avoid a full retest of the production build, including Maven publishing), but could consider upgrading to 5.3.1 at a later time if there is a need.

PR #427, opened 2019-04-01
JDK-8214716: Record the versions of tools used to build FX in the repo

ararunprasad:


tools.info
```diff
@@ -0,0 +1,8 @@
+GCC_VERSION="gcc7.3.0-OEL6.4+1.1"
+MSVC_VERSION="VS2017-15.5.5+1.1"
+XCODE_VERSION="Xcode9.4-MacOSX10.13+1.0"
+CMAKE_VERSION="3.13.3"
+NINJA_VERSION="1.8.2"
+GRADLE_VERSION="4.8"
```
johanvos: Gradle is now at 5.3

tools.info
```diff
@@ -0,0 +1,8 @@
+GCC_VERSION="gcc7.3.0-OEL6.4+1.1"
+MSVC_VERSION="VS2017-15.5.5+1.1"
```
johanvos: Do you want to prefix platform-specific tools (e.g. WIN_... ) versus generic tools (e.g. gradle) ?

tools.info
ararunprasad: Thanks @johanvos . I will fix it.

tools.info
ararunprasad: I thought it would be redundant. If you think it is good to have then I don't have any objection.

kevinrushforth: I don't have a strong opinion one way or the other.

ararunprasad:
@kevinrushforth Moved the build tools information to build.properties. Please take a look.

Thanks.

build.properties
```diff
@@ -72,6 +72,7 @@ javadoc.header=JavaFX&nbsp;13
 #
 ##############################################################################
 
+# JDK
 jfx.build.jdk.version.min=11
```
ararunprasad: Should I change this to 12 & build number to 33?

kevinrushforth: No. The minimum is still JDK 11. Instead, you can reintroduce the `jfx.build.jdk.version` and `jfx.build.jdk.buildnum` variables to indicate the version that we use in production.

build.properties
```diff
@@ -83,3 +84,13 @@ jfx.build.jdk.buildnum.min=28
 # that is supported. It must be <= jfx.gradle.version.
 jfx.gradle.version=5.3
 jfx.gradle.version.min=4.8
+
+# Toolchains
+jfx.build.linux.gcc.version.min=gcc7.3.0-OEL6.4+1.1
+jfx.build.windows.msvc.version.min=VS2017-15.5.5+1.1
+jfx.build.macosx.xcode.version.min=Xcode9.4-MacOSX10.13+1.0
```
kevinrushforth: I think for the above three, you can just drop the `.min` (although it would be helpful to note what the minimum is in a comment, we probably don't need a variable for the min unless we are going to check and fail the build if it is below the minimum).

build.properties
```diff
@@ -83,3 +84,13 @@ jfx.build.jdk.buildnum.min=28
...
+
+# Build tools
+jfx.build.cmake.version.min=3.13.3
+jfx.build.ninja.version.min=1.8.2
+jfx.build.ant.version.min=1.10.5
```
kevinrushforth: Similar to my earlier comment, you can probably just drop the `.min` from the variable name. If you do add variables for the minimum, they might be lower than the ones we use in production (e.g., the minimum `ant` version is `1.8.2`).

kevinrushforth:
Pointer to JBS bug: https://bugs.openjdk.java.net/browse/JDK-8214716

build.properties
ararunprasad: Thanks Kevin. I will drop the min suffix.

build.properties
ararunprasad: Thanks Kevin. I will drop the min suffix.

tools.info
johanvos: I agree it is redundant, but it helps developers who want to get a list of requirements. However, this might be better on the wiki then. Hence, I'm ok skipping the prefix.

johanvos: That is, I'm ok with the current proposal. (which has prefixes for the toolchain)

build.properties
```diff
@@ -72,6 +72,7 @@ javadoc.header=JavaFX&nbsp;13
 #
 ##############################################################################
 
+# JDK
```
kevinrushforth: Can you also add the following?

```
jfx.build.jdk.version=12
jfx.build.jdk.buildnum=33
```

PR #430, opened 2019-04-04
JDK-8221993: Create test to validate package declarations in source files

kevinrushforth:
https://bugs.openjdk.java.net/browse/JDK-8221993

This adds a task for each project, including systemTests, to test the `java` files in each sourceSet in the project. It is added as a dependency of each project's `compileTestJava` task, so that it will be run as part of `gradle test`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1654,12 +1654,7 @@ void validatePackages(File sourceRoot, boolean hasModuleName) {
         }
     }
     if (err) {
-        // FIXME: uncomment the call to "fail" and remove the print statement
-        //fail("Package validation failed")
-        println("Package validation failed")
-        println "**************************"
-        println "* FAIL THE BUILD!!!!!!!! *"
-        println "**************************"
+        fail("Package validation failed")
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -1616,6 +1616,99 @@ void addMavenPublication(Project project, List<String> projectDependencies) {
 }
 
 /**
+ * Verifies that all of the *.java files in a source tree have the
+ * correct package name. If not, fail the build.
+ *
+ * @param sourceRoot the root of the source tree to check
+ * @param hasModuleName true if the sources under sourceRoot are being
+ * compiled as modules, with the immediate child directories being the
+ * root(s) of the named modules being compiled. In this case the package
+ * root is one directory below the sourceRoot.
+ */
+void validatePackages(File sourceRoot, boolean hasModuleName) {
+    if (!sourceRoot.isDirectory()) {
+        return;
+    }
+    def err = false;
+    String sourceString = sourceRoot.toString()
+    def startPos = sourceString.length() + 1
+    logger.info("validating packages for ${sourceRoot}")
+    def inputFiles = fileTree(dir: sourceRoot, include: "**/*.java")
+    inputFiles.each { file ->
+        def packageRoot = file.toString().substring(startPos)
+        if (hasModuleName) {
+            packageRoot = packageRoot.substring(packageRoot.indexOf("/") + 1)
+        }
+        def endPos = packageRoot.lastIndexOf("/")
+        def pkgEx = endPos > -1 ?  packageRoot.substring(0, endPos).replace("/", ".") : ""
+        def pkg = ""
+        file.eachLine { line ->
+            def words = line.trim().split("[ ;]+")
+            if (words.length > 1 && words[0] == "package") {
+                pkg = words[1]
+            }
+        }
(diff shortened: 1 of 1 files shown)


kevinrushforth:
This is ready for a preliminary review. Once the fix for [https://bugs.openjdk.java.net/browse/JDK-8221941](https://bugs.openjdk.java.net/browse/JDK-8221941) is integrated, I will merge it into my branch and change the warning to an error, and send it out for formal review.

kevinrushforth:
The output of the CI jobs looks good to me now.

kevinrushforth:
NB: I expect the build to fail as a result of the previous push. Once it does, I will merge in the changes from develop, and send a formal RFR.

kevinrushforth:
Actually, no, this won't fail since it is building the result of merging my changes into the current `develop` branch of the `javafxports/openjdk-jfx` repo.

The Appveyor build on Windows fails because of a package mismatch in the antlr-generated sources in `modules/javafx.graphics/build/gensrc/antlr`. It looks like a real problem, so needs to be investigated.

kevinrushforth:
The Windows failure does, in fact, point to a real problem. For some reason, when the JSL grammar parsing files are generated on Windows they are written directly to the base directory in `modules/javafx.graphics/build/gensrc/antlr` rather than `modules/javafx.graphics/build/gensrc/antlr/com/sun/scenario/effect/compiler` as is expected, and as is done correctly on Linux and Mac.  This predates the antlr 4.7 upgrade. I will investigate further, but might need to exclude `sourceSets.jslc` in the javafx.graphics project from the package validation (if so, I'll file a build bug to fix the build issue on Windows). It looks to been this way for years, and hasn't caused any problems, so there seems to be no urgency to fix it.

kevinrushforth:
This is now ready for review. I filed [JDK-8222036](https://bugs.openjdk.java.net/browse/JDK-8222036) for the antlr bug. As a workaround for that bug, I have excluded sourceSets.jslc from package validation.

build.gradle
```diff
@@ -1615,6 +1615,94 @@ void addMavenPublication(Project project, List<String> projectDependencies) {
...
+        def packageRoot = file.toString().replace("\\", "/").substring(startPos)
+        if (hasModuleName) {
+            packageRoot = packageRoot.substring(packageRoot.indexOf("/") + 1)
+        }
+        def endPos = packageRoot.lastIndexOf("/")
+        def pkgEx = endPos > -1 ?  packageRoot.substring(0, endPos).replace("/", ".") : ""
+        def pkg = ""
+        file.eachLine { line ->
+            def words = line.trim().split("[ ;]+")
+            if (words.length > 1 && words[0] == "package") {
```
ararunprasad: I see that it is failing for snippets like below. :) It will be a false positives.
/*
package com.sun.javafx.webkit1.prism;
*/

build.gradle
```diff
@@ -1615,6 +1615,94 @@ void addMavenPublication(Project project, List<String> projectDependencies) {
...
+    def inputFiles = fileTree(dir: sourceRoot, include: "**/*.java")
+    inputFiles.each { file ->
+        def packageRoot = file.toString().replace("\\", "/").substring(startPos)
+        if (hasModuleName) {
+            packageRoot = packageRoot.substring(packageRoot.indexOf("/") + 1)
+        }
+        def endPos = packageRoot.lastIndexOf("/")
+        def pkgEx = endPos > -1 ?  packageRoot.substring(0, endPos).replace("/", ".") : ""
+        def pkg = ""
+        file.eachLine { line ->
```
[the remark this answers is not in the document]
kevinrushforth: Good idea. The package statement is always the first non-comment statement, so this seems like a good improvement.

build.gradle
kevinrushforth: It will also miss the following valid (but unlikely) declaration:

```
/* hi */ package javafx.scene;
```

Which will also cause a false positive. Probably not worth worrying about...

brcolow:
Very interesting adding a test inside of `build.gradle`. Would it be more reasonable to add this behavior to jlint? I am sure you have a reason for doing it the way you did, though.

kevinrushforth:
> Would it be more reasonable to add this behavior to jlint? I am sure you have a reason for doing it the way you did, though.

Anything else I could think of would be more work, and would still require changes to 'build.gradle` to both figure out what source-roots to check and to pass the source-root to the test program or tool. As for `jlint`, that is a third-party tool, and getting approval for such a tool would be unlikely (and a fair bit of work), especially given how simple it is to implement the check ourselves.

Issue #408, opened 2019-03-15
NPE in javafx.graphics/javafx.stage.Window$TKBoundsConfigurator.apply (JDK-8221987)


christianheilmann:
In "javafx.graphics/javafx.stage.Window$TKBoundsConfigurator.apply(...)" is missing the check that the peer is Null for the call "peer.setBounds(...);"
I run in situations where the peer is Null.

```
Exception in thread "JavaFX Application Thread" java.lang.NullPointerException
	at javafx.graphics/javafx.stage.Window$TKBoundsConfigurator.apply(Window.java:1550)
	at javafx.graphics/javafx.stage.Window$TKBoundsConfigurator.pulse(Window.java:1559)
	at javafx.graphics/com.sun.javafx.tk.Toolkit.lambda$runPulse$2(Toolkit.java:412)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at javafx.graphics/com.sun.javafx.tk.Toolkit.runPulse(Toolkit.java:411)
	at javafx.graphics/com.sun.javafx.tk.Toolkit.firePulse(Toolkit.java:435)
	at javafx.graphics/com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:562)
	at javafx.graphics/com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:542)
	at javafx.graphics/com.sun.javafx.tk.quantum.QuantumToolkit.pulseFromQueue(QuantumToolkit.java:535)
	at javafx.graphics/com.sun.javafx.tk.quantum.QuantumToolkit.lambda$runToolkit$11(QuantumToolkit.java:341)
	at javafx.graphics/com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
```

christianheilmann:
Bug Report is submitted to Oracle Bug Database under internal review ID : 9059794. I will include the bug id when they posted the bug.

kevinrushforth:
https://bugs.openjdk.java.net/browse/JDK-8221987

PR #431, opened 2019-04-05
JDK-8221987: javafx.graphics/javafx.stage.Window$TKBoundsConfigurator.apply (#408)          

christianheilmann:
STEPS TO FOLLOW TO REPRODUCE THE PROBLEM :
Use org.controlsfx.samples.HelloNotifications from https://github.com/controlsfx/controlsfx/blob/master/controlsfx-samples/src/main/java/org/controlsfx/samples/HelloNotifications.java and press very frequent on "Top-right notification" then the NPE will occur. It is independent of the os. We tested it on windows and mac.

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/stage/Window.java
+++ b/modules/javafx.graphics/src/main/java/javafx/stage/Window.java
@@ -1531,6 +1531,10 @@ public class Window implements EventTarget {
 
         public void apply() {
             if (dirty) {
+                if (peer == null) {
+                    reset();
+                    return;
+                }
                 // Snapshot values and then reset() before we call down
                 // as we may end up with recursive calls back up with
                 // new values that must be recorded as dirty.
@@ -1547,12 +1551,10 @@ public class Window implements EventTarget {
                 float newRX = (float) renderScaleX;
                 float newRY = (float) renderScaleY;
                 reset();
-                if (peer != null) {
-                    peer.setBounds(newX, newY, xSet, ySet,
-                            newWW, newWH, newCW, newCH,
-                            newXG, newYG,
-                            newRX, newRY);
-                }
+                peer.setBounds(newX, newY, xSet, ySet,
+                        newWW, newWH, newCW, newCH,
+                        newXG, newYG,
+                        newRX, newRY);
             }
         }
 
--- a/modules/javafx.graphics/src/main/java/javafx/stage/Window.java
+++ b/modules/javafx.graphics/src/main/java/javafx/stage/Window.java
@@ -1547,10 +1547,12 @@ public class Window implements EventTarget {
                 float newRX = (float) renderScaleX;
                 float newRY = (float) renderScaleY;
                 reset();
-                peer.setBounds(newX, newY, xSet, ySet,
-                                    newWW, newWH, newCW, newCH,
-                                    newXG, newYG,
-                                    newRX, newRY);
+                if (peer != null) {
+                    peer.setBounds(newX, newY, xSet, ySet,
+                            newWW, newWH, newCW, newCH,
+                            newXG, newYG,
+                            newRX, newRY);
+                }
             }
         }
 


christianheilmann:
https://github.com/javafxports/openjdk-jfx/issues/408

kevinrushforth:
This looks like a safe and obvious fix to me. I have tagged @arapte to do the formal review.

kevinrushforth:
https://bugs.openjdk.java.net/browse/JDK-8221987

modules/javafx.graphics/src/main/java/javafx/stage/Window.java
```diff
@@ -1547,10 +1547,12 @@ public void apply() {
...
-                peer.setBounds(newX, newY, xSet, ySet,
-                                    newWW, newWH, newCW, newCH,
-                                    newXG, newYG,
-                                    newRX, newRY);
+                if (peer != null) {
+                    peer.setBounds(newX, newY, xSet, ySet,
+                            newWW, newWH, newCW, newCH,
+                            newXG, newYG,
+                            newRX, newRY);
+                }
```
arapte: The set of local variables will not be needed when peer is null.
Please consider change as below,

public void apply() {
            if (dirty) {
                if (peer != null) {
                    // Snapshot values and then reset() before we call down
                    // as we may end up with recursive calls back up with
                    // new values that must be recorded as dirty.
                    boolean xSet = !Double.isNaN(x);
                    float newX = xSet ? (float) x : 0f;
                    boolean ySet = !Double.isNaN(y);
                    float newY = ySet ? (float) y : 0f;
                    float newWW = (float) windowWidth;
                    float newWH = (float) windowHeight;
                    float newCW = (float) clientWidth;
                    float newCH = (float) clientHeight;
                    float newXG = xGravity;
                    float newYG = yGravity;
                    float newRX = (float) renderScaleX;
                    float newRY = (float) renderScaleY;
                    reset();
                    peer.setBounds(newX, newY, xSet, ySet,
                                        newWW, newWH, newCW, newCH,
                                        newXG, newYG,
                                        newRX, newRY);
                } else {
                    reset();
                }
            }
        }

arapte:
I could observe the issue with, ControlsFX 8.40.14 & ControlsFX 9.0.0, downloaded from here https://github.com/controlsfx/controlsfx
But there is no controlsFX build for JDK 9 onwards. However the issue looks easily reproducible.

modules/javafx.graphics/src/main/java/javafx/stage/Window.java
kevinrushforth: A second alternative would be to return early from the method if peer is null, leaving the rest of the method unchanged.

```
    public void apply() {
        if (dirty) {
            if (peer == null) {
                reset();
                return;
            }
            ...
        }
```

I don't have a strong opinion. Any of the three choices (the original fix, Ambarish's proposed change, or my variation on that) are fine with me.

Ciruman:
> I could observe the issue with, ControlsFX 8.40.14 & ControlsFX 9.0.0, downloaded from here https://github.com/controlsfx/controlsfx
> But there is no controlsFX build for JDK 9 onwards. However the issue looks easily reproducible.

@arapte There are RC. You can find them here: 
https://search.maven.org/artifact/org.controlsfx/controlsfx/11.0.0-RC2/jar

modules/javafx.graphics/src/main/java/javafx/stage/Window.java
Ciruman: > A second alternative would be to return early from the method if peer is null, leaving the rest of the method unchanged.
> 
> ```
>     public void apply() {
>         if (dirty) {
>             if (peer == null) {
>                 reset();
>                 return;
>             }
>             ...
>         }
> ```
> I don't have a strong opinion. Any of the three choices (the original fix, Ambarish's proposed change, or my variation on that) are fine with me.

I find the proposal from @kevinrushforth clearer.

christianheilmann:
@kevinrushforth thanks for your code review. I integrated your suggestion.

PR #434, opened 2019-04-06
JDK-8222036: antlr generated files written to wrong location on Windows

kevinrushforth:
See [JDK-8222036](https://bugs.openjdk.java.net/browse/JDK-8222036).

This fix accounts for a limitation of antlr on the Windows platform in that it won't write its output files to the proper location underneath the output directory tree unless the file name passed into antlr has Windows-style backslashes.

The fix is to ensure that we use backslashes as the file separator when running antlr.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2456,14 +2456,7 @@ project(":graphics") {
 
     addMavenPublication(project, [ 'base' ])
 
-    /*
-     * The following is a workaroud for JDK-8222036 (antlr generated
-     * files written to wrong location on Windows). This workaround should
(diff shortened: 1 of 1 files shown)


brcolow:
Glad it's not a regression from the upgrade to antlr 4. Good work discovering this.

kevinrushforth:
This is now ready for review. I merged in the fix for JDK-8221993 ( PR #430 ), and removed the workaround added for this bug (JDK-8220036) which was done as part of that fix.

I'll send the RFR shortly.

brcolow:
LGTM.

PR #442, opened 2019-04-13
JDK-8221702: Use HTTPS to download all build dependencies

kevinrushforth:
This fixes [JDK-8221702](https://bugs.openjdk.java.net/browse/JDK-8221702).

We should always use HTTPS to download artifacts used during a build.


johanvos:
+1

PR #445, opened 2019-04-15
JDK-8220272: Window order is not correct when Modality.WINDOW_MODAL

tsayao:
Explained here:
https://mail.openjdk.java.net/pipermail/openjfx-dev/2019-April/023220.html

Fixes:
https://bugs.openjdk.java.net/browse/JDK-8220272

Signed OCA sent.

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
-  </component>
-</project>
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="hg4idea" />
+  </component>
+</project>
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="hg4idea" />
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
   </component>
 </project>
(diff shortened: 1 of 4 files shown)


kevinrushforth:
We will review this once your OCA has been processed (it usually takes a week, but sometimes longer).

kevinrushforth:
OCA has been recorded.

kevinrushforth:

kevinrushforth:
@arapte - since you were able to reproduce the problem, can you test the fix on your system?

tsayao:
Sorry, did a mess with the commits. Let me fix it.

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kevinrushforth: There are still diffs in this file relative to the base, such that it shows up in your aggregate patch. Rather than manually trying to fix it you need to revert it to the state it was in the develop branch (assuming that is what your patch is based off of). You can copy it manually or more easily doing this, while on your branch:

```
git checkout develop -- .idea/vcs.xml
```

And then commit and push it. It should then no longer show up in the aggregate patch.

tsayao: Thanks, I did a big mess (sorry). Seems to be ok now. Sorry for the long commit list.

kevinrushforth:
The patch looks correct now (same as initial, but with the reversion of the `.idea` file)

arapte:
> 
> 
> @arapte - since you were able to reproduce the problem, can you test the fix on your system?

Hi Kevin, I confirmed that the issue does not occur with this fix applied.

kevinrushforth:
@arapte Thanks. I don't know why I missed seeing it before, but I can now also reproduce the failure and see that it works correctly with this patch applies.

tests/system/src/test/java/test/robot/javafx/stage/CheckWindowOrderTest.java
```diff
@@ -0,0 +1,117 @@
...
+        TestStage openLastWindow(Window owner) {
+            TestStage stage = new TestStage(owner, "Last Window");
+            stage.initModality(Modality.WINDOW_MODAL);
+
+            return stage;
+        }
+    }
+
+    public static void waitForLatch(CountDownLatch latch, int seconds, String msg) throws Exception {
+        org.junit.Assert.assertTrue("Timeout: " + msg, latch.await(seconds, TimeUnit.SECONDS));
```
kevinrushforth: Minor nit: you can just use `Assert` rather than the fully-qualified class, since it is imported.

tsayao: done!

PR #446, opened 2019-04-16
JDK-8212060: Fix Window show and then move glitch.

tsayao:
This fixes the show and move glitch.

Explanation here:
https://mail.openjdk.java.net/pipermail/openjfx-dev/2019-April/023234.html

OCA pending.

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
@@ -1343,7 +1343,12 @@ void WindowContextTop::exit_fullscreen() {
 }
 
 void WindowContextTop::request_focus() {
-    gtk_window_present(GTK_WINDOW(gtk_widget));
+    //JDK-8212060: Window show and then move glitch.
+    //The WindowContextBase::set_visible will take care of showing the window.
+    //The below code will only handle later request_focus.
+    if (is_visible()) {
+        gtk_window_present(GTK_WINDOW(gtk_widget));
+    }
 }
 
 void WindowContextTop::set_focusable(bool focusable) {


kevinrushforth:
Thanks. As with #445 we will review this once your OCA has been processed.

kevinrushforth:
OCA has been recorded

kevinrushforth:

kevinrushforth:
JBS bug: https://bugs.openjdk.java.net/browse/JDK-8212060

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
```diff
@@ -1343,7 +1343,12 @@ void WindowContextTop::exit_fullscreen() {
 }
 
 void WindowContextTop::request_focus() {
-    gtk_window_present(GTK_WINDOW(gtk_widget));
+    //JDK-8212060: Window show and then move glitch.
+    //The WindowContextBase::set_visible will take care of showing the window.
+    //The below code will only handle later request_focus.
+    if (is_visible()) {
+        gtk_window_present(GTK_WINDOW(gtk_widget));
```
kevinrushforth: This looks OK as long as you are sure that `gtk_window_present` is unneeded when the window is set visible. Otherwise, it might be safer to set a `presentPending` flag in the `else` case and call `gtk_window_present` in set_visible if `presentPending` is `true`. My suggestion is overkill if the call to `gtk_widget_show all` effectively does a present as well.

kevinrushforth:
I ran several more tests, including a full system test, and I can confirm that it fixes a very similar symptom that I see on my system. I don't see a window briefly showing at the upper-left as mentioned in the description, but I often see a very thin (maybe 3 or 4 pixels wide) and short (maybe 30 pixels tall) window briefly show of at what looks like the correct position, but obviously the wrong size, before the window is shown. This problem is now gone when running with your patch.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
tsayao: From gtk docs:

> Presents a window to the user. This may mean raising the window in the stacking order, deiconifying it, moving it to the current desktop, and/or giving it the keyboard focus, possibly dependent on the user’s platform, window manager, and preferences.
> 
> If window is hidden, this function calls gtk_widget_show() as well.

So gtk_widget_show_all is sufficient.

PR #450, opened 2019-04-21
JDK-8222214: Fix TableView content disappearing on user pressing UP

tsayao:
image

When the TableView first row is focused and no selected as in the image, when the users presses UP, the content disappears.

Why?

```
TableViewSkinBase:
    protected void onSelectAboveCell() {
        SelectionModel<S> sm = getSelectionModel();
        if (sm == null) return;

        flow.scrollTo(sm.getSelectedIndex());
    }
```

The above code is called and getSelectedIndex() is -1. So VirtualFlow.scrollTo() gets called with -1, and that's out of the VirtualFlow visibility.

```
TableViewBehaviourBase:
    protected void selectPreviousRow() {
        selectCell(-1, 0);
        if (onSelectPreviousRow != null) onSelectPreviousRow.run();
    }
```

This piece of code moves the selected cell -1 when the intent was to move to previous row. The problem is that selecCell(..., ...) bases the current row on the focused row and not the selected one.

So there was an If doing nothing if **focused** row index was zero. The change sets selected row index to ZERO if focused row is ZERO.

Fixes:
https://bugs.openjdk.java.net/browse/JDK-8222214

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,10 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <option name="DEFAULT_COMPILER" value="Javac" />
-    <excludeFromCompile>
-    </excludeFromCompile>
-    <resourceExtensions />
     <wildcardResourcePatterns>
       <entry name="!?*.java" />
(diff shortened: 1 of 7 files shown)


kevinrushforth:
OCA has been recorded

PR #451, opened 2019-04-21
JDK-8210973: Focus goes to wrong Window when dismissing an Alert

tsayao:
All windows can have an owner (WINDOW_MODAL, APPLICATION_MODAL, NONE). So when the window is closed, requestToFront() on it's owner. Previos code was not doing it when APPLICATION_MODAL.

Fixes:
https://bugs.openjdk.java.net/browse/JDK-8210973


kevinrushforth:

kevinrushforth:
OCA has been recorded

kevinrushforth:
> @pankaj-bansal can you also review this?

No need, unless you want to. @arapte and I will review it (the bug is in `WindowStage` in the  quantum toolkit).

tsayao:
> The fix looks good, and I can confirm that it fixes the bug.
> 
> Can you provide a test case for it? An automated test in `tests/system` is preferred, but if that turns out to be too difficult, then a manual test in `tests/manual` would be an OK substitute.

Test case provided.

tests/system/src/test/java/test/robot/javafx/stage/FocusParentWindowOnChildCloseTest.java
```diff
@@ -0,0 +1,159 @@
...
+            stage2.setTitle("Stage - 2");
+            StackPane root = new StackPane();
+            stage2.setScene(new Scene(root, 300, 250));
+            stage2.addEventHandler(WindowEvent.WINDOW_SHOWN, e -> Platform.runLater(startupLatch::countDown));
+            stage2.show();
+        }
+    }
+
+    public static void waitForLatch(CountDownLatch latch, int seconds, String msg) throws Exception {
+        org.junit.Assert.assertTrue("Timeout: " + msg, latch.await(seconds, TimeUnit.SECONDS));
```
kevinrushforth: Minor nit: you can just use `Assert` rather than the fully-qualified class, since it is imported.

tsayao: done!

tests/system/src/test/java/test/robot/javafx/stage/FocusParentWindowOnChildCloseTest.java
```diff
@@ -0,0 +1,159 @@
...
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import test.util.Util;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+// See JDK8210973
```
arapte: Please add a hyphen, JDK-8210973

tests/system/src/test/java/test/robot/javafx/stage/FocusParentWindowOnChildCloseTest.java
```diff
@@ -0,0 +1,159 @@
...
+            robot.mousePress(MouseButton.PRIMARY);
+            robot.mouseRelease(MouseButton.PRIMARY);
+        });
+    }
+
+    private static void fireOkInAlert() throws Exception {
+        alertCloseLatch = new CountDownLatch(1);
+        Platform.runLater(() -> {
+                    Button okButton = (Button) alert.getDialogPane().lookupButton(ButtonType.OK);
+                    System.out.println(okButton);
```
arapte: The print statement is not required here. please remove.

tests/system/src/test/java/test/robot/javafx/stage/FocusParentWindowOnChildCloseTest.java
```diff
@@ -0,0 +1,159 @@
...
+            button = new Button();
+            button.setText("Show Alert");
+            button.setOnAction(event -> showAlert(stage));
+
+            StackPane root = new StackPane();
+            root.getChildren().add(button);
+            scene = new Scene(root, 300, 250);
+            stage.setScene(scene);
+            stage.addEventHandler(WindowEvent.WINDOW_SHOWN, e -> Platform.runLater(startupLatch::countDown));
+            stage.show();
```
arapte: The test may fail, if the stage does not get shown on foreground.
Please include this call : stage.setAlwaysOnTop(true);  before stage.show();

tests/system/src/test/java/test/robot/javafx/stage/FocusParentWindowOnChildCloseTest.java
tsayao: sure!

tests/system/src/test/java/test/robot/javafx/stage/FocusParentWindowOnChildCloseTest.java
tsayao: Oops, forgot that.

tests/system/src/test/java/test/robot/javafx/stage/FocusParentWindowOnChildCloseTest.java
tsayao: Added. Did not change results.

kevinrushforth:
Latest changes to the test lgtm. I was initially a little concerned that the suggestion to use alwaysOnTop (which is the right thing to do for tests that use robot mouse presses or screen capture) might make the test unable to catch the bug if the fix were reverted, but I ran the latest test without the fix and it still catches the bug.

@arapte go ahead and integrate this fix when you get a chance.

@tsayao thank you for fixing this bug

PR #459, opened 2019-05-01
JDK-8222883: Ensemble: Update version of Lucene to 7.7.1

kevinrushforth:
https://bugs.openjdk.java.net/browse/JDK-8222883

Updates the version of Lucene used by Ensemble to the latest release of Lucene 7 which is version 7.7.1.

Notes to reviewers:

* Rather than squashing the commits into a single commit, I left them in as 4 separate commits in case anyone wanted to look at them independently, by clicking on the "Commits" tab above (click on the "Files Changed" tab to view the combined changes). The commits are as follows:

1. Bump the version numbers from 7.4.0 to 7.7.1 in all needed places
2. Add [`UPDATING-lucene.txt`](https://github.com/javafxports/openjdk-jfx/blob/7a380acc5227a4261106883cc6ea3042910dd6cd/apps/samples/Ensemble8/UPDATING-lucene.txt) with upgrade instructions
3. Fix a build problem in the Ensemble `build.xml` script preventing the index from being regenerated from a JDK that doesn't include the `javafx.*` modules (this was missed when we removed FX from JDK 11).
4. Regenerate the index files. I used JDK 11 to do this (out of an abundance of caution, not because I think it matters).

NOTE: Future upgrades should just require following the recipe in `UPDATING-lucene.txt` to do steps 1 and 4 above.

* To test this: apply the patch, run `gradle apps`, launch Ensemble8, and search for a sample or JavaFX component in the upper-right search box.


kevinrushforth:
NOTE: As with any third-party software, we need legal approval for the new version of Lucene. I will update this PR to indicate when this is done. Until then, the review can proceed, but the PR must not be merged.

aghaisas:
File - apps/samples/Ensemble8/nbproject/project.properties has windows line endings (Not introduced in this PR, but pre-existing) - This can be corrected before push.

aghaisas:
build succeeded.
Apps build also succeeded.

While running the Ensemble8, when I tried searching in the search box provided, there were exceptions on the terminal and search did not work.
Most probable cause of this is - the patch that I downloaded does not contain index binary files.

kevinrushforth:
> While running the Ensemble8, when I tried searching in the search box provided, there were exceptions on the terminal and search did not work.
> Most probable cause of this is - the patch that I downloaded does not contain index binary files.

It looks like the `.diff` patch does not include binary files but the `.patch` patch does. Try this:

https://github.com/javafxports/openjdk-jfx/pull/459.patch

kevinrushforth:
The needed legal approval is done.

PR #468, opened 2019-05-12
JDK-8223377: JavaFX can crash due to loading the wrong native libraries if system libraries are installed

matthiasblaesing:
This is an attempt to implement the change of the load order
of the native libraries as described in JDK-8223377.

The first commit refactors the library loading code into distinct
methods per loading variant. That commit does not change the
behavior of the library loading. Based on the refactored method,
the second commit implements to reordering.

For the legal part: I have an OCA on file

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
@@ -116,7 +116,7 @@ public class NativeLibLoader {
         //   (only on non-modular builds)
         // - if the native library comes bundled as a resource it is extracted
         //   and loaded
-        // - the jna.library.path is searched for the library in definition
+        // - the java.library.path is searched for the library in definition
         //   order
         // - the library is loaded via System#loadLibrary
         // - on iOS native library is staticly linked and detected from the
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
@@ -111,9 +111,16 @@ public class NativeLibLoader {
     }
 
     private static void loadLibraryInternal(String libraryName, List<String> dependencies, Class caller) {
-        // Look for the library in the same directory as the jar file
-        // containing this class.
-        // If that fails, then try System.loadLibrary.
+        // The search order for native library loading is:
+        // - try to load the native library from the same folder as this jar
+        //   (only on non-modular builds)
+        // - if the native library comes bundled as a resource it is extracted
+        //   and loaded
+        // - the jna.library.path is searched for the library in definition
+        //   order
+        // - the library is loaded via System#loadLibrary
+        // - on iOS native library is staticly linked and detected from the
+        //   existence of a JNI_OnLoad_libraryname funtion
         try {
             // FIXME: JIGSAW -- We should eventually remove this legacy path,
             // since it isn't applicable to Jigsaw.
@@ -121,6 +128,11 @@ public class NativeLibLoader {
         } catch (UnsatisfiedLinkError ex) {
             if (verbose && !usingModules) {
                 System.err.println("WARNING: " + ex);
+            }
+
+            // if the library is available in the jar, copy it to cache and load it from there
+            if (loadLibraryFromResource(libraryName, dependencies, caller)) {
+                return;
             }
 
             // NOTE: First attempt to load the libraries from the java.library.path.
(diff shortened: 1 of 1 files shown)


kevinrushforth:
I can confirm that you have an OCA on file. I will review it (and ask either @tiainen or @johanvos to review as well).

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -373,18 +387,21 @@ private static void loadLibraryFullPath(String libraryName) {
...
+            System.load(libFileName);
+            if (verbose) {
+                System.err.println("Loaded " + libFile.getAbsolutePath()
+                    + " from relative path");
+            }
+        } catch (IOException | UnsatisfiedLinkError | URISyntaxException e) {
+            if (verbose) {
+                System.err.println("WARNING: " + e);
+            }
+            if(e instanceof Exception) {
```
nlisker: Space after `if`.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -111,90 +112,98 @@ public static synchronized void loadLibrary(String libname, List<String> depende
...
-                    } catch (UnsatisfiedLinkError ex3) {
-                        throw ex3;
-                    }
+                    System.err.println("Loaded " + libFile.getAbsolutePath()
+                        + " from java.library.path");
                 }
-                // Rethrow exception
-                throw ex2;
+                return;
+            }catch (UnsatisfiedLinkError ex) {
```
nlisker: Space before `catch`.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -111,90 +112,98 @@ public static synchronized void loadLibrary(String libname, List<String> depende
...
+    }
 
-            // Finally we will use System.loadLibrary.
+    /**
+     * This allows FX to find more recent versions of the shared libraries from
+     * java.library.path instead of ones that might be part of the JRE
+     */
+    private static void loadLibraryFromFromJavaLibraryPath(String libraryName) {
+        String [] libPath = initializePath("java.library.path");
+        for (int i = 0; i<libPath.length; i++) {
```
nlisker: Can be replaced with a for-each loop.

kevinrushforth:
Two general comments:

1. It would be good to add a comment at the beginning of `loadLibraryInternal` that lists the order in which the method will try to find the library.

2. The refactoring makes it a little harder to verify that there are no unintended changes in behavior. One thing I see that would be a minor change is that the exception messages will now say the following when it fails to find or load the library:

```
Not on iOS or libraryName did not contain hyphens
```

I think that providing a fix without the refactoring might be easier to review. It would seem to be as simple as moving the `loadLibraryFromResource` call before trying `java.library.path`.

matthiasblaesing:
Thank you both for review. Kevin you have a point, that the refactoring is not necessary for this use-case. Based on your comment I updated the changeset with that approach. Ignoring the new/updated comment, this change only moves the extraction before the codepath, that tries the `java.library.path`.

This should be save, as the `installLibraryFromResource` catches all throwables and just indicates by its return value whether or not loading succeeded. If loading succeeded `loadLibraryInternal` will be exited through an early return, if it failed, processing will continue with the `java.library.path`.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -111,9 +111,16 @@ public static synchronized void loadLibrary(String libname, List<String> depende
...
     private static void loadLibraryInternal(String libraryName, List<String> dependencies, Class caller) {
-        // Look for the library in the same directory as the jar file
-        // containing this class.
-        // If that fails, then try System.loadLibrary.
+        // The search order for native library loading is:
+        // - try to load the native library from the same folder as this jar
+        //   (only on non-modular builds)
+        // - if the native library comes bundled as a resource it is extracted
+        //   and loaded
+        // - the jna.library.path is searched for the library in definition
```
kevinrushforth: That should be `java.library.path`

matthiasblaesing: Sorry, I had much more need to write `jna.library.path`, than `java.library.path`. Updated accordingly.

johanvos:
The only issue I see is that there have been issues reported with loading libraries from resources that we do not understand or can reproduce yet. If we make loading libraries from resources the default, we might run into more cases where these issues arise. But then it also makes more sense to fix them.

Apart from that remark, +1 from me.

kevinrushforth:
> If we make loading libraries from resources the default, we might run into more cases where these issues arise.

A fair point, but this change in default only affects the corner case of using the maven jars (which include the native libs as resources), while pointing to a set of DLLs from somewhere else, either installed in the JDK or via java.library.path. Since that represents a mismatch between class files and DLLs, it isn't really any worse in practice than a failure to load from resources.

Thanks.

matthiasblaesing:
Thank you both for reviewing!

kevinrushforth:
Certainly. And thank you for fixing this bug.

Issue #469, opened 2019-05-13
[JDK-8223760] static libs for macos


johanvos:
See https://bugs.openjdk.java.net/browse/JDK-8223760: provide a mechanism to create static libs for macosx

kevinrushforth:
I think this issue can be closed, too...

johanvos:
> I think this issue can be closed, too...

true.

johanvos:
Fixed via https://bugs.openjdk.java.net/browse/JDK-8223760

PR #470, opened 2019-05-13
JDK-8223760: allow to build static libraries for macosx.

johanvos:
This requires build logic and the presence of JNI_OnLoad_L functions for library L

Related to #469

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -396,6 +396,9 @@ ext.IS_COMPILE_HARFBUZZ = Boolean.parseBoolean(COMPILE_HARFBUZZ)
 defineProperty("COMPILE_PARFAIT", "false")
 ext.IS_COMPILE_PARFAIT = Boolean.parseBoolean(COMPILE_PARFAIT)
 
+defineProperty("STATIC_BUILD", "false")
+ext.IS_STATIC_BUILD = Boolean.parseBoolean(STATIC_BUILD)
+
 // BUILD_TOOLS_DOWNLOAD_SCRIPT specifies a path of a gradle script which downloads
 // required build tools.
 defineProperty("BUILD_TOOLS_DOWNLOAD_SCRIPT", "")
--- a/build.gradle
+++ b/build.gradle
@@ -399,6 +399,10 @@ ext.IS_COMPILE_PARFAIT = Boolean.parseBoolean(COMPILE_PARFAIT)
 defineProperty("STATIC_BUILD", "false")
 ext.IS_STATIC_BUILD = Boolean.parseBoolean(STATIC_BUILD)
 
+if (IS_STATIC_BUILD && IS_COMPILE_MEDIA) {
+    throw new GradleException("Can not have COMPILE_MEDIA when STATIC_BUILD is enabled");
+}
+
 // BUILD_TOOLS_DOWNLOAD_SCRIPT specifies a path of a gradle script which downloads
 // required build tools.
 defineProperty("BUILD_TOOLS_DOWNLOAD_SCRIPT", "")
@@ -2797,7 +2801,6 @@ project(":media") {
                          "OUTPUT_DIR=${nativeOutputDir}", "BUILD_TYPE=${buildType}", "BASE_NAME=jfxmedia",
                          "COMPILE_PARFAIT=${compileParfait}",
                          IS_64 ? "ARCH=x64" : "ARCH=x32",
-                         STATIC_BUILD ? "STATIC_BUILD=true" : "STATIC_BUILD=false",
                         "CC=${mediaProperties.compiler}", "LINKER=${mediaProperties.linker}")
 
                     if (t.name == "win") {
@@ -2825,7 +2828,6 @@ project(":media") {
                         commandLine ("make", "${makeJobsFlag}", "-C", "${nativeSrcDir}/gstreamer/projects/${projectDir}/gstreamer-lite")
                         args("OUTPUT_DIR=${nativeOutputDir}", "BUILD_TYPE=${buildType}", "BASE_NAME=gstreamer-lite",
                              IS_64 ? "ARCH=x64" : "ARCH=x32", "CC=${mediaProperties.compiler}",
-                             STATIC_BUILD ? "STATIC_BUILD=true" : "STATIC_BUILD=false",
                              "AR=${mediaProperties.ar}", "LINKER=${mediaProperties.linker}")
 
                         if (t.name == "win") {
@@ -2844,7 +2846,6 @@ project(":media") {
                         commandLine ("make", "${makeJobsFlag}", "-C", "${nativeSrcDir}/gstreamer/projects/${projectDir}/fxplugins")
                         args("OUTPUT_DIR=${nativeOutputDir}", "BUILD_TYPE=${buildType}", "BASE_NAME=fxplugins",
                              IS_64 ? "ARCH=x64" : "ARCH=x32",
-                             STATIC_BUILD ? "STATIC_BUILD=true" : "STATIC_BUILD=false",
                              "CC=${mediaProperties.compiler}", "AR=${mediaProperties.ar}", "LINKER=${mediaProperties.linker}")
 
                         if (t.name == "win") {
@@ -3162,7 +3163,6 @@ project(":media") {
                                         args("CC=${mediaProperties.compiler}", "LINKER=${mediaProperties.linker}",
                                              "OUTPUT_DIR=${nativeOutputDir}", "BUILD_TYPE=${buildType}",
                                              "BASE_NAME=avplugin", "VERSION=${version}", "LIBAV_DIR=${libavDir}",
-                                             STATIC_BUILD ? "STATIC_BUILD=true" : "STATIC_BUILD=false",
                                              "SUFFIX=", IS_64 ? "ARCH=x64" : "ARCH=x32")
                                     }
                                 }
@@ -3177,7 +3177,6 @@ project(":media") {
                                         args("CC=${mediaProperties.compiler}", "LINKER=${mediaProperties.linker}",
                                              "OUTPUT_DIR=${nativeOutputDir}", "BUILD_TYPE=${buildType}",
                                              "BASE_NAME=avplugin", "VERSION=${version}", "LIBAV_DIR=${libavDir}",
-                                              STATIC_BUILD ? "STATIC_BUILD=true" : "STATIC_BUILD=false",
                                              "SUFFIX=-ffmpeg", IS_64 ? "ARCH=x64" : "ARCH=x32")
                                     }
                                 }
@@ -3192,7 +3191,6 @@ project(":media") {
                                         args("CC=${mediaProperties.compiler}", "LINKER=${mediaProperties.linker}",
                                              "OUTPUT_DIR=${nativeOutputDir}", "BUILD_TYPE=${buildType}",
                                              "BASE_NAME=avplugin", "VERSION=${version}", "LIBAV_DIR=${libavDir}",
-                                             STATIC_BUILD ? "STATIC_BUILD=true" : "STATIC_BUILD=false",
                                              "SUFFIX=-ffmpeg", IS_64 ? "ARCH=x64" : "ARCH=x32")
                                     }
                                 }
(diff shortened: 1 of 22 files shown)


kevinrushforth:
JBS issue: https://bugs.openjdk.java.net/browse/JDK-8223760

modules/javafx.graphics/src/main/native-prism-es2/macosx/MacGLFactory.c
```diff
@@ -33,6 +33,24 @@
 #include "../PrismES2Defs.h"
 #include "com_sun_prism_es2_MacGLFactory.h"
 
+#ifdef STATIC_BUILD
+JNIEXPORT jint JNICALL
+JNI_OnLoad_prism_es2(JavaVM *vm, void * reserved) {
+fprintf(stderr, "ONLOAD PRISM_ES2\n");
```
kevinrushforth: Maybe qualify this printf by `#if VERBOSE` or similar?

buildSrc/mac.gradle
```diff
@@ -117,16 +117,24 @@ def ccFlags = [
...
         "-framework", "ApplicationServices",
         "-framework", "OpenGL",
         "-framework", "QuartzCore",
         "-framework", "Security",
         "-dynamiclib", "-lobjc"].flatten();
+def dynamicLinkFlagsAlt = ["-dynamiclib", commonParams].flatten()
+
+def staticLinkFlags = ["-static"].flatten();
+
+def linkFlags = IS_STATIC_BUILD ? staticLinkFlags : dynamicLinkFlags
```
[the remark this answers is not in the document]
kevinrushforth: Answering my own question, I can confirm that the build fails with `-PCOMPILE_MEDIA=true`.

modules/javafx.graphics/src/main/native-font/MacFontFinder.c
```diff
@@ -30,7 +30,7 @@
 #include <jni.h>
 #include <com_sun_javafx_font_MacFontFinder.h>
 
-#if TARGET_OS_IPHONE /* iOS */
+#if TARGET_OS_IPHONE || defined(STATIC_BUILD) /* iOS or static build */
```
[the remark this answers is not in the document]
johanvos: That was a nice side-effect. We only need CoreText.h, so those 2 bigger imports can be replaced with #import <CoreText/CoreText.h> (but that's for another discussion)

buildSrc/mac.gradle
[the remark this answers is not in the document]
johanvos: The native media files now build static libraries as well.

kevinrushforth:
@johanvos Once you are ready for formal review, remove the WIP label and I'll look at the updated version.

kevinrushforth:
@sashamatveev Can you also review this?

buildSrc/linux.gradle
```diff
@@ -37,7 +37,7 @@ LINUX.compileSWT = true;
 LINUX.libDest = "lib"
 
 // Lambda for naming the generated libs
-LINUX.library = { name -> return "lib${name}.so" as String }
+LINUX.library = { name ->  return (IS_STATIC_BUILD ? "lib${name}.a" : "lib${name}.so")  as String }
```
sashamatveev: Extra spaces before "return" and between ")" and  "as String".

buildSrc/linux.gradle
```diff
@@ -192,15 +200,18 @@ setupTools("linux_freetype_tools",
...
         } else {
             throw new IllegalStateException("Linux freetype packages not found.\nIf freetype pacakges are installed, please remove the build directory and try again.")
         }
     }
 )
 
 def compiler = IS_COMPILE_PARFAIT ? "parfait-gcc" : "${toolchainDir}gcc";
-def linker = IS_COMPILE_PARFAIT ? "parfait-g++" : "${toolchainDir}g++";
+def AR = "ar"
+def linker = IS_STATIC_BUILD ? AR : IS_COMPILE_PARFAIT ? "parfait-g++" : "${toolchainDir}g++";
```
sashamatveev: Do we need this extra def AR = "ar"? I think we can just use "ar" to be consistent with "parfait-g++".

buildSrc/linux.gradle
```diff
@@ -224,23 +235,27 @@ LINUX.glass.glass.nativeSource = ft_gtk_launcher.getFiles()
...
 LINUX.glass.glassgtk2 = [:]
 LINUX.glass.glassgtk2.nativeSource =  ft_gtk.getFiles()
 LINUX.glass.glassgtk2.compiler = compiler
-LINUX.glass.glassgtk2.ccFlags = [ccFlags, gtk2CCFlags, "-Werror"].flatten()
+LINUX.glass.glassgtk2.ccFlags = IS_STATIC_BUILD ?
+        ["-fno-threadsafe-statics", ccFlags, gtk2CCFlags].flatten() :
+        [ccFlags, gtk2CCFlags, "-Werror"].flatten()
 LINUX.glass.glassgtk2.linker = linker
-LINUX.glass.glassgtk2.linkFlags = [linkFlags, gtk2LinkFlags ].flatten()
+LINUX.glass.glassgtk2.linkFlags = IS_STATIC_BUILD ? linkFlags :[linkFlags, gtk2LinkFlags ].flatten()
```
sashamatveev: Space after ":". Also, formatting at lines 258, 303, 320 and 328 with missing spaces.

buildSrc/src/main/groovy/com/sun/javafx/gradle/LinkTask.groovy
```diff
@@ -39,12 +39,18 @@ class LinkTask extends DefaultTask {
...
+              args("rcs");
+              args("$lib");
+            }
             // Exclude parfait files (.bc)
             args(objectDir.listFiles().findAll{ !it.getAbsolutePath().endsWith(".bc") });
             if (project.IS_WINDOWS) {
                 args("/out:$lib");
             } else {
-                args("-o", "$lib");
+                if(! ((project.IS_LINUX) && (project.IS_STATIC_BUILD))) {
```
sashamatveev: Space after "if".

sashamatveev:
Changes to media looks fine, however media will not work when compiled statically, so I have three suggestions:
1) Do not build media statically and exclude it from static build.
2) Add support for static plugin initialization.
gchar *gstlite_plugins_list[] in gstregistry.c contains list of plugins we loading dynamically (they needs to be dynamic libraries). gst_registry_scan_path_level() in same file has code which loads these plugins dynamically. For libavplugin (linux only) we will first determine which libavcodec version is available and then load appropriate libavplugin.
GStreamer provided plugins are loaded statically and linked into GStreamer library itself. See lite_plugins_init() on how it is done.
So, for fxplugins we need to do something similar. Register fxplugins_init using gst_plugin_register_static() similar to lite_plugins_init().
For libavplugin it will be more complicated. First we need to make sure all init functions are unique (audiodecoder_plugin_init, videodecoder_plugin_init and mpegts_demuxer_plugin_init) and call statically required init function based on libavcodec available on system. Also, symbol collision for other libavplugin needs to be resolved. It might be easy to just support one libavcodec which is known available on system.
3) Only build jfxmedia layer statically since it is only layer which has JNI calls and keep gstreamer-lite, glib-lite and other media libraries dynamic.

kevinrushforth:
Thanks for the review. I think option 1 or option 3 would be best. Option 2 would be an intrusive change.

XQDD:
How about Linux?

johanvos:
> Thanks for the review. I think option 1 or option 3 would be best. Option 2 would be an intrusive change.

I agree with that. Hence, I'll remove the changes in gstreamer-lite etc?

johanvos:
> How about Linux? I want to use the static libs in the old Arm CPU (32-bit Armel) because of
> 
> ```
> /lib/libc.so.6: version `GLIBC_2.12' not found
> ```

The PR has changes for Linux as well, and allows to build static versions of the JavaFX libs, but that is not related to the version of GLIBC

XQDD:
> > How about Linux? I want to use the static libs in the old Arm CPU (32-bit Armel) because of
> > ```
> > /lib/libc.so.6: version `GLIBC_2.12' not found
> > ```
> 
> The PR has changes for Linux as well, and allows to build static versions of the JavaFX libs, but that is not related to the version of GLIBC

I know, the reason why I choose static lib is the /lib can't be update.

kevinrushforth:
> I agree with that. Hence, I'll remove the changes in gstreamer-lite etc?

Yes, that seems best. If going with option 3, then you would leave it in for jfxmedia. If going with option 1, then remove from jfxmedia, too.

johanvos:
> > I agree with that. Hence, I'll remove the changes in gstreamer-lite etc?
> 
> Yes, that seems best. If going with option 3, then you would leave it in for jfxmedia. If going with option 1, then remove from jfxmedia, too.

There are a number of options to implement this. We can either create the dynamic libs for those components, or avoid building them. The latter would affect build.gradle alot, so I think it is safest to just build the dynamic libs in this case (without using them).

kevinrushforth:
Yes, I think the option of building, but not using, is probably best.

johanvos:
> Yes, I think the option of building, but not using, is probably best.

I'm going for option 1 for now. Creating a static build for jfxmedia is not trivial, as it depends on the subcomponents. Rather than making too many changes, I think at this point we can simply state that in case of static builds, the media component will not be available.
It is still possible to use the shared libraries of the media components at runtime though.

buildSrc/mac.gradle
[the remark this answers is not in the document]
johanvos: As a follow-up to recent discussions, I will revert the changes to enable static builds for the media files, and exclude the combination of COMPILE_MEDIA and STATIC_BUILD.

modules/javafx.graphics/src/main/native-prism-es2/macosx/MacGLFactory.c
johanvos: should be fixed

buildSrc/mac.gradle
```diff
@@ -214,7 +222,11 @@ MAC.font.lib = "javafx_font"
 MAC.media = [:]
 MAC.media.compiler = "${compiler} ${ccBaseFlags.join(" ")}"
 //MAC.media.ccFlags = ccBaseFlags
-MAC.media.linker = "${linker} ${commonParams.join(" ")}"
+if (IS_STATIC_BUILD) {
+    MAC.media.linker = "${linker}"
+} else {
+    MAC.media.linker = "${linker} ${commonParams.join(" ")}"
+}
```
sashamatveev: Is this used by media only? If yes, then I think it would better to revert it as well.

build.gradle
```diff
@@ -396,6 +396,13 @@ ext.IS_COMPILE_HARFBUZZ = Boolean.parseBoolean(COMPILE_HARFBUZZ)
 defineProperty("COMPILE_PARFAIT", "false")
 ext.IS_COMPILE_PARFAIT = Boolean.parseBoolean(COMPILE_PARFAIT)
 
+defineProperty("STATIC_BUILD", "false")
+ext.IS_STATIC_BUILD = Boolean.parseBoolean(STATIC_BUILD)
+
+if (IS_STATIC_BUILD && IS_COMPILE_MEDIA) {
```
kevinrushforth: Do you need to do the same for WebKit?

johanvos:
> I haven't had time to test it yet. The main test to verify will be that the changes have no effect in the default case -- meaning in the absence of `-PSTATIC_BUILD=true` -- on all three platforms (Windows should be trivial to prove since there are no changes to the Windows build args).

The changes in native code are all between
`#ifdef STATICBUILD
#endif`

The changes in the build.gradle. mac.gradle and linux.gradle are guarded by
`IS_STATIC_BUILD` and in case this is false, the statements are exactly the same as before (plus/minus some whitespace)

build.gradle
johanvos: No, WebKit has its own build system which we don't touch. Hence, we're making shared libs for webkit, even if STATIC_BUILD is set. A static library for webkit is probably not very practical (on iOS and Android, we use the native webkit implementations of the device).

We can modify the build to create a static build of webkit in a follow-up issue, but even when the javafx-specifc libs are static, it still makes sense to have a dynamic version of the webkit lib.

kevinrushforth:
@johanvos - I think this still needs an RFR email to openjfx-dev (unless I missed it)

build.gradle
kevinrushforth: I just figured that if that you already fail the build if you specify both `STATIC_BUILD=true` and `COMPILE_MEDIA=true` then it would make sense to also fail the build if `COMPILE_WEBKIT=true`, but as long as it doesn't break the build, then it should be OK.

build.gradle
```diff
@@ -396,6 +396,13 @@ ext.IS_COMPILE_HARFBUZZ = Boolean.parseBoolean(COMPILE_HARFBUZZ)
 defineProperty("COMPILE_PARFAIT", "false")
 ext.IS_COMPILE_PARFAIT = Boolean.parseBoolean(COMPILE_PARFAIT)
 
+defineProperty("STATIC_BUILD", "false")
+ext.IS_STATIC_BUILD = Boolean.parseBoolean(STATIC_BUILD)
```
kevinrushforth: It might be useful to print the value of static build -- you could add it to the block of `logger.quiet` statements somewhere around line 1320.

build.gradle
kevinrushforth: I can confirm that `gradle -PSTATIC_BUILD=true -PCOMPILE_WEBKIT=true` doesn't blow up the build on Mac, but it doesn't copy the .dylib file to the sdk/lib directory either. In any case, this can be handled as a follow-up issue.

kevinrushforth:
Everything checks out in the default (no static build) case. Building the sdk works fine on Mac and Linux with `-PSTATIC_BUILD=true`.

kevinrushforth:
Since this is already pushed (via HG), I think this PR can be closed.

johanvos:
Fixed via https://bugs.openjdk.java.net/browse/JDK-8223760

PR #473, opened 2019-05-16
JDK-8211302: DND does not work on GTK3

tsayao:
Took me a while and some head scratching..

It's working on ubuntu 19.04.

Needs some testing (on other systems).

Will provide test cases.

To explain some changes: Please see https://developer.gnome.org/gdk3/stable/GdkDevice.html#gdk-device-grab where it recommends getting the device from the GdkEvent (when there is an event).

Funny thing: You will notice that I have changed the DND window move from -100, -100 to -99, -99. On GTK3 it does not work with -100.


pankaj-bansal:
@tsayao  I have done some initial testing by running some manual tests I have for Dnd. The patch does fix the change in my 18.04 machine with gtk3.20. I tested on 16.04 LTS as well where the dnd is working fine as of now without patch. I see that the patch is causing some issues there as the dnd stop if I move my mouse a bit faster of I take mouse over different windows. The dnd stops working if I cross a window edge. These must be small issue and can be fixed.

kevinrushforth:
I'll do some testing of this as well on an older Ubuntu (I have 16.04) to make sure it doesn't fail with older GTK3 systems.

> Funny thing: You will notice that I have changed the DND window move from -100, -100 to -99, -99. On GTK3 it does not work with -100.

Any ideas why? Is this just a workaround for a GTK bug or is there some significance to the change that can be explained by the spec for that method?

tsayao:
> Any ideas why? Is this just a workaround for a GTK bug or is there some significance to the change that can be explained by the spec for that method?

Well, I could not get it working, so I decided to read Gtk code and see how they do it for Gtk DND. They used a similar method, but with a Gtk Window instead of a Gdk Window and -99 positioning. So I copied their method and keep removing the differences from glass to gtk method until i found out that the difference was  the positioning value. Not sure why tho. "Greping" Gtk code for "99" did not show any fixed values.

kevinrushforth:
I don't see any issue on Ubuntu 16.04, which has an older version of GTK 3 (3.18). I don't have easy access to an Ubuntu 18.04 or 19.10 system right now.

pankaj-bansal:
> I don't see any issue on Ubuntu 16.04, which has an older version of GTK 3 (3.18). I don't have easy access to an Ubuntu 18.04 or 19.10 system right now.

I can see some issue on 16.04. Which test case are you using? I am using the one attached in https://bugs.openjdk.java.net/browse/JDK-8211302 and I see issue on 16.04. After you start the dnd, take the cursor on terminal window and then try to come back to test window. The dnd always stops when it leaves the terminal window. I can always reproduce the issue. If you still not able to see this, I will try to create a small screen capture for you of what I see.
All works fine on 18.04 LTS.

kevinrushforth:
@pankaj-bansal yes, you are right. I saw that yesterday, but had seen something similar with the existing code, so I didn't look closely. Perhaps it is related to the point @tsayao made about not handling enter / exit correctly?

tsayao:
I fixed it for Ubuntu 16.04, but the fix breaks DND on Gtk 3.20+.

The current problem is that gdk does not send GDK_DROP_FINISHED event. The weird thing is that the X event is firing:

Gdk-Message: 10:32:43.594: XdndFinished: dest_window: 0x280000a

(set env var GDK_DEBUG=dnd)

tsayao:
Could not find a way to get GDK_DROP_FINISHED event to fire. On the wayland dnd code in gdk it's not even called (it seems). So this fix seems fine (not relying on the event to end dnd).

tsayao:
Drag seems ok, Drop still not ok - will have to look into it.

tsayao:
DND Test:

```
import javafx.application.Application;
import javafx.event.EventHandler;
import javafx.scene.Group;
import javafx.scene.Scene;
import javafx.scene.input.Dragboard;
import javafx.scene.input.ClipboardContent;
import javafx.scene.input.TransferMode;
import javafx.scene.paint.Color;
import javafx.scene.text.Text;
import javafx.stage.Stage;

public class HelloDragDrop extends Application {

    @Override public void start(Stage stage) {
        stage.setTitle("Hello Drag And Drop");

        Group root = new Group();
        Scene scene = new Scene(root, 400, 200);
        scene.setFill(Color.LIGHTGREEN);

        final Text source = new Text(50, 100, "DRAG ME");
        source.setScaleX(2.0);
        source.setScaleY(2.0);

        final Text target = new Text(250, 100, "DROP HERE");
        target.setScaleX(2.0);
        target.setScaleY(2.0);

        source.setOnDragDetected(event -> {
            Dragboard db = source.startDragAndDrop(TransferMode.ANY);

            ClipboardContent content = new ClipboardContent();
            content.putString(source.getText());
            db.setContent(content);

            event.consume();
        });

        target.setOnDragOver(event -> {
            if (event.getGestureSource() != target &&
                    event.getDragboard().hasString()) {
                event.acceptTransferModes(TransferMode.COPY_OR_MOVE);
            }

            event.consume();
        });

        target.setOnDragEntered(event -> {
            if (event.getGestureSource() != target &&
                    event.getDragboard().hasString()) {
                target.setFill(Color.GREEN);
            }

            event.consume();
        });

        target.setOnDragExited(event -> {
            /* mouse moved away, remove the graphical cues */
            target.setFill(Color.BLACK);

            event.consume();
        });

        target.setOnDragDropped(event -> {
            Dragboard db = event.getDragboard();
            boolean success = false;
            if (db.hasString()) {
                target.setText(db.getString());
                success = true;
            }
            event.setDropCompleted(success);

            event.consume();
        });

        source.setOnDragDone(event -> {
            if (event.getTransferMode() == TransferMode.MOVE) {
                source.setText("");
            }

            event.consume();
        });

        root.getChildren().add(source);
        root.getChildren().add(target);
        stage.setScene(scene);
        stage.show();
    }

        Application.launch(args);
    }
}
```

tsayao:
Now it works for the gtk3 variations (tested on ubuntu 16.04 and 19.04).

I just need some hints on how to do the robot test. DND does not work with robot.mousePress(),  robot.mouseMove() and robot.mouseRelease().

kevinrushforth:
That's great news! We'll test it and review it.

As for automated testing, we've never found a good way to do this for Drag and Drop, so if it needs to be a manual test, that's OK.

tsayao:
@kevinrushforth Wayland is not perfect yet, it does not finish the DND because it's not sending the GDK SELECT event (don't know why yet).

But many things are not ok on Wayland yet (not related to DND).

kevinrushforth:
@tsayao yes, there are problems with Wayland, which is why it isn't a supported platform for the JDK. FX generally runs now (whereas we used to crash), but there are several problems, some of which will require Wayland changes (they lack the needed support for Robot, for example).

kevinrushforth:
@pankaj-bansal can you also review the updated version of this?

modules/javafx.graphics/src/main/native-glass/gtk/glass_general.cpp
```diff
@@ -584,17 +584,23 @@ glass_gdk_mouse_devices_ungrab() {
...
+        gdk_device_grab(device, window, GDK_OWNERSHIP_NONE, FALSE,
+                        (GdkEventMask)
+                             (GDK_POINTER_MOTION_MASK
+                                 | GDK_BUTTON_MOTION_MASK
+                                 | GDK_BUTTON1_MOTION_MASK
+                                 | GDK_BUTTON2_MOTION_MASK
+                                 | GDK_BUTTON3_MOTION_MASK
+                                 | GDK_BUTTON_RELEASE_MASK),
+                           cursor, GDK_CURRENT_TIME);
 #else
```
pankaj-bansal: Is there any specific reason to change GDK_ALL_EVENTS_MASK to these specific events?

tsayao: Yes. Those are the same events used on gtk2, so I thought only to grab the events needed. Since gtk2 DND is well tested, I thought it's the best choice of events.

tsayao:
@kevinrushforth, @pankaj-bansal: This fix still uses "unmanaged" GDK DND. On version 3.20+ they changed GDK to use a "managed" version where GDK handles the DND and fire some events thru glib signals. The ideal fix would be to port DND to use the "managed vesion" of GDK DND or even use GTK DND (if possible). I suspect wayland DND will only work properly on "managed" DND. But since OpenJfx supports earlier gtk 3 versions, this needs more thinking.

pankaj-bansal:
I have tested this on Ubuntu 18.04 and 16.04. All works fine for me.

tsayao:
DragView is not working yet. Please, wait before merging.

tsayao:
Fixed DragView and updated the test to show the dragging content (using the DragView).

tsayao:
@pankaj-bansal Can you redo the testing?

I have tested on Ubuntu 16.04, 18.04 and 19.04 - all ok.

pankaj-bansal:
@tsayao I tried the latest patch on Ubuntu 16.04 and 18.04. All works fine for me. I have one question. I see that if we are in dnd and I drag the mouse outof test window, the cursor changes from hand-cursor to normal arrow-cursor. If you bring it back to test window, it again changes to hand-cursor. This is not consistent with gtk2 as during dnd, the cursor is always hand-cursor, even if you drag the mouse out of test window. Can you please check this and let me know if you are expecting this to happen. If I am not clear in explaining, I can create a screen capture video for you. I am using the test attached to the https://bugs.openjdk.java.net/browse/JDK-8211302

tsayao:
@pankaj-bansal You're right. Fixed it. Turns out gdk docs are not totally accurate (or maybe ubuntu changes cursor names).

pankaj-bansal:
@tsayao Yes, the issue is fixed now. I am currently running the full system test on Ubuntu. I am sure all will work fine, but need to run the tests.
I see that you have added some .idea files in the patch. Please remove those files.

pankaj-bansal:
@tsayao : I see that you created one more branch for using GTK instead of GDK for DND. Is there specific reason for trying to do this as this will be quite big change and the fix in current state is working fine?

tsayao:
@pankaj-bansal I have created a separated PR so you (or any person that can make the choice) could choose. One option is to merge this one and then the Gtk port when it's ready (and if decided to do so).

kevinrushforth:
> One option is to merge this one and then the Gtk port when it's ready (and if decided to do so).

Indeed, that is the only viable option for the short term. Longer-term we could consider what you propose in #490 , but that's more like an enhancement request than a bug fix (although it has some benefits as you note).

kevinrushforth:
@pankaj-bansal Once we are both done reviewing / testing this PR, let's get it integrated. As both you and I noted separately, the change in PR #490 is much larger, and would be better considered separately, in the context of making the Glass GTK port easier to maintain, and not as a fix for this bug.

pankaj-bansal:
I have run the full system test on all three platforms and run some manual dnd tests. All works fine for me.

Issue #439, opened 2019-04-10
[JDK-8222788] javafx.web build fails on XCode 10.2


johanvos:
Since macOS 10.2, bad_optional_access is included indirectly by LLIntOffsetsExtractor.cpp from both wtf/Optional.h and the system definitions in /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/optional

```
[ 64%] Building CXX object Source/JavaScriptCore/CMakeFiles/LLIntOffsetsExtractor.dir/llint/LLIntOffsetsExtractor.cpp.o
In file included from /Users/johan/open-jfx/openjdk-jfx/modules/javafx.web/src/main/native/Source/JavaScriptCore/llint/LLIntOffsetsExtractor.cpp:28:
In file included from /Users/johan/open-jfx/openjdk-jfx/modules/javafx.web/src/main/native/Source/JavaScriptCore/bytecode/ArrayProfile.h:29:
In file included from /Users/johan/open-jfx/openjdk-jfx/modules/javafx.web/src/main/native/Source/JavaScriptCore/runtime/Structure.h:28:
In file included from /Users/johan/open-jfx/openjdk-jfx/modules/javafx.web/src/main/native/Source/JavaScriptCore/runtime/ClassInfo.h:25:
In file included from /Users/johan/open-jfx/openjdk-jfx/modules/javafx.web/src/main/native/Source/JavaScriptCore/interpreter/CallFrame.h:30:
In file included from /Users/johan/open-jfx/openjdk-jfx/modules/javafx.web/src/main/native/Source/JavaScriptCore/runtime/VM.h:40:
In file included from /Users/johan/open-jfx/openjdk-jfx/modules/javafx.web/src/main/native/Source/JavaScriptCore/runtime/FunctionHasExecutedCache.h:29:
In file included from /Users/johan/open-jfx/openjdk-jfx/modules/javafx.web/build/mac/Release/DerivedSources/ForwardingHeaders/wtf/StdUnorderedMap.h:28:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/unordered_map:385:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/__node_handle:16:
[...]

PR #474, opened 2019-05-16
JDK-8222788: javafx.web build fails on XCode 10.2

arajkumar:
It is been already addressed in https://trac.webkit.org/changeset/235661.

We just need to cherry-pick the above changeset from upstream WebKit repo.

JBS link: https://bugs.openjdk.java.net/browse/JDK-8222788

It will address the issue #439 .


johanvos:
I will build on XCode 10.2 to check but I'm pretty sure it will work. I wonder though if it will still work on older versions, as the optional access was probably used?

arajkumar:
@johanvos , `bad_optional_access` never been used within WebKit.

PR #479, opened 2019-05-23
JDK-8224636: CSS "pointer-events" property "stroke" is not respected for SVG renderings

matthiasblaesing:
Pointer events for SVG elements, that have pointer-events enabled for
clicks on the stroke were not correctly recorded.

It was found, that this sequence is responsible:
RenderSVGShape::nodeAtFloatPoint
calls into RenderSVGShape::strokeContains
calls into RenderSVGShape::shapeDependentStrokeContains
calls into Path::strokeContains

The problem is, that Path::strokeContains only a stub (found in 
TemporaryLinkStubsJava.cpp):

bool Path::strokeContains(StrokeStyleApplier*, const FloatPoint&) const
{
    notImplemented();
    return false;
}

To fix this, the implementation is added. Using
BasicStroke#createCenteredStrokedShape the shape is created, that
encloses the stroke of the shape it is based on. The stroke shape is
then used to test if the given point is contained in that shape.

JBS link: https://bugs.openjdk.java.net/browse/JDK-8224636


modules/javafx.web/src/test/java/test/javafx/scene/web/PointerEventsTest.java
```diff
@@ -0,0 +1,233 @@
...
+            primaryStage.show();
+
+            view.getEngine().getLoadWorker().stateProperty().addListener(new ChangeListener<State>() {
+                @Override
+                public void changed(ObservableValue<? extends State> ov, State t, State t1) {
+                    if(t1 == State.SUCCEEDED) {
+                        waitTestDocumentLoaded.countDown();
+                    }
+                }
+            });
```
arajkumar: All these boilerplate code has been already taken care by TestBase, just derive it from TestBase and use it's helper methods. Refer SVGTest.java for more information.

`testClickOnStrokePointerEventsStroke` shall be simplified like below,
```
    @Test
    public void testClickOnStrokePointerEventsStroke() throws Exception {
        load(PointerEventsTest.class.getClassLoader().getResource("test/html/pointerevents-stroke.html").toExternalForm());
        submit(() -> {
            final WebPage page = WebEngineShim.getPage(getEngine());
            // Render WebPage so that all of the svg paths also will be rendered.
            WebPageShim.paint(page, 0, 0, 800, 600);

            WebPageShim.click(page, 130, 80);
            WebPageShim.click(page, 330, 80);
            WebPageShim.click(page, 530, 80);
            WebPageShim.click(page, 130, 280);
            WebPageShim.click(page, 330, 280);
            WebPageShim.click(page, 530, 280);

            for (String s : asList("polyline", "path", "rect", "circle", "ellipse", "polygon")) {
                getEngine().executeScript("document.getElementById('%s').click()".format(s));
                assertTrue("Expected element '" + s + "' to be activated", (boolean) getEngine().executeScript("isActivated('" + s + "')"));
            }
        });
    }
```

modules/javafx.web/src/test/java/test/javafx/scene/web/PointerEventsTest.java
```diff
@@ -0,0 +1,233 @@
...
+     *
+     * <p>With a javascript event click handler, mouse events on the shapes
+     * are recorded.</p>
+     *
+     * <p>This tests simulates clicks on the fill, so the shapes should not
+     * receive clicks.</p>
+     */
+    @Test
+    public void testClickOnFillPointerEventsStroke() throws Exception {
+        WebView[] view = new WebView[1];
```
arajkumar: Ditto

matthiasblaesing:
@arajkumar thank you for your review. I pushed updates, that should address your comments. The change in the test case is indeed a vast improvement. The formatting for the C++ side should now be more in line with the existing code.
For the stroke parameters you are right. The biggest issue is getting the parameters from the graphics context. As far as I understand the code base, `GraphicsContext` only holds setters for the properties, but no getters. Instead of adding method to the `GraphicsContext` on the C++ side, I moved the `strokeContains` method from `WCPathImpl` to `WCGraphicsPrismContext`, as there the necessary infrastructure is already in place. The `WCRenderQueue` was modified to pass through the query, as the queue needs to be flushed before the query can be invoked. The test was updated to hold a dash sample, that verifies, the dash sample.

modules/javafx.web/src/test/java/test/javafx/scene/web/PointerEventsTest.java
```diff
@@ -0,0 +1,139 @@
...
+            WebPageShim.click(page, 30, 410);
+
+            for (String s : asList("polyline", "path", "rect", "circle", "ellipse", "polygon", "dashed")) {
+                getEngine().executeScript("document.getElementById('%s').click()".format(s));
+                assertFalse("Expected element '" + s + "' not to be activated", (boolean) getEngine().executeScript("isActivated('" + s + "')"));
+            }
+        });
+    }
+
+    public static MouseEvent generateMouseEvent(EventType<MouseEvent> type,
```
arajkumar: Remove this unused method.

modules/javafx.web/src/test/java/test/javafx/scene/web/PointerEventsTest.java
```diff
@@ -0,0 +1,139 @@
...
+import javafx.scene.input.MouseEvent;
+import com.sun.webkit.WebPage;
+import com.sun.webkit.WebPageShim;
+import javafx.scene.web.WebEngineShim;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+public class PointerEventsTest extends TestBase {
```
arajkumar: Add SVG prefix to the test for a better clarity.

modules/javafx.web/src/main/java/com/sun/webkit/graphics/WCRenderQueue.java
```diff
@@ -171,6 +171,15 @@ public boolean isOpaque() {
         return opaque;
     }
 
+    /*is called from native*/
+    private boolean strokeContains(WCPath path, float x, float y) {
```
arajkumar: Though it works, It is weird to have a totally unrelated method in RenderQueue. Rather I would consider to expose a GraphicsContext from RenderQueue and add some helper class(say WCPathHelper) to do the same.

@kevinrushforth, What is your opinion on this?

(Unrelated: This method must be synchronized like other methods of WCRenderQueue)

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/WCPathImpl.java
```diff
@@ -38,6 +38,7 @@
 import com.sun.webkit.graphics.WCRectangle;
 import com.sun.javafx.logging.PlatformLogger;
 import com.sun.javafx.logging.PlatformLogger.Level;
+import com.sun.prism.BasicStroke;
```
arajkumar: Remove unused import

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/WCGraphicsPrismContext.java
```diff
@@ -700,6 +701,23 @@ public void setMiterLimit(float miterLimit) {
...
     }
 
+    @Override
+    public boolean strokeContains(WCPath path, float x, float y) {
+        BasicStroke stroke = state.getStrokeNoClone().getPlatformStroke();
+        boolean result = stroke
+            .createCenteredStrokedShape((Path2D) path.getPlatformPath())
+            .contains(x, y);
+
+        if (log.isLoggable(Level.FINE)) {
```
arajkumar: Log must be added as a first statement of the method. Refer other methods.

modules/javafx.web/src/test/resources/test/html/pointerevents-stroke.html
```diff
@@ -0,0 +1,70 @@
...
+            }
+            rect.clicked, path.clicked, polyline.clicked, line.clicked, circle.clicked, ellipse.clicked, polygon.clicked {
+                stroke: green;
+            }
+        </style>
+        <script>
+            var testElements = [ "polyline", "path", "rect", "circle", "ellipse", "polygon", "dashed" ];
+            window.addEventListener( 'DOMContentLoaded', ( event ) => {
+                for ( var el of testElements ) {
+                    document.getElementById( el ).addEventListener( "mousedown", function ( evt ) {
```
arajkumar: Remove extra space before and after the parameter.
```
window.addEventListener('DOMContentLoaded', (event) => {
```

Ditto for the whole file.

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/WCGraphicsPrismContext.java
matthiasblaesing: I repaired the logging call (it was the wrong syntax), but left it at the current place. My reasoning: the other methods in `WCGraphicsPrismContext` are mostly `void` methods. For these logging is only relevant to get the input parameters. For `strokeContains` the return value is very interesting and I consider it more valuable, than being in line with methods, that are a different category.

modules/javafx.web/src/main/java/com/sun/webkit/graphics/WCRenderQueue.java
matthiasblaesing: The problem is, that the RenderQueue needs to be flushed/decoded before the GraphicsContext can be queried for the containment test. In an ideal world, `SVGRenderSupport::applyStrokeStyleToContext` would work against a minimal interface, that is just implemented by `GraphicsContext` and we could just supply a minimal implementation, that buffers the set values. I tried another approach:

https://github.com/matthiasblaesing/openjdk-jfx/commit/45a7c0a7f21882054578b802ab16b7ede555fd9c

In that implementation the `GraphicsContext` is modified, to buffer the set stroke properties in the `PlatformContextJava`. The `PlatformContextJava` information is then used to create the `BasicStroke` and do the `strokeContains` check.

matthiasblaesing:
I think I addressed all open requests. Anything more that I can do?

kevinrushforth:
> I think I addressed all open requests. Anything more that I can do?

I'll put this on my queue to review.

modules/javafx.web/src/main/java/com/sun/webkit/graphics/WCRenderQueue.java
kevinrushforth: This seems like the biggest unresolved question: should we go with an approach that requires flushing the render queue and querying the context for the containment test, or go with the other approach to shadow the needed properties in the GraphicsContext. In the latter case, there is no need to touch the rendering. My gut feel is that the latter approach might be better, since it avoids interacting with the render queue just to query state, but I will defer to Arun as long as he is comfortable with the current approach.

modules/javafx.web/src/main/java/com/sun/webkit/graphics/WCRenderQueue.java
```diff
@@ -171,6 +171,15 @@ public boolean isOpaque() {
         return opaque;
     }
 
+    /*is called from native*/
+    private synchronized boolean strokeContains(WCPath path, float x, float y) {
```
kevinrushforth: We usually prefix methods that are called form native WebKit code with `fwk`. Arun can comment as to whether it is needed here.

modules/javafx.web/src/main/java/com/sun/webkit/graphics/WCRenderQueue.java
```diff
@@ -171,6 +171,15 @@ public boolean isOpaque() {
...
+        if (gc == null) {
+            return false;
+        }
+        decode();
```
kevinrushforth: Is calling `decode` from a thread other than the PrismRenderer thread a safe thing to do? I realize that you have synchronized this, which is necessary, but is it sufficient?

matthiasblaesing:
@kevinrushforth @arajkumar you both raise doubts about pushing the strokeContains detection into the GraphicsContext/RenderQueue. I see the problem in the interaction between the event thread and the render thread. I don't like mixing these two as I can't predict possible deadlocks there, if the detection needs to be run on the render thread.
As I already explored the option of storing the stroke parameters on the native side, I moved to that approach. The updated PR caches the stroke properties in the `PlatformContextJava`. From there the parameters are extracted and used to check the containment. This mostly returns the changeset to it original form, but `strokeContains` is enhanced by considering all stroke parameters.

kevinrushforth:
Btw, I'll finish my review once you answer the above (minor) questions and decide what to do about the name of `JSLVisitor` (as I mentioned above, leaving it alone is fine).

kevinrushforth:
I think this looks much cleaner now. I'll review next week some time.

matthiasblaesing:
Updated the PR with fixed formatting.

arajkumar:
Hello @matthiasblaesing, As mentioned by @kevinrushforth , currently I'm working on 608.1 WebKit upgrade and I may take another 2 days to complete the review. Sorry for the delay.

matthiasblaesing:
@arajkumar thank you for the info. I noticed that the merge of the webkit update caused build error for this PR. I merged develop into this PR and fixed the build errors (https://github.com/javafxports/openjdk-jfx/pull/479/commits/4eddaf050d49665b9249763a2fb6268c14ca9ff7).

modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/java/PathJava.cpp
```diff
@@ -490,10 +491,58 @@ void Path::apply(const PathApplierFunction& function) const
...
+
+    ASSERT(mid);
+
+    size_t size = strokeStyle == SolidStroke ? 0 : dashes.size();
+    jdouble* dashArrayJavaPrep = new jdouble[size];
+    for (size_t i = 0; i < size; i++) {
+        dashArrayJavaPrep[i] = (jdouble) dashes.at(i);
+    }
+
+    jdoubleArray dashArray = env->NewDoubleArray(size);
```
arajkumar: Use `JLocalRef<jdoubleArray>`, it will automatically deletes the reference when it goes out of scope.(RAII).

```
JLocalRef<jdoubleArray> dashArray(env->NewDoubleArray(size))
env->SetDoubleArrayRegion(jobject(dashArray), 0, size, dashes.data());
```

Since the type of DashArrayElement is `float`, why to use jdouble(instead of jfloat) and again converting it back to `float` from java side?

modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/java/PathJava.cpp
```diff
@@ -490,10 +491,58 @@ void Path::apply(const PathApplierFunction& function) const
...
+    env->SetDoubleArrayRegion(dashArray, 0, size, dashArrayJavaPrep);
+
+    delete[] dashArrayJavaPrep;
+
+    jboolean res = env->CallBooleanMethod(*m_path, mid, (jdouble)p.x(),
+        (jdouble)p.y(), (jdouble) thickness, (jdouble) miterLimit,
+        (jint) cap, (jint) join, (jdouble) dashOffset, dashArray);
+
+    env->DeleteLocalRef(dashArray);
```
arajkumar: You can remove this statement once you use `JLocalRef<jdoubleArray> `

modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/java/PlatformContextJava.h
```diff
@@ -83,9 +83,55 @@ namespace WebCore {
...
+            m_miterLimit = miterLimit;
+        }
     private:
         RefPtr<RenderingQueue> m_rq;
         RefPtr<RQRef> m_jRenderTheme;
         Path m_path;
+        // Buffer the last set stroke styles on the native side to make them
+        // acessible outside the java graphics context
+        DashArray m_dashArray;
+        float m_dashOffset;
```
arajkumar: All the following should be assigned to default values(like below), otherwise it might hold garbage values.
```
float m_dashOffset { };
LineCap m_lineCap { };
LineJoin m_lineJoin { };
float m_miterLimit { };
```

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/WCPathImpl.java
```diff
@@ -344,4 +345,48 @@ public void transform(double mxx, double myx,
         }
         path.transform(BaseTransform.getInstance(mxx, myx, mxy, myy, mxt, myt));
     }
+
+    @Override
+    public boolean strokeContains(double x, double y,
+                                  double thickness, double miterLimit,
```
arajkumar: Since corresponding native types are float, the args `thickness, miterLimit, dashOffset, dashArray`  can be declared as float instead of double.

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/WCPathImpl.java
```diff
@@ -344,4 +345,48 @@ public void transform(double mxx, double myx,
...
+        BasicStroke stroke = new BasicStroke(
+            (float) thickness, cap, join, (float) miterLimit);
+
+        boolean hasDash = false;
+        for (double d : dashArray) {
+            if (d > 0) {
+                hasDash = true;
+                break;
+            }
+        }
```
nlisker: This "contains non-zero element" check is already done in `BasicStroke#set(double[], float)`. I think you can change the constructor to `BasicStroke(thickness, cap, join, meterLimit, dashArray, dashOffest)`.

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/WCPathImpl.java
```diff
@@ -344,4 +345,48 @@ public void transform(double mxx, double myx,
...
+        if (log.isLoggable(Level.FINE)) {
+            StringBuilder dashArrayString = new StringBuilder();
+            dashArrayString.append("[");
+            for (int i = 0; i < dashArray.length; i++) {
+                if (i != 0) {
+                    dashArrayString.append(",");
+                }
+                dashArrayString.append(dashArray[i]);
+            }
+            dashArrayString.append("]");
```
nlisker: Isn't this just `Arrays.toString[double[])`?

nlisker:
Check if copyright years need to be updated in the header.

matthiasblaesing:
@nlisker @arajkumar I updated the PR with fixes of parts of your feedback. I addressed all comments, excluding:

The signature of `WCPathImpl#strokeContains` remained as `double`. While I understand the reasoning, that the native values fit into a java float, I think keeping the signature in line with the rest of the class is worth it. 

That in turn means, that the java array can't be initialized from the `DashArray` directly as the type needs to be converted.

I also kept the `hasDash` check in `WCPathImpl.java`. `BasicStroke#set` runs a similar check, but raises an `IllegalArgumentException`, which is not appropriate for this use-case.

nlisker:
> I also kept the `hasDash` check in `WCPathImpl.java`. `BasicStroke#set` runs a similar check, but raises an `IllegalArgumentException`, which is not appropriate for this use-case.

Then catch the `IllegalArgumentException` and don't rethrow it. The current implementation can also throw that exception if there is a `d < 0.0`, why is it appropriate to throw the exception in that case and not in the "all zero" case?

matthiasblaesing:
Not catching IllegalArgumentException was  a conscious decision, as this might become a hot path, the common case (no dash) must not go through exception handling. I reevaluated this and dug through the webkit code:

- negative lengths are caught in the css property parser (CSSPropertyParser.cpp, line 2124, ValueRangNonNegative)
- at least one non-zero element is enforced in SVRenderSupport.cpp. line 441)

=> So if we have at least one element in the dashArray, we can set it. Updated the PR accordingly.

nlisker:
I don't know enough about Webkit and the css parser to comment. However, the possibility of having a 0 length array in `BasicStroke` to begin with is concerning. `isDashed()` returns `true` in this case, but I think it shouldn't. This is out of scope though.

arajkumar:
> > I also kept the `hasDash` check in `WCPathImpl.java`. `BasicStroke#set` runs a similar check, but raises an `IllegalArgumentException`, which is not appropriate for this use-case.
> 
> Then catch the `IllegalArgumentException` and don't rethrow it. The current implementation can also throw that exception if there is a `d < 0.0`, why is it appropriate to throw the exception in that case and not in the "all zero" case?

In that case, you can add a custom type for java `DashArrayElement` in `DashArray.h`. Something like below?
```
...
#elif PLATFORM(JAVA)
typedef double DashArrayElement;
#else
...
```

If you do that, you can remove the float to double conversion and directly use `env->SetDoubleArrayRegion(dashArray, 0, size, dashes.data());`

matthiasblaesing:
Ok - done.

matthiasblaesing:
Thank you all for advice and patience - would it be possible to consider this bugfix to be brought into OpenJFX 13?

kevinrushforth:
> would it be possible to consider this bugfix to be brought into OpenJFX 13?

Possibly. It's a P4 bug, although since there is no good workaround, an argument could be made for raising it to P3. It seems safe enough. I need to run a full set of Web tests anyway. If I don't see any regressions, then I would be willing to pull it into JavaFX 13.

kevinrushforth:
Testing looks fine, so pushed to jfx-13/rt for JavaFX 13 release.

matthiasblaesing:
Thank you.

PR #501, opened 2019-06-17
JDK-8226234: Simplify iOS property requirements

### LINKED ISSUE #500: Fix properties for ios
There are a few places in the codebase where ios-specific properties are used.

* NativeLibLoader will do a hard check on os.name="iOS" in order to manipulate library names. At all other places, os.name is first converted to lowercase.

* The property "glass.platform" must be said in com.sun.glass.ui.Platform, as there is no default check on ios as os.name

johanvos:
This fixes #500

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
@@ -39,6 +39,7 @@ import java.security.PrivilegedAction;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 
 public class NativeLibLoader {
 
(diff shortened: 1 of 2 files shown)


modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -169,7 +170,7 @@ private static void loadLibraryInternal(String libraryName, List<String> depende
                 //is recognized by existence of JNI_OnLoad_libraryname() C function.
                 //If libraryname contains hyphen, it needs to be translated
                 //to underscore to form valid C function indentifier.
-                if ("iOS".equals(System.getProperty("os.name"))
+                if ("ios".equals(System.getProperty("os.name").toLowerCase(Locale.ROOT))
```
nlisker: Doesn't `equalsIgnoreCase` work here?

johanvos: I doubt. We explicitly need to set the Locale.ROOT to avoid locale-sensitive conversions (see https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#toLowerCase())

Also, in Platform the toLowerCase is used as well.

PR #503, opened 2019-06-18
JDK-8226274: NPE in WinWindow.notifyMoving when Stage with no Scene is shown on 2nd monitor

kevinrushforth:
https://bugs.openjdk.java.net/browse/JDK-8226274

As indicated in the JBS bug, this is a regression introduced by the fix for [JDK-8146920](https://bugs.openjdk.java.net/browse/JDK-8146920). On the Windows platform, when a `Stage` is moved (or initially shown) on a screen other than the one it was one (or other than the default when first showing the Stage), the `View` associated with the `Stage` is notified via a call to `view.updateLocation()`. However, the `view` is null until a scene is set on the stage. The simple, and correct fix is to add a null check.

I also created an automated test based on the test case provided in [JDK-8226274](https://bugs.openjdk.java.net/browse/JDK-8226274). The test is skipped when run on a system with only one monitor. It catches the bug without the fix applied, and passes with the fix applied.


kevinrushforth:
@hjohn Can you take a look as well?

PR #505, opened 2019-06-20
JDK-8225203: Update SQLite to version 3.28.0

arajkumar:
We currently use SQLite version 3.26.0. We should update to the latest stable release version 3.28.0 
https://www.sqlite.org/index.html

JBS link: https://bugs.openjdk.java.net/browse/JDK-8225203


arajkumar:
I will wait one more day for @johanvos's review.

PR #506, opened 2019-06-20
JDK-8226537: gnome-shell crashes when closing a multi-level stage initOwner()

tsayao:
Its a one line fix. It just calls gtk_window_set_transient_for with a NULL second argument to remove the "set_transient_for", which prevents the crash.

https://bugs.openjdk.java.net/browse/JDK-8226537

A test case is not doable because a fail would eventually crash X.org and interrupt all tests. Also, there is no way to assert success.


tsayao:
image

The fails are unrelated to the fix: web related tests and a couple of failing tests because of the system locale.

kevinrushforth:
@pankaj-bansal Since you have easy access to Ubuntu 18.04, can you also review this?

pankaj-bansal:
> @pankaj-bansal Since you have easy access to Ubuntu 18.04, can you also review this?

Yes Kevin, I will review this on Ubuntu 18.04

tsayao:
@johanvos Yes, same behaviour on Wayland (it just hangs a little more before crashing). The fix also works. Tested on Ubuntu 18.04.

PR #517, opened 2019-07-05
8185937: Editable Spinner up/down keys do not increment/decrement

swpalmer:
Added basic check for up/down keys in the KeyEvent filters when Spinner behaviour indicates the arrows are vertical.  Had to make a private method in SpinnerBehavior public to keep the vertical arrow check in one spot.


kevinrushforth:
This looks promising. Can you add a unit test for this?

swpalmer:
Unit test added. I’m not big on git, so I accidentally made a commit with the exact same commit message.  I'm too used to Mercurial, so I tried to amend the existing changeset and had to clean up that mistake. (Mercurial + evolve would be so much better.)
If it’s too much of a mess I can just start over.

kevinrushforth:
> I accidentally made a commit with the exact same commit message

That's fine. The commit messages of subsequent commits don't matter. When this is done, reviewed, and ready to go in, we'll squash them into a single commit with a suitable message.

I do see that there is a failure in the CI build that you will need to look at (probably a white-space issue).

kevinrushforth:
Pointer to JBS issue: https://bugs.openjdk.java.net/browse/JDK-8185937

PR #518, opened 2019-07-06
JDK-8227367 Add missing @Override's and fix static usage.

brcolow:
Very simple PR that adds missing `@Override` annotations to classes in `javafx.graphics/src/main/java/com/sun/prism/*`. This is always good practice but in this specific case it will help anyone trying to implement a new rendering pipeline (Vulkan, DirectX 11/12, etc.) as it makes it easier to see what methods are required to be implemented as they are used by Prism and which methods are implementation details.

I also took the liberty to fix a couple places where a static method was being called on an instance of a class and make the formatting of the empty `traceDispose` methods consistent (putting opening and closing curly brackets on same line).

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DContext.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2019, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DContext.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DContext.java
@@ -308,6 +308,7 @@ class D3DContext extends BaseShaderContext {
         validate(res);
     }
 
+    @Override
     protected void updateWorldTransform(BaseTransform xform) {
         if ((xform == null) || xform.isIdentity()) {
             nSetWorldTransformToIdentity(pContext);
--- a/modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DGraphics.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DGraphics.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009, 2013, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2009, 2019, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DGraphics.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DGraphics.java
@@ -65,6 +65,7 @@ class D3DGraphics extends BaseShaderGraphics implements D3DContextSource {
         return new D3DGraphics(context, target);
     }
 
+    @Override
     public void clearQuad(float x1, float y1, float x2, float y2) {
         // note that unlike clear(), this method does not currently
         // attempt to clear the depth buffer...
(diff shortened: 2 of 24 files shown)


brcolow:
Opened JBS issue: https://bugs.openjdk.java.net/browse/JDK-8226635

nlisker:
You linked the wrong JBS issue, it's https://bugs.openjdk.java.net/browse/JDK-8227367.

brcolow:
Thanks on both accounts, Nir. I should not have been trying to do this so late last night, lol. Added the ones I missed and edited the JBS link. Thanks again.

nlisker:
I think that warnings cleanup is something that should be done for the whole codebase. Warnings are there for something, and if we drown in them we start to ignore them and miss the important ones.
While it's simple to create fixes (the IDE can mass fix all the obvious warnings in one go), I have a feeling that it will clog the review queue pretty badly even when done in batches.

modules/javafx.graphics/src/main/java/com/sun/prism/es2/ES2Graphics.java
```diff
@@ -57,6 +57,7 @@ static void clearBuffers(ES2Context context, Color color, boolean clearColor,
```
nlisker: Update copyright year.

modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DGraphics.java
```diff
@@ -65,6 +65,7 @@ static Graphics create(RenderTarget target, D3DContext context) {
         return new D3DGraphics(context, target);
```
nlisker: Update copyright year.

modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DContext.java
```diff
@@ -308,6 +308,7 @@ protected void updateShaderTransform(Shader shader, BaseTransform xform) {
         validate(res);
     }
```
nlisker: Update copyright year.

modules/javafx.graphics/src/main/java/com/sun/prism/es2/ES2Texture.java
```diff
@@ -631,6 +631,7 @@ public int getNativeSourceHandle() {
         return resource.getResource().getTexID();
     }
```
nlisker: Unused import `com.sun.prism.impl.PrismSettings`.

brcolow:
@nlisker In terms of your comment about warnings - are you saying that you think PRs such as this are less than useful? I don't necessarily disagree, it was low-hanging fruit that I had from my experiments with a D3D11 pipeline.

nlisker:
> In terms of your comment about warnings - are you saying that you think PRs such as this are less than useful?

No, they *are* useful and warning cleanup should be done to all of JavaFX. It's just that it will take a lot of the Reviewers' time and there aren't enough (active) Reviewers.

> experiments with a D3D11 pipeline.

Exactly today I read about [upgrading to D3D11](https://docs.microsoft.com/en-us/windows/uwp/gaming/porting-considerations). Is that what your experiments are geared towards? I'd say it's a worthwhile effort; It's a lot of work, but it's achievable.

brcolow:
Your review helps a lot I think and given that you are now a committer maybe you can be made an official reviewer. Trying to make an experimental D3D11 pipeline is something I was working on, and still am in a limited capacity. That "work" is here: https://github.com/brcolow/openjdk-jfx/tree/d3d11 - most of it is copy paste at this point.

nlisker:
Thanks, but becoming a [Reviewer](https://openjdk.java.net/projects/#project-reviewer) has quite a high bar and it takes Reviewers to make Reviewers.

In any case, you should send a review request.

kevinrushforth:
Btw, with the impending RDP1 date for openjfx13, and other issues in flight, this should be targeted for openjfx14.

modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DShader.java
```diff
@@ -59,19 +59,21 @@ private static native int setConstantsI(long pCtx, long pData, int register,
...
     private static native int nGetRegister(long pCtx, long pData, String name);
 
+    @Override
     public void enable() {
         // res >= 0 is equivalent to D3D's SUCCEEDED(res) macro
         int res = enable(d3dResRecord.getContext().getContextHandle(),
                           d3dResRecord.getResource());
         valid &= res >= 0;
-        d3dResRecord.getContext().validate(res);
+        D3DContext.validate(res);
```
kevinrushforth: This change isn't necessarily equivalent if there are side effects in `D3DRecord::getcontext` that the existing code might be relying upon.

I checked, and there aren't any, so this is just an FYI for future reference.

brcolow: Very interesting, I never considered that actually. So calling a static method on an instance of a class sometimes makes sense?

nlisker: Not really. If `getContext` had side effects, the correct way would be:

```
d3dResRecord.getContext(); // no need to capture the return value 'Context context'
Context.validate(res);
```

The way it's written would be just a shortcut by using the returned instance, but it is not advised to do that anyhow.

kevinrushforth: I would probably characterize that use case as an "unwise" side-effect (at a minimum, a comment would be in order), but it's something to be aware of. Ideally, if that case ever did arise, such that the factory method needed to be called before the static method, I might lean towards making the method in question a "final" instance method rather than a static...

PR #520, opened 2019-07-07
JDK-8227366: Wrong stage gets focused after modal stage creation

tsayao:
Small fix for the problem described on the bug:
https://bugs.openjdk.java.net/browse/JDK-8227366

The commented code does not seem to affect usage, but will do more testing.

Robot tests passes:
image


tsayao:
Played with Ensemble8. No visible issues.

tsayao:
Will do some more testing and a test (if possible) before requesting the review.

kevinrushforth:
OK, you can remove the "WIP" label when you send out the formal RFR.

tsayao:
Did a new fix, explained here:
https://mail.openjdk.java.net/pipermail/openjfx-dev/2019-July/023467.html

Test not possible since Stage.isFocused() reports true even when the stage is not focused (without the fix).

tsayao:
Did more tests with a real application that does modal (WINDOW_MODAL and APPLICATION_MODAL).
Works nice, it removes some flickering and visual glitches, makes usage feels better.

kevinrushforth:
I didn't look at your email before now. To answer the question you posed on the alias:

> Any ideas why this is necessary?
> 
>         if (enabled) {
>             // Check if window is really enabled - to handle nested case
>             if (platformWindow != null && platformWindow.isEnabled()) {
>                 requestToFront();
>             }
>         } 
> 
> Seems wrong to me.

This logic has been present since JDK 7 (FX 2) and was added to fix [JDK-8112637](https://bugs.openjdk.java.net/browse/JDK-8112637). You will need to test and ensure that you haven't reintroduced it (e.g., on Windows 7 or Windows 10).

kevinrushforth:
Update on the above concern: I applied the patch from this PR, and ran it on two different Windows 10 machines. I cannot reproduce the original problem. I had someone else try it on a Windows 7 machine and also cannot reproduce the original problem.

It's possible that other changes (either in more recent versions of cygwin or Windows) have eliminated the need for the workaround.

tsayao:
@kevinrushforth Thanks, I've been struggling to build on windows to test this.

The removal of this workaround saves a lot o event handling on the native side, making a better experience in general with openjfx on linux.

kevinrushforth:
> Thanks, I've been struggling to build on windows to test this.

Yeah, I saw. Windows is currently the most difficult platform to build on. Were you ever able to get it to build?

Anyway, I'll put this on my review queue now.

kevinrushforth:
Pending review from @arapte

Issue #521, opened 2019-07-07
8227425: Add support for e-paper displays on i.MX6 devices


jgneff:
This is a request to add support for e-paper displays with the i.MX 6 Series of Applications Processors. It's a follow-up request to [Issue #339](https://github.com/javafxports/openjdk-jfx/issues/339), "Add support for e-paper displays," which was implemented by [Pull Request #369](https://github.com/javafxports/openjdk-jfx/pull/369).

## Report Classification

**Issue Type:** Request for Enhancement  
**Component:** JavaFX  
**Subcomponent:** *graphics:* JavaFX Graphics  
**Operating System:** Ubuntu  
**Java Release:** 13

### Synopsis

Add support for e-paper displays on i.MX6 devices

### Description

This request follows up on the plans in [Issue #339](https://github.com/javafxports/openjdk-jfx/issues/339) which stated:

> Once the implementation is stable, other more powerful models can be added, such as the Kobo Glo HD with an i.MX6SL processor and the Kobo Clara HD with an i.MX6SLL processor, each based on the ARM Cortex-A9 core.

While [Pull Request #369](https://github.com/javafxports/openjdk-jfx/pull/369) added support for the two oldest e-reader models still being updated by Kobo, the Kobo Touch B and C, this is a request to add support for two more recent models. The following sections provide details on the new device models, new NXP processors, and new Ubuntu releases to be supported.

#### New device models

The following table lists the device characteristics and Linux kernel versions of all four models. The SD column is the storage size of the internal microSD card. The new models to be supported are shown in boldface type.

| Year | Product Name      |  CPU     | Screen (px) | RAM (MiB) | SD (GB) | Kernel |
| ---- | ----------------- | -------- |:-----------:|:---------:|:-------:| ------ |
| 2012 | Kobo Touch B      | i.MX507  |  800 × 600  | 256       | 2       | 2.6.35 |
| 2013 | Kobo Touch C      | i.MX507  |  800 × 600  | 256       | 2       | 2.6.35 |
| 2015 | **Kobo Glo HD**   | i.MX6SL  | 1448 × 1072 | 512       | 4       | 3.0.35 |
| 2018 | **Kobo Clara HD** | i.MX6SLL | 1448 × 1072 | 512       | 8       | 4.1.15 |

#### New NXP processors

[...]

PR #60, opened 2018-04-05
Created a basic .gitignore file

kevinrushforth:
Fixes #21

PR #522, opened 2019-07-08
WIP: JDK-8227425: Add support for e-paper displays on i.MX6 devices

jgneff:
My preliminary tests indicate that this pull request allows JavaFX to support the newer Kobo Glo HD and Kobo Clara HD e-readers of Issue #521 without causing any regression errors for the Kobo Touch B and Kobo Touch C e-readers of Issue #339.

To my surprise, the Kobo Clara HD is backward compatible and works with the support already added to JavaFX by Pull Request #369. The changes in the first commit of this pull request (0dc43591ceaf3ba07285b0741c630ad402d32f22) are due primarily to two problems on the Kobo Glo HD: a minor incompatibility in one of the IOCTL calls, and a bug in the reported size of the frame buffer's virtual resolution.

I marked this pull request as a *Work In Progress* so that I can do some additional testing on all of the devices, along with a comparison of their relative performance.


modules/javafx.graphics/src/main/java/com/sun/glass/ui/monocle/EPDFrameBuffer.java
```diff
@@ -571,20 +586,57 @@ ByteBuffer getOffscreenBuffer() {
          * "QuantumRenderer modifies buffer in use by JavaFX Application Thread"
          * <https://bugs.openjdk.java.net/browse/JDK-8201567>.
          */
-        int size = xresVirtual * yresVirtual * Integer.SIZE;
+        int size = xresVirtual * yres * Integer.BYTES;
```
jgneff, line 589: This is the only problem fixed by this commit that also exists in JavaFX 13. It results in an off-screen buffer that is eight times larger than necessary when using an 8-bit or 16-bit frame buffer. The fix is to use the number of bytes per pixel (`Integer.BYTES`), not the number of bits per pixel (`Integer.SIZE`), when calculating the capacity of the 32-bit off-screen byte buffer.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/monocle/EPDFrameBuffer.java
```diff
@@ -617,6 +675,14 @@ long getNativeHandle() {
      * Gets the virtual horizontal resolution of the frame buffer. See the notes
      * for the {@linkplain EPDFrameBuffer#EPDFrameBuffer constructor} above.
      *
+     * @implNote When using an 8-bit frame buffer, set by the system property
+     * {@code monocle.epd.bitsPerPixel=8}, the Kobo Touch Models N905B and N905C
+     * and the Kobo Glo HD Model N437 work only when this method returns
+     * {@code xresVirtual}, while the Kobo Clara HD Model N249 works only when
+     * this method returns {@code xres}.
+     * <p>
+     * TODO: Make 8-bit frame buffers work for all devices.</p>
```
jgneff, line 684: One way to work around the bogus frame buffer resolution might be to add a system property that overrides the reported value, such as `-Dmonocle.epd.xresVirtual=1448`. Oddly enough, the error does not occur on the Kobo Clara HD when the 8-bit frame buffer is inverted with `-Dmonocle.epd.bitsPerPixel=8 -Dmonocle.epd.y8inverted=true`.

kevinrushforth:

jgneff:
Thanks, Kevin. This pull request is almost done, so I plan to move it over to the new repository when it is ready for review and no longer a _work in progress_. The only other change I might make is to allow for working around the [bogus frame buffer value](#pullrequestreview-260458110) from one of the device models. I could also put the one-line [JavaFX 13 bug fix](#pullrequestreview-260456893) into its own bug report, but JavaFX 13 is not a target for back-porting fixes, so it's probably not worth making it separate.

kevinrushforth:
Sounds like a plan.

jgneff:
This pull request is now [Pull Request #60](https://github.com/openjdk/jfx/pull/60) in the new [JavaFX repository](https://github.com/openjdk/jfx).

PR #524, opened 2019-07-09
JDK-8222912: Websocket client doesn't work in WebView

arajkumar:
Root cause:
`SocketStreamHandleImpl::platformSendHandshake` is not implemented.

Solution:
SocketStreamHandleImpl.cpp already has boilerplate implementation and used by other WebKit ports as well. Switch to the above mentioned implementation and provide only platform specific functionality like connect, send, receive and close. 

JBS link: https://bugs.openjdk.java.net/browse/JDK-8222912

### DIFF
--- a/modules/javafx.web/src/main/native/Source/WebCore/platform/network/java/SocketStreamHandleImplJava.cpp
+++ b/modules/javafx.web/src/main/native/Source/WebCore/platform/network/java/SocketStreamHandleImplJava.cpp
@@ -106,7 +106,7 @@ Optional<size_t> SocketStreamHandleImpl::platformSendInternal(const uint8_t* dat
     if (WTF::CheckAndClearException(env)) {
         return { };
     }
-    return { res };
+    return { static_cast<size_t>(res) };
 }
 
 void SocketStreamHandleImpl::platformClose()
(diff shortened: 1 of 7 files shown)


modules/javafx.web/src/main/native/Source/WebCore/platform/network/java/SocketStreamHandleImplJava.cpp
```diff
@@ -103,14 +104,9 @@ void SocketStreamHandleImpl::platformSend(const uint8_t* data, size_t len, Funct
...
-        completionHandler(false);
-    } else {
-        completionHandler(res == (int)len);
+        return { };
     }
-}
-
-void SocketStreamHandleImpl::platformSendHandshake(const uint8_t*, size_t, const Optional<CookieRequestHeaderFieldProxy>&, Function<void(bool, bool)>&&)
-{
+    return { res };
```
kevinrushforth: This fails to compile on Windows x64 with:

```
modules\javafx.web\src\main\native\Source\WebCore\platform/network/java/SocketStreamHandleImplJava.cpp(109): error C2398:
 Element '1': conversion from 'jint' to 'const T &' requires a narrowing convers
ion
        with
        [
            T=uint64_t
        ]
```

I think you will need a `static_cast<size_t>`

PR #527, opened 2019-07-12
JDK-8193911: Create README file to document bumping the version number

kevinrushforth:
JBS issue: [JDK-8193911](https://bugs.openjdk.java.net/browse/JDK-8193911)

This PR adds a new `UPDATING-VERSION.md` file to the repo with instruction for bumping the JavaFX version number, either for a feature release (13 to 14 to 15) or for a security release (13 to 13.0.1 to 13.0.2).

### DIFF
--- a/UPDATING-VERSION.md
+++ b/UPDATING-VERSION.md
@@ -2,6 +2,8 @@
 
 Here are the instructions for updating the JavaFX release version number
 for a feature release or security (dot-dot) release.
+See [JDK-8226365](https://bugs.openjdk.java.net/browse/JDK-8226365)
+for a recent example.
 
 ## Incrementing the feature version
 
--- a/UPDATING-VERSION.md
+++ b/UPDATING-VERSION.md
@@ -1,0 +1,31 @@
+# Updating the JavaFX Release Version
+
+Here are the instructions for updating the JavaFX release version number
+for a feature release or security (dot-dot) release.
+
+## Incrementing the feature version
+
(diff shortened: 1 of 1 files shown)


arapte:
Will it be helpful to include pointer to  #523 as a reference ?

kevinrushforth:
> Will it be helpful to include pointer to #523 as a reference ?

I don't want to point to any particular repository (especially this one, since it is a mirror and not the official repo), but I could point to an example JBS issue: [JDK-8226365](https://bugs.openjdk.java.net/browse/JDK-8226365).

brcolow:
What about in-lining the diff of #523?

johanvos:
Very minor detail, as I would use 'N' instead of 'M' when talking about the security version number. But I'm ok with 'M' or any other symbol, so +1 from me.

kevinrushforth:
> What about in-lining the diff of #523?

I think it might be overly verbose, and the diff is available from the JBS issue by following the changeset URL or going to the PR. I will add the inline diff to the JBS issue to remove a level of indirection, since that might be useful.

> I would use 'N' instead of 'M' when talking about the security version number

I wanted to use a distinct placeholder symbol from the feature version, mainly in case I wanted to mention `N.0.M` as the version, although I ended up not using that. In any case, I'd prefer to keep it as is.

brcolow:
Using different variable names makes sense. And removing 1 level of indirection is 👍.

PR #533, opened 2019-07-18
8209086: Some javafx.web files are missing GPLv2+Classpath copyright header

kevinrushforth:
JBS issue: [JDK-8209086](https://bugs.openjdk.java.net/browse/JDK-8209086)

As indicated in the bug report, a few of the Oracle-created native files in `modules/javafx.web/src/main/native` have a simple copyright header without any specific license terms. These should have a GPLv2+Classpath exception copyright header.

This PR adds the missing copyright headers to these few (26) files.


arapte:
Could you please check if the below file also needs the copyright change,
`modules\javafx.web\src\main\native\Source\WTF\wtf\java\DbgUtils.h`

kevinrushforth:
Good catch. I missed it because it didn't have any copyright header at all.

PR #535, opened 2019-07-22
JDK-8223722: Cleanup .gitignore file

kevinrushforth:
JBS issue: [JDK-8223722](https://bugs.openjdk.java.net/browse/JDK-8223722)

Simple fix to cleanup the `.gitignore` file and add a few useful missing patterns.

As a help to reviewers, I pushed two commits:

1. Fix/add comments and reorganize the existing patterns without adding or removing any patterns
2. Add new patterns / modify existing patterns

If you want to see what was actually changed (as opposed to reordered), you can just look at the second commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,23 +1,25 @@
-# FIXME: Eventually we should include the needed patterns from .hgingore
-# The goal is that a full build & test, followed by "git status" would show
-# no untracked files
-
-*.orig
-*.bak
-*.rej
-*.swp
-build/
+# Ignore the following directories and files
 .gradle/
-**/nbproject/private/
-dist/
-webrev/
-webrev.zip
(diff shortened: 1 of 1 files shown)


kevinrushforth:
@nlisker can you take a look as well?

nlisker:
Yes, I'll take a look soon.

nlisker:
Looks good with one comment: every Eclipse project has a `.settings` folder with some settings files. The files that start with `org.eclipse.core` are needed and are checked in. If someone uses Buildship (Gradle plugin for Eclipse), then another file is added there that starts with `org.eclipse.buildship`. These are not checked in, so perhaps they should be ignored. They can also be ignored locally.

nlisker:
+1

PR #543, opened 2019-07-30
JDK-8228766: [Mac] fix deadlock in platform.startup(Runnable r)

### LINKED ISSUE #542: Platform.startup() deadlock on mac
Consider these 2 classes:

```
public class First {
         Second.sayHi();
    }
}
```
and

```
public class Second {
    static {
        javafx.application.Platform.startup(() -> {System.err.println("JavaFX started");});
    }

    static void sayHi() {
        System.err.println("HI");
    }
}
```

Running `First` on Linux and Windows will work as expected (printing `HI` on the main thread and `JavaFX started` on the JavaFX Application Thread).

On Mac, nothing is printed. Taking a thread dump shows there is a deadlock: the main thread is waiting in _runLoop and the JavaFX Application Thread is waiting to call the lambda. 
Note: when an inner class is used instead of a lambda, it works fine on mac (see https://bugs.openjdk.java.net/browse/JDK-8036728 for an explanation why).

The wait in _runLoop will only return when "Glass is started", which is done by setting 
`self->started = YES` (in GlassApplication.m). The problem is that this is only done _after_ the lambda is executed on the JavaFX Application Thread, but since the Second class, containing the lambda expression, is not yet class-initialized, the method can not be invoked.

If we set `self->started` to `YES` before executing the runnable passed to `Platform.startup`, the main thread is no longer blocked and can continue with the class initialization of `Second`, and then the lambda expression can be called.

johanvos:
allow main thread to continue before calling the runnable that is provided with Platform.startup().
This fixes #542 or JDK-8228766

### DIFF
--- a/tests/system/src/test/java/test/com/sun/javafx/application/StaticStartupTest.java
+++ b/tests/system/src/test/java/test/com/sun/javafx/application/StaticStartupTest.java
@@ -42,6 +42,15 @@ public class StaticStartupTest {
 
     @Test (timeout=15000)
     public void testStartupFromClinit() throws Exception {
+        Thread thr = new Thread(() -> {
+            try {
+                Thread.sleep(20000);
+            } catch (InterruptedException ex) {}
+            System.err.println("Test timeout exceeded -- calling System.exit");
+            System.exit(1);
+        });
+        thr.setDaemon(true);
+        thr.start();
         StaticClass.doSomething();
     }
 
--- a/tests/system/src/test/java/test/com/sun/javafx/application/StaticStartupTest.java
+++ b/tests/system/src/test/java/test/com/sun/javafx/application/StaticStartupTest.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package test.com.sun.javafx.application;
+
+import javafx.application.Platform;
+import junit.framework.AssertionFailedError;
+import org.junit.AfterClass;
+import org.junit.Test;
+import test.util.Util;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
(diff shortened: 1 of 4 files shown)


johanvos:
This PR introduces different behavior than before on mac, but it is in line with Linux. By declaring that Glass is ready (started), the main thread can continue, similar to e.g. the GTK platform.
The behavior change is that we now first set glass->started to YES and only later call the Application.notifyWillFinishLaunching() . I don't think this is wrong.

kevinrushforth:
JBS issue: https://bugs.openjdk.java.net/browse/JDK-8228766

kevinrushforth:
I'll review this in detail tomorrow.

I note that the exact behavior is currently unspecified. The only thing the docs for `Platform.startup` say is:

```
This method may or may not return to the caller before the run method of the specified Runnable has been called.
```

In the case of the Mac, the startup method *never* returns until after the `Runnable` has run, which seems like undesirable behavior. On the other platforms the startup is at least capable of finishing and returning before (or even during) the execution of the `Runnable`.

I agree that we probably should fix this bug on the Mac. I expect that your proposed fix will be fine, but it will need to be tested in all launch modes -- `Application` (with and without a main method), `JFXPanel`, `FXCanvas`, and `Platform::runLater` -- to ensure no regressions.

Also, to avoid the bug being reintroduced in the future it would be best to add a system test for this use case. Possibly a new test case added to [MainLauncherTest](https://github.com/javafxports/openjdk-jfx/blob/develop/tests/system/src/test/java/test/launchertest/MainLauncherTest.java) or possibly a dedicated launcher test (whichever is easier).

kevinrushforth:
I took a look at adding a test, and I think a separate standalone system test in test/com/sun/javafx/application is best, since there are other Platform startup tests there. I took a stab at creating one. I did it in such a way that it verifies concurrent execution of Platform::runLater and the launch Runnable without needing to use a static block (although a second test that uses a static block could be useful, it might not be necessary).

I will attach a draft of the system test.

kevinrushforth:
Here is the draft of the test: [ConcurrentStartupTest.java](https://gist.github.com/kevinrushforth/ad7739e022748e960b00708112d94c87)

johanvos:
I added a test that uses a clinit to start the Platform.
It passes with the patch, and without the patch it fails (as expected).
However, without the patch, the JVM is in this deadlock and the next tests don't get executed. Not sure what is the most elegant way to fix this?

johanvos:
Do we have to modify the docs as well? With the patch in this PR, we can state that the startup() method does not wait for the runnable to be executed (although it is possible that the runnable is finished first).
This sounds a bit useless, however, it prevents changes in one of the platforms to be incompatible (the new tests would detect this).

kevinrushforth:
> Do we have to modify the docs as well? With the patch in this PR, we can state that the startup() method does not wait for the runnable to be executed...

Yes, I think this sounds like a good idea.

I'll take a look at the static test when I can.

tests/system/src/test/java/test/com/sun/javafx/application/StaticStartupTest.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class StaticStartupTest {
+
+    @Test (timeout=15000)
+    public void testStartupFromClinit() throws Exception {
```
kevinrushforth: Running a (daemon) timeout thread, either here or in a `@BeforeClass` method, that sleeps for 20 seconds and then calls `System.exit(1)` will work:

```
        // Call System.exit after timeout so a deadlock doesn't hang the test harness
        Thread thr = new Thread(() -> {
            try {
                Thread.sleep(20000);
            } catch (InterruptedException ex) {}
            System.err.println("Test timeout exceeded -- calling System.exit");
            System.exit(1);
        });
        thr.setDaemon(true);
        thr.start();
```

kevinrushforth:
I confirmed that the new tests pass on Windows and Linux (unsurprisingly).

I can finish my review once you add the watchdog timeout thread suggested above.

tests/system/src/test/java/test/com/sun/javafx/application/StaticStartupTest.java
johanvos: with this watchdog, the failing test doesn't block the execution of other tests anymore. Thanks.

PR #555, opened 2019-08-08
JDK-8229328: [windows] PlatformFileHandle type should be JGObject rather than void*

guruhb:
On Windows platform PlatformFileHandle type is considered as HANDLE (i.e void *) which is specific to Windows flavor of webkit port. 
For OpenJFX, type of PlatformFileHandle should be JGObject. 

Due to this failure, randomly (at times) the type of FileHandle returned from native webkit will be other than java.io.RandomAccessFile

### DIFF
--- a/modules/javafx.web/src/main/native/Source/WTF/wtf/FileSystem.h
+++ b/modules/javafx.web/src/main/native/Source/WTF/wtf/FileSystem.h
@@ -70,7 +70,7 @@ namespace FileSystemImpl {
 #if USE(GLIB) && !OS(WINDOWS)
 typedef GFileIOStream* PlatformFileHandle;
 const PlatformFileHandle invalidPlatformFileHandle = 0;
-#elif OS(WINDOWS)
+#elif OS(WINDOWS) && !PLATFORM(JAVA)
 typedef HANDLE PlatformFileHandle;
 // FIXME: -1 is INVALID_HANDLE_VALUE, defined in <winbase.h>. Chromium tries to
 // avoid using Windows headers in headers. We'd rather move this into the .cpp.


modules/javafx.web/src/main/native/Source/WTF/wtf/FileSystem.h
```diff
@@ -70,7 +70,7 @@ namespace FileSystemImpl {
 #if USE(GLIB) && !OS(WINDOWS)
 typedef GFileIOStream* PlatformFileHandle;
 const PlatformFileHandle invalidPlatformFileHandle = 0;
-#elif OS(WINDOWS)
+#elif OS(WINDOWS) && !PLATFORM(JAVA)
```
kevinrushforth: This looks like a safe and non-intrusive fix. Longer term, it might be worth refactoring so we don't have to repeat the `PLATFORM(JAVA)` test, but that can be done later.

kevinrushforth:
Pointer to JBS issue: https://bugs.openjdk.java.net/browse/JDK-8229328

modules/javafx.web/src/main/native/Source/WTF/wtf/FileSystem.h
arajkumar: Either changing OS(WINDOWS) to PLATFORM(WIN) or moving PLATFORM(JAVA) to the top would address @kevinrushforth's concerns :)

kevinrushforth: The latter might be best, and can be done in a follow-up (the current fix is fine for now). Thanks.

Issue #567, opened 2019-08-19
WritableImage update fails for empty region


mipastgt:
The newly introduced WritableImages, which are backed by a java.nio.Buffer,
fail to update with an exception when the callback returns an empty Rectangle2D.

```java
java.lang.IllegalArgumentException: srcw (0) and srch (0) must be > 0
	at com.sun.prism.impl.BaseTexture.checkUpdateParams(BaseTexture.java:354)
	at com.sun.prism.es2.ES2Texture.update(ES2Texture.java:640)
	at com.sun.prism.impl.BaseResourceFactory.getCachedTexture(BaseResourceFactory.java:232)
	at com.sun.prism.impl.BaseResourceFactory.getCachedTexture(BaseResourceFactory.java:156)
	at com.sun.javafx.sg.prism.NGImageView.renderContent(NGImageView.java:121)
	at com.sun.javafx.sg.prism.NGNode.doRender(NGNode.java:2072)
	at com.sun.javafx.sg.prism.NGImageView.doRender(NGImageView.java:103)
	at com.sun.javafx.sg.prism.NGNode.render(NGNode.java:1964)
	at com.sun.javafx.sg.prism.NGGroup.renderContent(NGGroup.java:270)
	at com.sun.javafx.sg.prism.NGRegion.renderContent(NGRegion.java:578)
	at com.sun.javafx.sg.prism.NGNode.doRender(NGNode.java:2072)
	at com.sun.javafx.sg.prism.NGNode.render(NGNode.java:1964)
	at com.sun.javafx.tk.quantum.ViewPainter.doPaint(ViewPainter.java:479)
	at com.sun.javafx.tk.quantum.ViewPainter.paintImpl(ViewPainter.java:321)
	at com.sun.javafx.tk.quantum.PresentingPainter.run(PresentingPainter.java:91)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)
	at com.sun.javafx.tk.RenderJob.run(RenderJob.java:58)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[...]

PR #568, opened 2019-08-20
Fix and test for JDK-8229890

mipastgt:
Modified PixelBuffer update method so that an empty rectangle returned by the callback is handled as a no-op and does not throw an exception.
Added a test to ensure that no exception is thrown anymore.
See:
https://bugs.openjdk.java.net/browse/JDK-8229890
https://github.com/javafxports/openjdk-jfx/issues/567

### DIFF
--- a/modules/javafx.graphics/src/test/java/test/javafx/scene/image/PixelBufferTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/scene/image/PixelBufferTest.java
@@ -134,6 +134,17 @@ public final class PixelBufferTest {
     }
 
     @Test
+    public void testUpdatePixelBufferEmptyBufferUpdate() {
+        // This test verifies that an empty dirty region does not cause any exception
+        PixelBuffer<ByteBuffer> pixelBuffer = new PixelBuffer<>(WIDTH, HEIGHT, BYTE_BUFFER, BYTE_BGRA_PRE_PF);
+        Callback<PixelBuffer<ByteBuffer>, Rectangle2D> callback = pixBuf -> {
+            // Assuming this Callback modifies the buffer.
+            return Rectangle2D.EMPTY;
+        };
+        pixelBuffer.updateBuffer(callback);
+    }
+
+    @Test
     public void testUpdatePixelBufferCallbackNull() {
         try {
             PixelBuffer<ByteBuffer> pixelBuffer = new PixelBuffer<>(WIDTH, HEIGHT, BYTE_BUFFER, BYTE_BGRA_PRE_PF);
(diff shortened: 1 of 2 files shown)


modules/javafx.graphics/src/test/java/test/javafx/scene/image/PixelBufferTest.java
```diff
@@ -133,6 +133,17 @@ public void testUpdatePixelBufferPartialBufferUpdate() {
         pixelBuffer.updateBuffer(callback);
     }
 
+    @Test
+    public void testUpdatePixelBufferEmptyBufferUpdate() {
+        // This test verifies that an empty dirty region does not cause any exception
+        PixelBuffer<ByteBuffer> pixelBuffer = new PixelBuffer<>(WIDTH, HEIGHT, BYTE_BUFFER, BYTE_BGRA_PRE_PF);
+        Callback<PixelBuffer<ByteBuffer>, Rectangle2D> callback = pixBuf -> {
+            // Assuming this Callback modifies the buffer.
```
kevinrushforth: This comment should be something like `Assuming no pixels were modified`.

mipastgt:
I fear I am not able to provide a further unit test as requested by Kevin.
1. The proposed path `openjdk-jfx/tests/system/src/main/java/test/scene` does not seem to exist. The best I could come up with is `openjdk-jfx/tests/system/src/test/java/test/javafx/scene`.
2. I have not been able to initialize the runtime system in such a way that I could reproduce the failure in a unit test case.
3. The failure actually does not happen during the update call. It happens later when the system actually tries to perform the update. Also the exception is caught internally and cannot be handled in the test.

The only thing I can provide is a further simplified manual test program. (See attachment.)

[WritableImageFromBufferFailureDemo.java.zip](https://github.com/javafxports/openjdk-jfx/files/3525189/WritableImageFromBufferFailureDemo.java.zip)

kevinrushforth:
Sorry, I did mean `src/test/java/...`

Since the exception is thrown in the, I suspect that the only way to get the test to fail without the fix would be to make the call to update (on the FX app thread), sleep for a while (500 msec should do it), and then check the log file to see whether the exception was reported. We have other examples of this, including the recently added [MenuButtonSkinBaseNPETest.java](https://github.com/javafxports/openjdk-jfx/blob/develop/tests/system/src/test/java/test/javafx/scene/control/MenuButtonSkinBaseNPETest.java) test.

arapte:
Fix looks good to me, +1 pending the test.

mipastgt:
@kevinrushforth I have now written a test following the pattern in [MenuButtonSkinBaseNPETest](https://github.com/javafxports/openjdk-jfx/blob/develop/tests/system/src/test/java/test/javafx/scene/control/MenuButtonSkinBaseNPETest.java). It works reliably outside of the build but which build options or other means do I have to use to actually execute the tests in this branch? With `bash ./gradlew -PFULL_TEST=true -PUSE_ROBOT=true -x web:test all test` these tests never seem to be executed.

kevinrushforth:
Make sure that the test is named something like `XxxxxTest.java` -- the class name *must* end with `Test`. Presuming that this under `tests/system/src/test/java/...` you can execute just that one test with:

```
bash ./gradlew -PFULL_TEST=true -PUSE_ROBOT=true cleanTest :systemTests:test --tests MyNewlyAddedTest
```

kevinrushforth:
Pending review from @arapte

PR #575, opened 2019-08-29
JDK-8207774: TextField must not forward ENTER if action consumed

kleopatra:
This is a (limited, see below) fix for https://bugs.openjdk.java.net/browse/JDK-8207774. 

The issue is that the keyPressed of ENTER on the TextField is forwarded up the parent chain, even if an actionHandler consumed the fired action. The technical reason is that the action may be copied during dispatch, such that the action that's consumed is a different instance than the instance that's sent. If that happens, the consumed state of the sent action will always be false.

This change sends an actionEvent with the textField as eventTarget - doing so prevents the _immediate_ copy of the action. 

To test the change, the sample code in the bug report was converted to two fixtures in TextFieldTest - they failed before the change and pass after.

Limitation: the actionEvent may be copied elsewhere during the dispatch, see related issues in the report (f.i. when having an eventFilter somewhere up the parent chain). If so, the fix doesn't help.

### DIFF
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TextFieldBehavior.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TextFieldBehavior.java
@@ -178,7 +178,8 @@ public class TextFieldBehavior extends TextInputControlBehavior<TextField> {
     @Override protected void fire(KeyEvent event) {
         TextField textField = getNode();
         EventHandler<ActionEvent> onAction = textField.getOnAction();
-        ActionEvent actionEvent = new ActionEvent(textField, null);
+        // fix JDK-8207774: use textField as target to prevent immediate copy in dispatch
+        ActionEvent actionEvent = new ActionEvent(textField, textField);
 
         textField.commitValue();
         textField.fireEvent(actionEvent);
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TextFieldBehavior.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TextFieldBehavior.java
@@ -178,7 +178,7 @@ public class TextFieldBehavior extends TextInputControlBehavior<TextField> {
     @Override protected void fire(KeyEvent event) {
         TextField textField = getNode();
         EventHandler<ActionEvent> onAction = textField.getOnAction();
-        // fix JDK-8207774: use textField as target to prevent immediate copy in dispatch
+        // use textField as target to prevent immediate copy in dispatch
         ActionEvent actionEvent = new ActionEvent(textField, textField);
 
         textField.commitValue();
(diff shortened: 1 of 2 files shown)


kevinrushforth:
@kleopatra - thank you for the proposed fix. Can you send an RFR email to the mailing this (this step will be automated once we have switch over to the Skara-tooling, but it is a manual step for now).

In case you didn't notice, you have whitespace issues, which is why the Travis CI job failed.

kleopatra:
request sent, will go hunting for the whitespace (thanks for the heads-up)

modules/javafx.controls/src/test/java/test/javafx/scene/control/TextFieldTest.java
```diff
@@ -284,6 +298,85 @@
         assertEquals("x", dummyTxtField.getText());
     }
 
-
-
+    private Toolkit tk;
```
kevinrushforth: Since this is unused, it doesn't need to be an instance variable.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TextFieldBehavior.java
```diff
@@ -178,7 +178,8 @@ public void setTextFieldSkin(TextFieldSkin skin) {
     @Override protected void fire(KeyEvent event) {
         TextField textField = getNode();
         EventHandler<ActionEvent> onAction = textField.getOnAction();
-        ActionEvent actionEvent = new ActionEvent(textField, null);
+        // fix JDK-8207774: use textField as target to prevent immediate copy in dispatch
```
kevinrushforth: Minor: I don't think there is a need to reference the bug ID here (the rest of the comment is good)

modules/javafx.controls/src/test/java/test/javafx/scene/control/TextFieldTest.java
```diff
@@ -284,6 +298,85 @@
...
+     *
+     * Here we test that an accelerator is not triggered.
+     */
+    @Test
+    public void testEnterWithConsumingActionHandlerAccelerator() {
+        initStage();
+        root.getChildren().add(txtField);
+        txtField.addEventHandler(ActionEvent.ACTION, e ->
+        e.consume()
+                );
```
kevinrushforth: The indentation is a little odd here: `e.consume()` should be indented and the closing `)` need not be.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TextFieldTest.java
```diff
@@ -284,6 +298,85 @@
...
+        keyboard.doKeyPress(ENTER);
+        assertEquals("actionHandler must be notified", 1, actions.size());
+        assertTrue("action must be consumed ", actions.get(0).isConsumed());
+    }
+
+    /**
+     * Helper method to init the stage only if really needed.
+     */
+    private void initStage() {
+        tk = (StubToolkit)Toolkit.getToolkit();//This step is not needed (Just to make sure StubToolkit is loaded into VM)
```
kevinrushforth: Minor: space after `;` (or else move comment to previous line by itself)

kleopatra:
@kevinrushforth  

thanks for the review - changed as requested. Travis fails, don't know why ..

kevinrushforth:
I looked at the log file for the Travis build. It failed only on Mac, and I also have no idea why it failed. It built one app successfully and then failed with a strange error while building the second app. We'll just chalk it up to some odd problem with that Mac build machine.

PR #576, opened 2019-08-29
JDK-8230361: [web] Cookies are not enabled in WebKit v608.1

guruhb:
Cookiejar Object didn't had valid StorageSessionProvider object, This was due to fix in webkit https://trac.webkit.org/changeset/240014/webkit.

Solution : Create CookieJar with PageStorageSessionProvider (with Page Ref) and add to Page Configuration. 

Test : Load https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_nav_cookieenabled and check for "Cookies enabled: true"

### DIFF
--- a/modules/javafx.web/src/main/native/Source/WebKitLegacy/java/WebCoreSupport/WebPage.cpp
+++ b/modules/javafx.web/src/main/native/Source/WebKitLegacy/java/WebCoreSupport/WebPage.cpp
@@ -53,12 +53,15 @@
 #include <JavaScriptCore/JSContextRefPrivate.h>
 #include <JavaScriptCore/JSStringRef.h>
 #include <JavaScriptCore/Options.h>
+#include <PageStorageSessionProvider.h>
 #include <WebCore/BackForwardController.h>
 #include <WebCore/BridgeUtils.h>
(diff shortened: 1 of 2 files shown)


kevinrushforth:
JBS bug: https://bugs.openjdk.java.net/browse/JDK-8230361

modules/javafx.web/src/main/native/Source/WebKitLegacy/java/WebCoreSupport/WebPage.cpp
```diff
@@ -53,12 +53,15 @@
 #include <JavaScriptCore/JSContextRefPrivate.h>
 #include <JavaScriptCore/JSStringRef.h>
 #include <JavaScriptCore/Options.h>
+#include <PageStorageSessionProvider.h>
```
arajkumar: Since this header is from WebCoreSupport, use "" to include it.

guruhb: Thanks Arun, i have done the changes by using "" include.

kevinrushforth:
Updates to `#include` statements look good to me.

Issue #379, opened 2019-02-15
Poor printing quality


Ciruman:
We are trying to print a Javafx Node with a good resolution and we found out that the javafx prints with a poor resolution(in Mac and Windows). This is related to the following ticket:

https://bugs.openjdk.java.net/browse/JDK-8166194

We found a non-acceptable workarround in the Javafx code:

We added in `J2DPrinterJob.printNode ` the following code:
__((com.sun.javafx.geom.transform.Affine2D)ppg.getTransformNoClone()).scale(1.1, 1.1);__
And looks like this:
```
private void printNode(Node node, Graphics g, int w, int h) {
            PrismPrintGraphics ppg =
                    new PrismPrintGraphics((Graphics2D) g, w, h);
//++++++++++++START
((com.sun.javafx.geom.transform.Affine2D)ppg.getTransformNoClone()).scale(1.0001, 1.0001);
//++++++++++++END
            NGNode pgNode = node.impl_getPeer();
            boolean errored = false;
```

The difference is huge in the javafx controls.

__Could someone help us about it?__

Source code to reproduce the problem:
```
import javafx.application.Application;
import javafx.print.PrinterJob;
import javafx.scene.Node;
import javafx.scene.Scene;
import javafx.scene.control.Button;
import javafx.scene.layout.HBox;
import javafx.stage.Stage;

public class PrintTest extends Application {

    public static void print(final Node node) {
        PrinterJob job = PrinterJob.createPrinterJob();
        job.showPrintDialog(node.getScene().getWindow());
        boolean success = job.printPage(node);
        if (success) {
            job.endJob();
        }
    }

    @Override
    public void start(Stage primaryStage) throws Exception {
        Button print = new Button("Print");
        HBox hBox = new HBox(print);
        print.setOnAction(event -> PrintTest.print(hBox));
        primaryStage.setScene(new Scene(hBox));
        primaryStage.show();
    }

}
```

streifi:
![javafx print comparation](https://user-images.githubusercontent.com/11438177/52870049-40bbcf80-3147-11e9-802e-c0955525f9ee.png)

Left side you see the printing from javafx (blurry edges) and on the right side you see the sharp printing with the workaround applied.

Ciruman:
Of course instead of using:
[...]

PR #577, opened 2019-09-03
JDK-8166194: Poor printing quality

FlorianKirmaier:
ticket: https://github.com/javafxports/openjdk-jfx/issues/379

Fixed poor printing quality for Regions.
The Printer uses vector graphics when possible.
By avoiding caching for Region we ensure high printing quality.

The printing quality is low when effects are used.
In this case, the printer must render pixels.
Ideally, JavaFX would handle printing pixels like rendering on a 4k monitor - by simulating a very high resolution.
But this is not part of the pull request.

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/sg/prism/NGRegion.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/sg/prism/NGRegion.java
@@ -72,6 +72,7 @@ import com.sun.prism.Texture;
 import com.sun.prism.impl.PrismSettings;
 import com.sun.prism.paint.ImagePattern;
 import com.sun.prism.paint.Paint;
+import com.sun.prism.PrinterGraphics;
 import com.sun.scenario.effect.Offset;
 
 /**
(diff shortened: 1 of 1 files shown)


modules/javafx.graphics/src/main/java/com/sun/javafx/sg/prism/NGRegion.java
```diff
@@ -612,7 +612,7 @@ private void renderAsShape(Graphics g) {
             Rectangle rect = null;
             // RT-25013: We need to make sure that we do not use a cached image in the case of a
             // scaled region, or things won't look right (they'll looked scaled instead of vector-resized).
-            if (cacheMode != 0 && g.getTransformNoClone().isTranslateOrIdentity()) {
+            if (cacheMode != 0 && g.getTransformNoClone().isTranslateOrIdentity() && !(g instanceof PrinterGraphics) {
```
kevinrushforth: You are missing a closing paren, which is why the CI build failed.

FlorianKirmaier: fixed

FlorianKirmaier:
Wubs, I thought I've already added the missing bracket.
Now it is correct and the build is green!

kevinrushforth:
I'll do a quick test and then finish the review later today.

For future reference, we prefer that you don't force-push after the initial review has started. It wasn't a big deal in this case, but in general it makes it much harder to review incremental changes. Instead, just push a new commit on top of your original one.

PR #580, opened 2019-09-04
8092352: Skip dispatch if there are no handlers/filters

kleopatra:
This is a fix for https://bugs.openjdk.java.net/browse/JDK-8092352:

EventHandlerManager’s methods dispatchCapturingEvent and dispatchBubblingEvent have _TODOs: skip when no filters/handlers are registered in the CompositeEventHandler_. This is not only a sanitary issue: a side-effect is to copy the fired event, such that the target/handler consumes that copy vs. the original - the sender has no means to test for consume.

The changes: 
- CompositeEventHandler: added internal api to query existence of
handlers/filters
- EventHandlerManager: changed to check existence of handlers/filters
before dispatch
- added tests which are failing before the fix and passing after

### DIFF
--- a/modules/javafx.base/src/test/java/test/com/sun/javafx/event/CompositeEventHandlerTest.java
+++ b/modules/javafx.base/src/test/java/test/com/sun/javafx/event/CompositeEventHandlerTest.java
@@ -27,6 +27,9 @@ package test.com.sun.javafx.event;
 
 import com.sun.javafx.event.CompositeEventHandler;
 import com.sun.javafx.event.CompositeEventHandlerShim;
+
+import static org.junit.Assert.*;
+
 import test.com.sun.javafx.event.EventCountingHandler;
 import test.com.sun.javafx.event.EmptyEvent;
 import javafx.event.Event;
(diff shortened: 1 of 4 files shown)


kevinrushforth:
I will do at least one pass of review before the cut-over to the Skara-enabled git repos. If there are no issues then we may be able to integrate this before the switch (which would save your having to port the PR).

modules/javafx.base/src/test/java/test/com/sun/javafx/event/CompositeEventHandlerTest.java
```diff
@@ -45,6 +48,132 @@ public void setUp() {
...
+     */
+    @Test
+    public void testHasHandlerAddWeakClear() {
+        final EventCountingHandler<Event> eventCountingHandler =
+                new EventCountingHandler<Event>();
+        final WeakEventHandler<Event> weakEventHandler =
+                new WeakEventHandler<Event>(eventCountingHandler);
+        compositeEventHandler.addEventHandler(weakEventHandler);
+        assertTrue("sanity: really added?", CompositeEventHandlerShim.containsHandler(
+                compositeEventHandler, weakEventHandler));
```
kevinrushforth: I see this hear, but not in the filter test. Is there a reason for the difference? I don't have a concern, I'm just curious.

modules/javafx.base/src/test/java/test/com/sun/javafx/event/CompositeEventHandlerTest.java
```diff
@@ -45,6 +48,132 @@ public void setUp() {
...
+     */
+    @Test
+    public void testHasHandlerAddWeak() {
+        final EventCountingHandler<Event> eventCountingHandler =
+                new EventCountingHandler<Event>();
+        final WeakEventHandler<Event> weakEventHandler =
+                new WeakEventHandler<Event>(eventCountingHandler);
+        compositeEventHandler.addEventHandler(weakEventHandler);
+        assertTrue("sanity: really added?", CompositeEventHandlerShim.containsHandler(
+                compositeEventHandler, weakEventHandler));
```
kevinrushforth: Same question here as above.

modules/javafx.base/src/test/java/test/com/sun/javafx/event/CompositeEventHandlerTest.java
```diff
@@ -45,6 +48,132 @@ public void setUp() {
...
+    /**
+     * test report after add/remove handler
+     */
+    @Test
+    public void testHasHandlerAdd() {
+        final EventCountingHandler<Event> eventCountingHandler =
+                new EventCountingHandler<Event>();
+        compositeEventHandler.addEventHandler(eventCountingHandler);
+        assertTrue("sanity: really added?", CompositeEventHandlerShim.containsHandler(
+                compositeEventHandler, eventCountingHandler));
```
kevinrushforth: Same question.

kevinrushforth (approved):
Looks good to me. I have a minor question on the test, but I've approved it regardless.

Either @arapte or @aghaisas will be the second reviewer.

arapte (approved):

modules/javafx.base/src/test/java/test/com/sun/javafx/event/CompositeEventHandlerTest.java
kleopatra: good question :)

Just checked and thought I forgot to make filter tests fully symmetric with handler tests (started work on handlers and copied the available asserts over). So tried a quick addition, and stumbled:

- signature of CompositeEventHandlerShim is containsFilter(CompositeEH, WeakEventHandler) - that is not applicable for not-weak. 
- package test api compositeEventHandler.containsHandler only checks for added, not singletonHandler

Options:

- make filter/handler tests consistent by change/add api of shim and add sanity asserts to filter/singleton handler tests
- remove sanity asserts from new handler tests 
- do nothing (my favorite :)

kevinrushforth: > Options:
> ...
>     * do nothing (my favorite :)

I think we have a winner. :)

I'll integrate this some time this morning.

PR #591, opened 2019-09-18
8200224: Multiple press event when JFXPanel gains focus

FlorianKirmaier:
A side effect of JDK-8200224 is, that the first click on a JFXPanel is always a double click.
The cause for JDK-8200224 is the fix for JDK-8087914 (Clicking on Menu in JFXPanel ignored if another swing component has focus).
This fix introduced synthesized press-events, which is a unsafe fix for the problem.

The pull request introduces a new fix for the problem.
Instead of simulating new input-events, we set JavaFX Scene/Window to focused.
We do so, by using setFocused.
When the original Swing-Focus event is called slightly later, it won't have any side-effects, 
because calling setFocused just sets the value of a boolean property.

### DIFF
--- a/modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
+++ b/modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
@@ -446,13 +446,12 @@ public class JFXPanel extends JComponent {
             (e.getButton() == MouseEvent.BUTTON1)) {
             if (isFocusable() && !hasFocus()) {
                 requestFocus();
-                // this focus request event goes to eventqueue and will be
-                // asynchronously handled so MOUSE_PRESSED event will not be
-                // honoured by FX immediately due to lack of focus in fx
-                // component. Fire the same MOUSE_PRESSED event after
-                // requestFocus() so that 2nd mouse press will be honoured
-                // since now fx have focus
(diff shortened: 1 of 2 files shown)


FlorianKirmaier:
I think the TravisCI build also failed on other commits and isn't caused by my changes.

kevinrushforth:
NOTE: the failing CI build on Mac is due to a bug in `ant 1.10.7`, which happens to be used on the Mac CI machines. See [JDK-8231316](https://bugs.openjdk.java.net/browse/JDK-8231316). You can ignore the failure.

kevinrushforth:
Would it be possible to create a unit test for this? In any case, give the time constraints, it is unlikely this can be integrated before we move the official FX repo to GIT. You will then need to open a new PR against the https://github.com/openjdk/jfx repo.

kevinrushforth:

This [sandbox repository](https://github.com/javafxports/openjdk-jfx) is being retired on 1-Oct-2019. You will need to migrate your pull request to the openjdk/jfx repo in order for this review to continue. Here are [instructions for migrating your pull request](https://mail.openjdk.java.net/pipermail/openjfx-dev/2019-September/023560.html). The updated [CONTRIBUTING.md](https://github.com/openjdk/jfx/blob/master/CONTRIBUTING.md) doc has additional information on how to proceed.

PR #593, opened 2019-09-18
8231126: libxslt.md has incorrect version string

kevinrushforth:
JBS bug: [JDK-8231126](https://bugs.openjdk.java.net/browse/JDK-8231126)

Simple fix to re-apply the correct version number string in the `/legal/libxslt.md` file, which was inadvertently reverted by the fix for [JDK-8222746](https://bugs.openjdk.java.net/browse/JDK-8222746).

### DIFF
--- a/modules/javafx.web/src/main/legal/libxslt.md
+++ b/modules/javafx.web/src/main/legal/libxslt.md
@@ -1,4 +1,4 @@
-## xmlsoft.org: libxslt v1.1.32
+## xmlsoft.org: libxslt v1.1.33
 
 ### libxslt License
 ```


kevinrushforth:
NOTE: the failing CI build on Mac is due to a bug in `ant 1.10.7`, which happens to be used on the Mac CI machines. See [JDK-8231316](https://bugs.openjdk.java.net/browse/JDK-8231316). You can ignore the failure.

kevinrushforth:
@guruhb please review when you get a spare moment (it's a one character fix in the `.md` doc file).

PR #595, opened 2019-09-20
8231310: Add .jcheck/conf to jfx git repo

kevinrushforth:
NOTE: DO NOT MERGE THIS PR

JBS issue: [JDK-8231310](https://bugs.openjdk.java.net/browse/JDK-8231310)

This adds a `.jcheck/conf` file to the `jfx` repo. As indicated in the JBS issue, I will wait to actually integrate this change until the day before the switch to GIT, as the last commit pushed to the HG repo right before that repo is made read-only.

The `.jcheck/conf file` enables the same checks as currently configured for `hg jcheck`. The following rules are enforced:

1. White-space checking in the set of files recognized by `hg jcheck`
2. No executable files in the repo
2. Commit message must start with an issue ID --includes changes for [SKARA-80](https://bugs.openjdk.java.net/browse/SKARA-80)
3. Each commit has a minimum of 1 review from someone with a `Reviewer` role in the project

### DIFF
--- a/.jcheck/conf
+++ b/.jcheck/conf
@@ -43,3 +43,6 @@ minimum=1
 
 [checks "merge"]
 message=Merge
+
+[checks "issues"]
+pattern=^([124-8][0-9]{6}): (\S.*)$
--- a/.jcheck/conf
+++ b/.jcheck/conf
@@ -29,7 +29,7 @@ jbs=jdk
 tags=(jdk-){0,1}([1-9]([0-9]*)(\.(0|[1-9][0-9]*)){0,3})(\+(([0-9]+))|(-ga))|[1-9]((\.\d{1,3}){0,2})-((b\d{2,3})|(ga))|[1-9]u(\d{1,3})-((b\d{2,3})|(ga))
 
 [checks]
-error=blacklist,author,reviewers,merge,hg-tag,message,issues,whitespace,executable
+error=blacklist,author,reviewers,merge,message,issues,whitespace,executable
 
 [census]
 version=0
@@ -40,3 +40,6 @@ files=.*\.java$|.*\.c$|.*\.h$|.*\.cpp$|.*\.hpp$
 
 [checks "reviewers"]
 minimum=1
+
+[checks "merge"]
+message=Merge
--- a/.jcheck/conf
+++ b/.jcheck/conf
@@ -1,0 +1,42 @@
+;
+; Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+; DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+;
+; This code is free software; you can redistribute it and/or modify it
+; under the terms of the GNU General Public License version 2 only, as
+; published by the Free Software Foundation.
+;
+; This code is distributed in the hope that it will be useful, but WITHOUT
+; ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+; FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+; version 2 for more details (a copy is included in the LICENSE file that
+; accompanied this code).
+;
+; You should have received a copy of the GNU General Public License version
+; 2 along with this work; if not, write to the Free Software Foundation,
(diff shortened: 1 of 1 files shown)


kevinrushforth:
@rwestberg @edvbld - if you can review this, it would be helpful.

kevinrushforth:
NOTE: the failing CI build on Mac is due to a bug in `ant`. See [JDK-8231316](https://bugs.openjdk.java.net/browse/JDK-8231316).

edvbld (commented):
Hi Kevin,

a few comments:

- you might want to drop the `hg-tag` check since it, as the name suggests, only applies to tags made with `hg` and you won't be doing those going forward (i.e. the check would only apply to older commits, but those commits will not have the .jcheck/conf, so enabling the `hg-tag` check does nothing).
- I would suggest making the expected message for merge commits explicit as in:
  ```
  [checks "merge"]
  message = "Merge"
  ```
  IIRC then the above are the default values for the `merge` check, but I think it is nice to be explicit.
- Do you want to enforce that a JBS issue can't be resused for multiple commits? If so you can enable the `duplicate-issues` check.

Thanks,
Erik

kevinrushforth:
Thanks for the review.

>     * you might want to drop the `hg-tag` check since it, as the name suggests, only applies to tags made with `hg` and you won't be doing those going forward 

Will do.

>     * I would suggest making the expected message for merge commits explicit as in:
>       ```
>       [checks "merge"]
>       message = "Merge"
>       ```

Good suggestion. I'll add this.

>     * Do you want to enforce that a JBS issue can't be resused for multiple commits? If so you can enable the `duplicate-issues` check.

Thanks for the suggestion, but no, we don't want this check, for the same reason that the JDK project eventually dropped them (the case of cherry-picking a bug fix that went into the mainline for backporting to a stabilization branch that is still being synced back into mainline).

kevinrushforth:
Update:

> [checks "merge"]
>  message = "Merge"

Turns out that should be:
```
message=Merge
```
Otherwise the literal `"` chars are expected to be in the message.

johanvos:
This looks good to me, but it seems easy to make typo's that will only surface once this is in production. Is there a way we can dry-run a PR through this check without actually committing?

johanvos (commented):
This looks good to me, but it seems easy to make typo's that will only surface once this is in production. Is there a way we can dry-run a PR through this check without actually committing?

kevinrushforth:
> Is there a way we can dry-run a PR through this check without actually committing?

You can run "git jcheck" locally if you have configured the Skara tooling. You would just need to ignore the complaint about not (yet) having a reviewer.

prrace (commented):
Why do we have to define what are presumably parameters for some of the checks ?
Aren't the values here what you'd expect as a default ?

kevinrushforth:
I think the only one that is default when you specify a full config file (as opposed to a legacy file without separate sections), is the message for merge commits, which is `Merge`. I don't think any others are default. I can double-check, and @edvbld probably knows off the top of his head, but even so, it might be cleaner to specify them.

I know that the `tag` format is not default, nor is the issue `pattern` string.

prrace (approved):
OK .. not sure how to directly reply to a comment. Seems like it should be possible but I don't see a reply button.

kevinrushforth:
> not sure how to directly reply to a comment

As far as I know GitHub doesn't do threaded comments unless they are inline as part of the code, but I might have missed a way to do it.

johanvos:
> I think the only one that is default when you specify a full config file (as opposed to a legacy file without separate sections), is the message for merge commits, which is `Merge`. I don't think any others are default. I can double-check, and @edvbld probably knows off the top of his head, but even so, it might be cleaner to specify them.
> 
> I know that the `tag` format is not default, nor is the issue `pattern` string.

I agree with specifying them here. Defaults can change.

johanvos (approved):

kevinrushforth:
Now that this is approved, I will push it to the HG repo as the last changeset pushed right before making it read-only, at which time I will close this PR.

kevinrushforth:
I pushed this to HG: https://hg.openjdk.java.net/openjfx/jfx-dev/rt/rev/6744f504e684

It will be synced to this sandbox repo within the hour. No further pushes to this repo are expected.

PR #596, opened 2019-09-20
WIP: Print out the version of "ant" being used on the CI machines

kevinrushforth:


kevinrushforth:
This is an attempt to see why the Mac CI builds are failing.

kevinrushforth:
The following is from the failing log on the Mac machine:

```
which ant: /usr/local/bin/ant
ant -version
Apache Ant(TM) version 1.10.7 compiled on September 1 2019
```

I can confirm the same failure if I download ant 1.10.7 and build locally. This is a problem with our relying on whatever version of the tool is installed on the machine rather than requiring and downloading the precise version that we specify in the `build.properties` file, which is 1.10.5. I note that we have this same potential problem with the compilers or other tools that we take from the system without downloading a specific version.

Since we are going to switch over to the openjdk/jfx repo and the Skara tools, this will be a short-lived problem, although one we will have to address when we enable CI builds for those repos (not planned for the initial transition, but something I hope to get going after a while).

kevinrushforth:
I filed [JDK-8231316](https://bugs.openjdk.java.net/browse/JDK-8231316) in JBS to track this. I plan to file a bug against `Apache ant`.

anthraxx:
@kevinrushforth did you by any chance report this to ant? If so, can you please post the report? If not, could you pleeeease :cat:

lipnitsk:
I reported it here: https://bz.apache.org/bugzilla/show_bug.cgi?id=63874

